PR #8, opened 2017-01-31
docs(readme): add docker version

ellerbrock:
yolo

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,16 @@ Installation
 See doc/install.txt in the source code archive.
 Python 2.7.2 or later is needed.
 
+
+Install using Docker
+---------------------
+You can try linkchecker in an isolated enviroment without changing any local files or have to install all the dependencies via a Docker Container:
+
+``docker pull ellerbrock/linkchecker``
+
+https://hub.docker.com/r/ellerbrock/link-checker/
+
+
 Usage
 ------
 Execute ``linkchecker http://www.example.com``.


README.rst
```diff
@@ -32,6 +32,16 @@ Installation
...
 Python 2.7.2 or later is needed.
 
+
+Install using Docker
+---------------------
+You can try linkchecker in an isolated enviroment without changing any local files or have to install all the dependencies via a Docker Container:
+
+``docker pull ellerbrock/linkchecker``
+
+https://hub.docker.com/r/ellerbrock/link-checker/
```
anarcat: can you be more explicit here? what is this link? how is it related to the project?

maybe mention that this is not an official build? otherwise why "ellerbrock" instead of "linkchecker"?

how often are those images rebuilt? what version?

so many questions... :)

i'm not very familiar with docker, so maybe that makes no sense at all... :)

mgedmin: Also: what should the user do after running the docker pull?  I assume it'll be something like `docker run linkchecker /usr/bin/linkchecker http://example.com`?

ellerbrock: hi long story short: you can run link-checker without the pain to install all the dependencies on the local machine in an isolated enviroment. in case there is an "offical" docker hub account i can make a pull request with the dockerfile itself if you like. this one gets maintained by me and an example how to use it for ci stuff you can see here:
https://github.com/ellerbrock/travis-link-checker

cheers

anarcat: sure. i would just like to make sure users have answers to the questions i asked before putting this in the main readme file... otherwise maybe it could just be marked as unofficial?

ellerbrock: hi @anarcat,

basically all the docker stuff on docker hub is "unofficial".
real big projects get added to the docker library: https://hub.docker.com/explore/, but not sure what it takes to get there.

look for the bash-it project we handle this like that.
they have a [bash-it](https://github.com/Bash-it/bash-it) main repo and a [docker-bash-it](https://github.com/Bash-it/bash-it-docker) which i maintain for example. when i find better ways to do stuff i update this and add it to the docker stuff i maintain ...

anarcat: i disagree. there are some official projects. here is the [debian repository](https://hub.docker.com/_/debian/). it says "official" on top. same as the [ubuntu](https://hub.docker.com/_/ubuntu/) one.

again, i don't really care if it's official or not. if it's unofficial, just say so in the patch. if it's official, we need to have access to it, have an automated procedure to rebuild it and so on.

Issue #7, opened 2017-01-31
tests are currently broken


anarcat:
It seems that tests fails on HEAD since 0ef00ee. Here's the travis build log:

https://travis-ci.org/wummel/linkchecker/builds/104276729

This affects all pull requests which also fail to pass tests.

this was previously reported in https://github.com/wummel/linkchecker/issues/681

mgedmin:
For some reason Travis excludes some of the tests.  If I run the full test suite (`make test` without the `TESTS="<explicit subset specified in .travis.yml>"`), I see four test failures:

```
====================================================================== FAILURES ======================================================================
__________________________________________________________ TestConsole.test_internal_error ___________________________________________________________

self = <tests.test_console.TestConsole testMethod=test_internal_error>

    def test_internal_error (self):
>       linkcheck.director.console.internal_error()

tests/test_console.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
linkcheck/director/console.py:99: in internal_error
    print_app_info(out=out)
linkcheck/director/console.py:152: in print_app_info
    print(configuration.get_modules_info(), file=out)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def get_modules_info():
        """Return unicode string with detected module info."""
        module_infos = []
        for (mod, name, version_attr) in Modules:
            if not fileutil.has_module(mod):
                continue
>           if hasattr(mod, version_attr):
E           TypeError: hasattr(): attribute name must be string

linkcheck/configuration/__init__.py:84: TypeError
---------------------------------------------------------------- Captured stderr call ----------------------------------------------------------------


********** Oops, I did it again. *************

You have found an internal error in LinkChecker. Please write a bug report
at https://github.com/wummel/linkchecker/issues
and include the following information:
- the URL or file you are testing
- the system information below

When using the commandline client:
[...]

PR #9, opened 2017-02-01
Add a tox.ini

mgedmin:
https://pypi.python.org/pypi/tox is a Python testing tool that lets you run
tests without polluting your system environment or managing virtualenvs by
hand.  It is especially invaluable if you want to support multiple Python
versions (e.g. 2.7 and 3.x).

The way tox normally works is
- it builds a source distribution with `python setup.py sdist`
- it installs the sdist with `pip instlal dist/package-version.tar.gz`
- it runs the tests
and so it also checks that your sdists are complete.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist = py27
 
 [testenv]
+usedevelop = true
 deps =
     -rrequirements.txt
     pyftpdlib
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,12 @@
+[tox]
+envlist = py27
+
+[testenv]
+deps =
+    -rrequirements.txt
+    pyftpdlib
+    pytest-xdist
+commands =
+    py.test {posargs:tests}
+setenv =
+    LC_ALL=en_US.utf-8
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,6 @@ deps =
     pyftpdlib
     pytest-xdist
 commands =
-    py.test {posargs:tests}
+    py.test {posargs:--tb=short tests}
 setenv =
     LC_ALL=en_US.utf-8
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ Changelog.linkchecker*
 /alexa*.log
 /testresults.txt
 /linkchecker.prof
+.cache
+.tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,12 @@ python:
 branches:
   only:
     - master
-# install required system packages
-addons:
-  apt:
-    packages:
-    - python-dev
 # command to install dependencies
 install:
-  - pip install -r requirements.txt
-  - pip install pyftpdlib pytest-xdist
+  - pip install tox-travis
 # command to run tests
-script: make test PYTESTOPTS="--tb=short" TESTS="tests/test_*.py tests/configuration/test_*.py tests/logger/test_*.py tests/checker/test_httpbin.py"
+script:
+  - tox
 # do not send emails of broken builds
 notifications:
   email: false
--- a/tests/checker/data/http.html.result
+++ b/tests/checker/data/http.html.result
@@ -48,14 +48,14 @@ real url http://localhost:%(port)d/?q=%%C3%%BC
 name html entities
 valid
 
-url file.css
-cache key http://localhost:%(port)d/tests/checker/data/file.css
-real url http://localhost:%(port)d/tests/checker/data/file.css
-valid
-
 url file.html
 cache key http://localhost:%(port)d/tests/checker/data/file.html
 real url http://localhost:%(port)d/tests/checker/data/file.html
+valid
+
+url file.css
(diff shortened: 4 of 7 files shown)


tox.ini
```diff
@@ -0,0 +1,13 @@
...
+envlist = py27
+
+[testenv]
+usedevelop = true
+deps =
+    -rrequirements.txt
+    pyftpdlib
+    pytest-xdist
+commands =
+    py.test {posargs:tests}
```
anarcat: i'm not sure this matters, but this is slightly different than what travis is currently doing, which is:

```
script: make test PYTESTOPTS="--tb=short" TESTS="tests/test_*.py tests/configuration/test_*.py tests/logger/test_*.py tests/checker/test_httpbin.py"
```

`make test`, in turn, does this:

```
test:   localbuild
        env LANG=en_US.utf-8 $(PYTHON) -m pytest $(PYTESTOPTS) $(TESTOPTS) $(TESTS)
```

And those variables are:

```
TESTS ?= tests
# set test options, eg. to "--verbose"
TESTOPTS=
# Pytest options:
# - use multiple processes
# - write test results in file
PYTESTOPTS:=-n 4 --resultlog=testresults.txt
```

Arguably, this is a mess, and I prefer to have your clean tox interface... but is there something in there we're missing we'd like to add?

in any case, i think it would be preferable to *switch* to tox instead of *adding* it. that would mean changing the `.travis.yml` configuration along with the `Makefile`.

arguably, we should also make `python setup.py test` actually work (and that's possible!) but that may be out of scope here (as it doesn't currently work).

mgedmin:
I've rebased on master to pick up the fix for #10, and updated `.travis.yml` to run `tox.`

Now this is once again work-in-progress:

- [ ] some that used to be excluded from Travis are now failing

I would prefer to disable those tests [using py.test's API](http://doc.pytest.org/en/latest/skipping.html#marking-a-test-function-to-be-skipped) instead of whitelisting known-good tests with shell glob patterns.

(Plus, the commands specified in `tox.ini` aren't filtered through a shell so you cannot use shell globs in there, unless you do an explicit `sh -c "command that expands test_*.py etc"`, which is just *ugh*).

tox.ini
mgedmin: `-n 4` tells pytest-xdist to run the tests in parallel using 4 CPU cores.  This speeds up the test run from 60 seconds to 40 seconds on my laptop.  It may be useful to preserve this.

`--resultlog` is a deprecated py.test command-line argument that's going to be removed in py.test 4.0.  I think we can drop it.

mgedmin: I do not like `python setup.py test`.

The very-long-term direction in the Python packaging ecosystem is to get rid of `setup.py`, see e.g. [PEP-517](https://www.python.org/dev/peps/pep-0517/).   It will take many years, but in the mean time I don't think it's worth to spend effort adding features in a build system that's going to be eventually deprecated.

mgedmin:
Well, another unexpected series of failures in Travis.  Interesting.

anarcat:
On 2017-02-01 06:21:02, Marius Gedminas wrote:
> I would prefer to disable those tests [using py.test's API](http://doc.pytest.org/en/latest/skipping.html#marking-a-test-function-to-be-skipped) instead of whitelisting known-good tests with shell glob patterns.

I agree.

> (Plus, the commands specified in `tox.ini` aren't filtered through a shell so you cannot use shell globs in there, unless you do an explicit `sh -c "command that expands test_*.py etc"`, which is just *ugh*).

Indeed.

mgedmin:
I can reproduce with `git clean -dfx && tox`.

I can make things work again with `git clean -dfx && make test && tox`.

Apparently the Makefile performs some extra steps that are omitted if you just use setuptools:
```
cp -f build/lib.linux-x86_64-2.7*/linkcheck/HtmlParser/htmlsax*.so linkcheck/HtmlParser
cp -f build/lib.linux-x86_64-2.7*/linkcheck/network/_network*.so linkcheck/network
```

tox.ini
anarcat: agreeing with everything. didn't know about setup.py, that's useful to know.

so the only thing left here is to integrate tox into travis, IMHO, plus the task you added to fix extra tests.

thanks!

anarcat:
yuck yuck yuck... i guess we need to clean that up as well...

i feel we're going to get a lot of that stuff.. ;)

anarcat:
hmm.. tests still seem to fail on travis... is test skipping not working?

tox.ini
anarcat: done in 97bb572

mgedmin:
> hmm.. tests still seem to fail on travis... is test skipping not working?

Nah, that test is non-deterministic, like I feared in my commit message.

anarcat:
so should we fix this before we merge? maybe we merge and bounce that back to #7... i kicked travis to see if we can get *some* green in here... :)

mgedmin:
I'll create separate issues for the known-to-fail tests.

anarcat:
awesome, thanks!

PR #11, opened 2017-02-01
Fix TypeError: hasattr(): attribute name must be string

### LINKED ISSUE #7: tests are currently broken

mgedmin:
The one test failure in Travis happens in TestConsole.test_internal_error, but only if you have the argcomplete package installed.

This was a real bug in error reporting code.

Fixes #7.

### DIFF
--- a/linkcheck/configuration/__init__.py
+++ b/linkcheck/configuration/__init__.py
@@ -81,7 +81,7 @@ def get_modules_info():
     for (mod, name, version_attr) in Modules:
         if not fileutil.has_module(mod):
             continue
-        if hasattr(mod, version_attr):
+        if version_attr and hasattr(mod, version_attr):
             attr = getattr(mod, version_attr)
             version = attr() if callable(attr) else attr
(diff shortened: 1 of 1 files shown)


mgedmin:
The tox failure is weird.

anarcat:
that *is* weird... why the heck does it fail there? it didn't fail before and you didn't change that part...

i restarted the build to see if that was transient (!).

but yes, this totally needs fixing. i believe, there were other PRs to fix this in the original project, you may want to look into those... if only to mark them as duplicates in the wiki:

https://github.com/linkcheck/linkchecker/wiki/Historic-issues

anarcat:
well shit, that did it. we can work on that transient issue in #7

PR #23, opened 2017-02-03
Wayback urls

seamang:
Allow linkchecker to process wayback style URLs, ie URLs used by web archives which contain a full scheme embedded in the query string, like:

http://web.archive.org/web/20010106023100/http://www.bbc.co.uk/

By default linkchecker url-encodes the second http[s]://, making the URL invalid. 

The use case for this feature is web archives wanting to crawl their archived sites as part of QA. This is very much a minority use, so the option has been made configurable with a default value 'off'.

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -49,6 +49,7 @@ from .const import (WARN_URL_EFFECTIVE_URL,
     WARN_URL_CONTENT_SIZE_ZERO, WARN_URL_CONTENT_SIZE_TOO_LARGE,
     WARN_URL_WHITESPACE, URL_MAX_LENGTH, WARN_URL_TOO_LONG,
     ExcList, ExcSyntaxList, ExcNoCacheList)
+from ..url import url_fix_wayback_query
 
 # helper alias
 unicode_safe = strformat.unicode_safe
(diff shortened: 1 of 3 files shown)


seamang:
So  would it be acceptable for me to remove the optional stuff and just silently fix wayback style urls for now?

anarcat:
yes, that would actually be more acceptable. :)

seamang:
Removed configuration option for wayback urls and made it default behaviour to pass http[s]:// in path section unencoded - except for atom feed:http://.. style urls. This is needed to pass the atom test in test_misc.py; I don't understand the need for this behaviour but it may be related to the Changelog entry for 2012: 'ignore feed: urls'.

anarcat:
alright, this looks good to me now. i've stared at this code for too long now though, maybe @mgedmin wants to have a look before merging?

otherwise i'm curious to hear if you can merge this yourself, @seamang :)

anarcat:
@mgedmin agreed... thanks for the review, and thanks @seamang for the fix! i can't believe i didn't notice archive.org urls were broken all this time. ;)

seamang:
I do agree about the whole mangle/unmangle process though… the problem for me is trying to unpick it is likely to have too many unforeseen consequences (like the impact on the atom feed:http:… URLs, which I'd never even been aware of before…)

PR #24, opened 2017-02-03
Make fix

seamang:
Can't find any trace of '-n 4' as a valid option in pytest documentation so have just removed it. 

This is a fix for https://github.com/wummel/linkchecker/issues/693 which I've closed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ endif
 # Pytest options:
 # - use multiple processes
 # - write test results in file
-PYTESTOPTS:=-n 4 --resultlog=testresults.txt
+PYTESTOPTS:= --resultlog=testresults.txt
 
 
 all:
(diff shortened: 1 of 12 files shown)


mgedmin:
`-n 4` is an option provided by the [pytest-xdist](https://pypi.python.org/pypi/pytest-xdist) plugin.  It's used to run the tests in parallel, for a moderate speedup (38s versus 57s).

If you rebase this PR to drop all the unrelated changes, you may also want to remove the `--resultlog=testresults.txt` argument, which is deprecated and is going to be removed in py.test 4.0.

seamang:
Oops. Sorry, my fault for rushing. But am unclear now about the `-n 4`. Can we close this PR and could @mgedmin  provide one which handles checking for pytest-xdist?

mgedmin:
I'm not sure how to check for pytest-xdist.  Maybe we should add a `dev-requirements.txt` and list it there, to act as a clue for developers who don't use tox.

If you use `tox` to run tests it will install pytest-xdist in a virtualenv automatically (but, ironically, `tox.ini` doesn't currently specify `-n 4`, although you can pass arbitrary py.test options with `tox -- --tb=short -n 4 tests`).

Issue #34, opened 2017-02-12
Two warnings building on OSX 10.10.5


richb-hanover:
Using OSX 10.10.5, Python 2.7.11:

I ran `make localbuild` as specified in https://github.com/linkcheck/linkchecker/blob/master/doc/development.mdwn, and got the following build progress and warnings. 

The resulting `./linkchecker` seems to work fine, though. I'm just reporting in case this is unknown info.

```
bash-3.2$ cd github/linkchecker-reborn/
bash-3.2$ make localbuild
python2.7 setup.py sdist --manifest-only
creating /Users/richb/github/linkchecker-reborn/_LinkChecker_configdata.py
running sdist
running check
warning: MySdist: standard file not found: should have one of README, README.txt

reading manifest template 'MANIFEST.in'
warning: no files found matching 'doc/html/*.ico'
warning: no files found matching 'doc/html/*.html'
warning: no files found matching 'doc/html/*.txt'
warning: no files found matching 'doc/html/*.png'
warning: no files found matching 'doc/html/*.qhp'
warning: no files found matching 'doc/html/*.qhcp'
warning: no files found matching 'doc/html/Makefile'
writing manifest file 'MANIFEST'
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C po
make[1]: Nothing to be done for `all'.
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C linkcheck/HtmlParser
make[1]: Nothing to be done for `all'.
python2.7 setup.py build
creating /Users/richb/github/linkchecker-reborn/_LinkChecker_configdata.py
running build
running build_py
creating build
creating build/lib.macosx-10.10-x86_64-2.7
creating build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/__init__.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/ansicolor.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/better_exchook2.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/cmdline.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/colorama.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
copying linkcheck/containers.py -> build/lib.macosx-10.10-x86_64-2.7/linkcheck
[...]

PR #36, opened 2017-02-13
Fix "operation on ... may be undefined" [-Wsequence-point] warnings

mgedmin:
Fixes a bunch of warnings like

  htmlparse.y:509:25: warning: operation on ‘self->userData->buf’ may be undefined [-Wsequence-point]
  htmlparse.y:518:29: warning: operation on ‘self->userData->tmp_buf’ may be undefined [-Wsequence-point]

which were a result of (macro-expanded) code like this (simplified):

  if ((tmp = (tmp = PyMem_Realloc(...))) == NULL) return NULL;

The PyMem_Resize(p, ...) macro assigns the new value to p before
returning it, so there's no need to assign it again.

See http://bugs.python.org/issue1668036 for evidence (from 2007) that
this is indeed a documented side-effect of the macro API.

This fixes one half of #34.

The diff is not very pretty because we have bison-generated code in our
Git repository.

### DIFF
--- a/linkcheck/HtmlParser/htmllex.c
+++ b/linkcheck/HtmlParser/htmllex.c
@@ -17,8 +17,8 @@
 
 #define FLEX_SCANNER
 #define YY_FLEX_MAJOR_VERSION 2
-#define YY_FLEX_MINOR_VERSION 5
-#define YY_FLEX_SUBMINOR_VERSION 39
+#define YY_FLEX_MINOR_VERSION 6
+#define YY_FLEX_SUBMINOR_VERSION 1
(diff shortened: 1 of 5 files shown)


mgedmin:
Perhaps we should get rid of a custom HTML parser written in C (because [those often have bugs](https://blog.cloudflare.com/incident-report-on-memory-leak-caused-by-cloudflare-parser-bug/)) and replace it with `lxml`?

anarcat:
absolutely. anything but a *generated* C parser would do... there's Beautifulsoup out there, iirc.

PR #39, opened 2017-02-13
Fix cache: Don't check one url multiple times

PetrDlouhy:
As mentioned in #608, Linkchecker does multiple checks for one URL. It is caused by inconsistent use of cache. The URL is added to the queue when checking page, but it doesn't get to cache until the moment of checking the URL itself. In the meantime, it can be added to queue multiple times.

This PR should fix it.

### DIFF
--- a/linkcheck/cache/results.py
+++ b/linkcheck/cache/results.py
@@ -59,6 +59,10 @@ class ResultCache(object):
         """Non-thread-safe function for fast containment checks."""
         return key in self.cache
 
+    def has_non_empty_result(self, key):
+        """Non-thread-safe function for fast containment checks."""
+        return self.cache.get(key)
+
     def __len__(self):
         """Get number of cached elements. This is not thread-safe and is
         likely to change before the returned value is used."""
--- a/linkcheck/cache/urlqueue.py
+++ b/linkcheck/cache/urlqueue.py
@@ -120,7 +120,9 @@ class UrlQueue (object):
         log.debug(LOG_CACHE, "queueing %s", url_data.url)
         key = url_data.cache_url
         cache = url_data.aggregate.result_cache
-        if url_data.has_result or cache.has_result(key):
+        if cache.has_result(key):
+                return
+        if url_data.has_result:
             self.queue.appendleft(url_data)
         else:
             assert key is not None, "no result for None key: %s" % url_data
@@ -131,6 +133,7 @@ class UrlQueue (object):
                 self.cleanup()
             self.queue.append(url_data)
         self.unfinished_tasks += 1
+        cache.add_result(key, None)  # add none value to cache to prevent checking this url multiple times
 
     def cleanup(self):
         """Move cached elements to top."""
@@ -139,7 +142,7 @@ class UrlQueue (object):
         for i, url_data in enumerate(self.queue):
             key = url_data.cache_url
             cache = url_data.aggregate.result_cache
-            if cache.has_result(key):
+            if cache.has_non_empty_result(key):
                 cached.append(i)
         for pos in cached:
             self._move_to_top(pos)
--- a/tests/cache/test_urlqueue.py
+++ b/tests/cache/test_urlqueue.py
@@ -1,0 +1,174 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2017 Petr Dlouhý
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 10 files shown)


anarcat:
tests seem to fail after this... can you investiage?

linkcheck/cache/results.py
```diff
@@ -59,6 +59,10 @@ def has_result(self, key):
         """Non-thread-safe function for fast containment checks."""
         return key in self.cache
 
+    def has_non_empty_result(self, key):
+        """Non-thread-safe function for fast containment checks."""
+        return key in self.cache and self.cache[key]
```
mgedmin: Could be simplified to `return self.cache.get(key)`

linkcheck/cache/urlqueue.py
```diff
@@ -120,17 +120,19 @@ def _put (self, url_data):
...
+            else:
+                assert key is not None, "no result for None key: %s" % url_data
+                if self.max_allowed_urls is not None:
+                    self.max_allowed_urls -= 1
+                self.num_puts += 1
+                if self.num_puts >= NUM_PUTS_CLEANUP:
+                    self.cleanup()
+                self.queue.append(url_data)
+            self.unfinished_tasks += 1
+            cache.add_result(key, None) # add none value to cache to prevent checking this url multiple times
```
mgedmin: This diff is hard to read. It would be easier if you added a

```python
if cache.has_result(key):
    return
```

above the old code instead of indenting everything.

tests/cache/test_urlqueue.py
```diff
@@ -0,0 +1,174 @@
...
+        with self.assertRaises(ValueError):
+            UrlQueue(max_allowed_urls=0)
+        with self.assertRaises(ValueError):
+            UrlQueue(max_allowed_urls=-1)
+
+    def test_qsize(self):
+        """ Test qsize() """
+        self.assertEquals(self.urlqueue.qsize(), 0)
+        self.urlqueue.put(self.urldata1)
+        self.assertEquals(self.urlqueue.qsize(), 1)
```
mgedmin: Very minor comment: in the future please use `assertEqual` instead of `assertEquals`.  The latter is [deprecated](https://docs.python.org/3/library/unittest.html#deprecated-aliases).

tests/cache/test_urlqueue.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_empty(self):
+        """ Test empty() """
+        self.assertEquals(self.urlqueue.empty(), True)
+        self.urlqueue.put(self.urldata1)
+        self.assertEquals(self.urlqueue.empty(), False)
+
+    def test_get_empty(self):
+        """ Test, that get() with empty queue throws Empty """
+        with self.assertRaises(Empty):
+            self.assertEquals(self.urlqueue.get(0.01), None)
```
mgedmin: If that's a timeout, maybe make it `0.001` or even smaller?  No need to sleep for 10 ms.

Would `0` work?

tests/cache/test_urlqueue.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def test_put_get(self):
+        """
+        Test, that after put() we can get()
+        the item and it can be get only once
+        """
+        self.urlqueue.put(self.urldata1)
+        cached_item = (
+            self.urldata1.aggregate.result_cache.get_result(self.urldata1)
+        )
```
mgedmin: You could use `self.result_cache` here for a shorter expression.

tests/cache/test_urlqueue.py
mgedmin: (The same applies to other places that call `get(0.01)` too, assuming I guessed right and this does cause a 10ms sleep.)

PetrDlouhy:
@mgedmin Here are more tests enabled, than in original project. I must explore what is failing in the new tests and fix it.
Thank you very much for your comments, I will use your suggestions and rebase to master.

seamang:
Is there any reason this pull request is stalled now? This is an important improvement it would be good to get through, but I'm not clear whether @mgedmin's requests are now satisfied.

mgedmin:
Sorry, I dropped the ball!  Will have a look later today.

linkcheck/cache/urlqueue.py
```diff
@@ -120,7 +120,9 @@ def _put (self, url_data):
         log.debug(LOG_CACHE, "queueing %s", url_data.url)
         key = url_data.cache_url
         cache = url_data.aggregate.result_cache
-        if url_data.has_result or cache.has_result(key):
+        if cache.has_result(key):
+                return
```
mgedmin: The return seems to be indented twice as much as it needs to be.

(Perhaps an issue with tabs vs spaces that GitHub hides from me?)

anarcat: seems minor enough that we can just merge. :)

mgedmin:
Thank you for your patience, and thank you for reminding me to look at this again!

PR #55, opened 2017-03-30
Fixed bug with cookies loaded in via cookiefile that resulted in requ…

jeffyuhaoliu:
…est having an empty cookie

Previously, it looks like `session.cookies = requests.cookies.merge_cookies(session.cookies, cookie)` tries to merge RequestsCookieJar class with a list. This ends up not setting any cookies to requests Session object.


jeffyuhaoliu:
Closed pull request to harden down code.

PR #56, opened 2017-03-30
Fixed bug with cookies loaded in via cookiefile that resulted in requ…

jeffyuhaoliu:
…est having an empty cookie

Previously, it looks like session.cookies = requests.cookies.merge_cookies(session.cookies, cookie) tries to merge RequestsCookieJar class with a list. This ends up not setting any cookies to requests Session object.

This is another attempt from previously closed pull request. (#55)

### DIFF
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -21,6 +21,7 @@ import threading
 import thread
 import requests
 import time
+import cookielib
 try:
     import urlparse
 except ImportError:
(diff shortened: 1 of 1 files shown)


mgedmin:
I would feel more confident about this if there was a regression test for this bug.

anarcat:
can we get a unit test here? @jeffyuhaoliu can you clarify how we can reproduce this? ideally, could you also patch the test suite to include a test case?

thanks!

cjmayo:
Fixed in 9.4 by #133.

Issue #76, opened 2017-06-27
fails to work with newer requests releases


anarcat:
as partially described in #74, linkchecker seems to fail when using requests releases newer than 2.15. here's a backtrace:

```
  File "/usr/lib/python2.7/site-packages/linkcheck/checker/httpurl.py", line 197, in _get_ssl_sock
    line: if raw_connection.sock is None:
    locals:
      raw_connection = <local> None
      raw_connection.sock = <local> !AttributeError: 'NoneType' object has no attribute 'sock'
      None = <builtin> None
AttributeError: 'NoneType' object has no attribute 'sock'
```

PR #86, opened 2017-10-03
workaround new limitation in requests

anarcat:
newer requests do not expose the internal SSL socket object so we
cannot verify certificates. there was work to allow custom
verification routines which we could use, but this never finished:

https://github.com/shazow/urllib3/pull/257

so right now, just treat missing socket information as if the cert was
missing.

Closes: #76

### DIFF
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -194,6 +194,10 @@ class HttpUrl (internpaturl.InternPatternUrl, proxysupport.ProxySupport):
         """Get raw SSL socket."""
         assert self.scheme == u"https", self
         raw_connection = self.url_connection.raw._connection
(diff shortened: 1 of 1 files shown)


anarcat:
note that this would introduce a regression whereas we cannot verify SSL expiry times with newer requests versions. i'm not sure how to workaround that issue, but at least we would run properly again...

mgedmin:
Perhaps we should open a `requests` issue asking for an API so we can do this?

anarcat:
i thought https://github.com/shazow/urllib3/pull/257 and https://github.com/kennethreitz/requests/issues/1966 were the upstream issues, but feel free to open another one...

mgedmin:
Ah, I didn't see the link to kennethreitz/requests#1966 before.

Issue #44, opened 2017-03-08
support $XDG_CONFIG_HOME / $XDG_DATA_HOME


legrostdg:
It would be great if linkchecker could support $XDG_CONFIG_HOME / $XDG_DATA_HOME.

anarcat:
yeah, it sure would! :)

legrostdg:
fixed by #87

legrostdg:
Could we expect a new release anytime soon?

anarcat:
see #4

PR #87, opened 2017-10-17
use xdg dirs for config & data

legrostdg:
~/.linkchecker is used instead of the xdg equivalents if the directory
exists (backward compatibility).

### DIFF
--- a/linkcheck/configuration/__init__.py
+++ b/linkcheck/configuration/__init__.py
@@ -365,13 +365,22 @@ class Configuration (dict):
                         pass
 
 
+def get_user_data():
+    """Get the user data folder.
+    Returns "~/.linkchecker/" if this folder exists, \
+    "$XDG_DATA_HOME/linkchecker" if it does not.
+    @rtype string
+    """
+    homedotdir = normpath("~/.linkchecker/")
+    userdata = homedotdir if os.path.isdir(homedotdir) \
+        else os.path.join(xdg_data_home, "linkchecker")
+    return userdata
+
 def get_plugin_folders():
     """Get linkchecker plugin folders. Default is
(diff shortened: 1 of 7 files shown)


legrostdg:
fixes #44

anarcat:
hmm... this looks good, but why does it make travis fail? https://travis-ci.org/linkcheck/linkchecker/builds/289126425?utm_source=github_status&utm_medium=notification

legrostdg:
Should be OK now :-)

linkcheck/configuration/__init__.py
```diff
@@ -365,9 +366,12 @@ def sanitize_ssl(self):
 
 
 def get_plugin_folders():
-    """Get linkchecker plugin folders. Default is ~/.linkchecker/plugins/."""
+    """Get linkchecker plugin folders. Default is
+    $XDG_DATA_HOME/linkchecker/plugins/."""
```
mgedmin: Maybe mention in the docstring that ~/.linkchecker/plugins/ is also supported for backwards compatibility, and what happens if both exist?

linkcheck/logger/blacklist.py
```diff
@@ -33,7 +34,7 @@ class BlacklistLogger (_Logger):
     LoggerName = "blacklist"
 
     LoggerArgs = {
-        "filename": "~/.linkchecker/blacklist",
+        "filename": os.path.join(xdg_data_home, "linkchecker", "blacklist"),
```
mgedmin: No backwards compat here?

anarcat: could we fix this please?

anarcat:
we do have a changelog, in doc/changelog.txt.

PR #100, opened 2017-10-30
Fix fragment identifier quoting

### LINKED ISSUE #96: To much quoting of URL fragments
I have several documents, where the fragment identifier of the URL contains colons (':'), which linkchecker claims to be "not normalized".
According to my reading of RFC 3986 ':' is allowed, but the specification is quiet unclear if the character must be escaped, as ':' is in the "reserved" set.

While researching this issue I found https://jazzy.id.au/static/fragment-encoding-test.html, which indicates that escaping as few as possible seems to be a good thing for browser compatibility.

http://pythonhosted.org/uritools/ also does not escape the ':'.

Maybe the following patch is in order:

``` patch
From fd5139e6cbbacc13591d957887429a13e8163abe Mon Sep 17 00:00:00 2001
Message-Id: <fd5139e6cbbacc13591d957887429a13e8163abe.1509350231.git.hahn@univention.de>
From: Philipp Hahn <hahn@univention.de>
Date: Mon, 13 Jul 2015 10:24:44 +0200
Subject: [PATCH] Fix fragment identifier quoting
Organization: Univention GmbH, Bremen, Germany

According to <https://tools.ietf.org/html/rfc3986>:
 fragment    = *( pchar / "/" / "?" )
 pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
 unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
 pct-encoded = "%" HEXDIG HEXDIG
 sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
---
 linkcheck/url.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/linkcheck/url.py b/linkcheck/url.py
index ad53cb43..60107d00 100644
--- a/linkcheck/url.py
+++ b/linkcheck/url.py
@@ -338,7 +338,7 @@ def url_norm (url, encoding=None):
     urlparts[2] = url_quote_part(urlparts[2], safechars=_nopathquote_chars, encoding=encoding) # path
     if not urlparts[0].startswith("feed"):
         urlparts[2] = url_fix_wayback_query(urlparts[2]) # unencode colon in http[s]:// in wayback path
-    urlparts[4] = url_quote_part(urlparts[4], encoding=encoding) # anchor
+    urlparts[4] = url_quote_part(urlparts[4], safechars="!$&'()*+,-./;=?@_~", encoding=encoding) # anchor
     res = urlunsplit(urlparts)
     if url.endswith('#') and not urlparts[4]:
         # re-append trailing empty fragment
-- 
2.11.0
```

Original issue: <https://github.com/wummel/linkchecker/issues/607>

pmhahn:
According to <https://tools.ietf.org/html/rfc3986>:

    fragment    = *( pchar / "/" / "?" )
    pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
    unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
    pct-encoded = "%" HEXDIG HEXDIG
    sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

Fixes #96

### DIFF
--- a/linkcheck/url.py
+++ b/linkcheck/url.py
@@ -338,7 +338,7 @@ def url_norm (url, encoding=None):
     urlparts[2] = url_quote_part(urlparts[2], safechars=_nopathquote_chars, encoding=encoding) # path
     if not urlparts[0].startswith("feed"):
         urlparts[2] = url_fix_wayback_query(urlparts[2]) # unencode colon in http[s]:// in wayback path
-    urlparts[4] = url_quote_part(urlparts[4], encoding=encoding) # anchor
+    urlparts[4] = url_quote_part(urlparts[4], safechars="!$&'()*+,-./;=?@_~", encoding=encoding) # anchor
     res = urlunsplit(urlparts)
     if url.endswith('#') and not urlparts[4]:
         # re-append trailing empty fragment
--- a/tests/checker/data/http.html.result
+++ b/tests/checker/data/http.html.result
@@ -67,7 +67,7 @@ valid
 
 url http://example.org/foo/ #a=1,2,3
 cache key http://example.org/foo/%%20
-real url http://example.org/foo/%%20#a%%3D1%%2C2%%2C3
+real url http://example.org/foo/%%20#a=1,2,3
 error
 
 url http://.example.org/
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -172,14 +172,16 @@ class TestUrl (unittest.TestCase):
         nurl = "http://example.com:81/"
         self.urlnormtest(url, nurl)
 
-    def test_norm_fragment (self):
+    def test_norm_fragment_empty (self):
         # Test url norm fragment preserving.
         # Empty fragment identifiers must be preserved:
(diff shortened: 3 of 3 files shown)


pmhahn:
@anarcat Each layer above linkcheck needs to handle its own quoting, as you can not know which layer is the next: Is it a UNIX shell, a windows shell, or a SQL database, which all have different quoting requirements.
Yes, $();& have a special meaning with (ba)sh, but if you're going to pipe the output of linkcheck to any shell for execution, you're doomed anyway.

Also take a look at linkcheck/url.py:453-456
```python
_nopathquote_chars = "-;/=,~*+()@!"
_nopathquote_chars += "|"
_safe_url_chars = re.escape(_nopathquote_chars + "_:.&#%?[]!")+"a-zA-Z0-9"
```
In other parts of an URL those characters are already allowed.
So I consider this not an issue for linkcheck.

anarcat:
alright then, LGTM. :)

PR #103, opened 2017-11-02
Install .mo files correctly

### LINKED ISSUE #102: linkchecker can't find localization files
When I call linkchecker on the command line, it exits before doing anything with this error message on trying to access its localization files:
```
$> linkchecker
  File "/usr/bin/linkchecker", line 30, in <module>
    import linkcheck
  File "/usr/lib/python2.7/site-packages/linkcheck/__init__.py", line 137, in <module>
    init_i18n()
  File "/usr/lib/python2.7/site-packages/linkcheck/__init__.py", line 124, in init_i18n
    i18n.init(configdata.name.lower(), locdir, loc=loc)
  File "/usr/lib/python2.7/site-packages/linkcheck/i18n.py", line 94, in init
    install_language(default_language)
  File "/usr/lib/python2.7/site-packages/linkcheck/i18n.py", line 100, in install_language
    languages=[get_lang(language)], fallback=True)
  File "/usr/lib/python2.7/site-packages/linkcheck/i18n.py", line 110, in get_translator
    languages=languages, class_=translatorklass, fallback=fallback)
  File "/usr/lib/python2.7/gettext.py", line 553, in translation
    with open(mofile, 'rb') as fp:
IOError: [Errno 21] Is a directory: u'/usr/share/locale/de/LC_MESSAGES/linkchecker.mo'
```

The problem seems to be in the directory structure created by `setup.py`:
```
/usr/share/locale/
├── fr/
│   └── LC_MESSAGES/
│       └── linkchecker.mo/
│           └── fr.mo
├── es/
│   └── LC_MESSAGES/
│       └── linkchecker.mo/
│           └── es.mo
└── de/
    └── LC_MESSAGES/
        └── linkchecker.mo/
            └── de.mo
```
I assume the lowest level of the tree is created accidentally, but I did not understand what causes this behaviour. Workaround is to start linkchecker with `LC_ALL=C`.

version: linkchecker 9.4 from [AUR](https://aur.archlinux.org/packages/linkchecker/)
python 2.7
setuptools 36.2.7-1

(Originally reported at https://github.com/wummel/linkchecker/issues/728 )

mgedmin:
Fixes #102.

To clarify: this will not help people using `pip install linkchecker` (even if we get PyPI access), but it should help people running `python setup.py install`, such as some distro packages (AUR).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,8 +173,36 @@ class MyInstallData (install_data, object):
 
     def run (self):
         """Adjust permissions on POSIX systems."""
+        self.install_translations()
         super(MyInstallData, self).run()
         self.fix_permissions()
+
+    def install_translations (self):
+        """Install compiled gettext catalogs."""
+        # A hack to fix https://github.com/linkcheck/linkchecker/issues/102
+        i18n_files = []
+        data_files = []
+        for dir, files in self.data_files:
+            if 'LC_MESSAGES' in dir:
+                i18n_files.append((dir, files))
+            else:
+                data_files.append((dir, files))
+        self.data_files = data_files
+        # We do almost the same thing that install_data.run() does, except
+        # we can assume everything in self.data_files is a (dir, files) tuple,
+        # and all files lists are non-empty.  And for i18n files, instead of
+        # specifying the directory we instead specify the destination filename.
+        for dest, files in i18n_files:
+            dest = util.convert_path(dest)
+            if not os.path.isabs(dest):
+                dest = os.path.join(self.install_dir, dest)
+            elif self.root:
+                dest = util.change_root(self.root, dest)
+            self.mkpath(os.path.dirname(dest))
+            for data in files:
+                data = util.convert_path(data)
+                (out, _) = self.copy_file(data, dest)
+                self.outfiles.append(out)
 
     def fix_permissions (self):
         """Set correct read permissions on POSIX systems. Might also


setup.py
```diff
@@ -173,9 +173,33 @@ class MyInstallData (install_data, object):
...
+        """Install compiled gettext catalogs."""
+        # A hack to fix https://github.com/linkcheck/linkchecker/issues/102
+        i18n_files = [
+            (dir, files) for (dir, files) in self.data_files
+            if 'LC_MESSAGES' in dir
+        ]
+        self.data_files = [
+            (dir, files) for (dir, files) in self.data_files
+            if 'LC_MESSAGES' not in dir
+        ]
```
mgedmin: Copying comment over from https://github.com/linkcheck/linkchecker/commit/d94bd0ee0660a989eb49d4179ed1db7af9c766d9#commitcomment-25354512:
> at the very least we should remove duplication between those two initializers which are identical.

The only other clean-ish way to split the list into two that I can come up with is
```python
i18n_files = []
data_files = []
for dir, files in self.data_files:
    if 'LC_MESSAGES' in dir:
        i18n_files.append((dir, files))
    else:
        data_files.append((dir, files))
self.data_files = data_files
```

and I'm not sure it's an improvement.  What do you think?

setup.py
```diff
@@ -173,9 +173,33 @@ class MyInstallData (install_data, object):
...
+        for dir, files in i18n_files:
+            dir = util.convert_path(dir)
+            if not os.path.isabs(dir):
+                dir = os.path.join(self.install_dir, dir)
+            elif self.root:
+                dir = util.change_root(self.root, dir)
+            self.mkpath(os.path.dirname(dir))  # <-- the magic happens here
```
mgedmin: From https://github.com/linkcheck/linkchecker/commit/d94bd0ee0660a989eb49d4179ed1db7af9c766d9#commitcomment-25354514:
> expand. what magic? what does this do? what problem does it solve?

This is essentially copied and pasted from [distutils.command.install_data](https://github.com/python/cpython/blob/b694770a2b23cd485c98bf673a8b2dc1a865d9df/Lib/distutils/command/install_data.py#L45-L75), except I pruned some branches (`if isinstance(f, str)`, since f is always a tuple for us, and `if f[1] == []`, since we're not creating empty directories here), changed the outer loop variables to destructure tuples, and changed one line of actual code.  The "magic" comment pinpoints that line: distutils's install_data.run() does a self.mkpath(dir) at this point, and we do `self.mkdir(os.path.dirname(dir))`.  Here's why:

For .mo files `dir` is something like `/usr/local/share/locale/fr/LC_MESSAGES/linkchecker.mo`, and `files` is `['po/fr.mo']`.  We want to `mkdir /usr/local/share/locale/fr/LC_MESSAGES` and copy `po/fr.mo` into `/usr/local/share/locale/fr/LC_MESSAGES/linkchecker.mo`.  `self.copy_file(src, dst)` behaves like shell's `cp`: if dst is a directory, it creates a file with the same name under it, and if it's not, it's taken as the destination filename.  So 

    self.copy_file('po/fr.mo', '/usr/local/share/locale/fr/LC_MESSAGES/linkchecker.mo')

will do what we want *as long as `/usr/local/share/locale/fr/LC_MESSAGES/linkchecker.mo`* is not a directory.

setup.py
anarcat: that sure looks better. i see now why it's two different loops. it does seem strange to recreate the `data_files` list here - it seems to me it would be better to generate it properly from the start...

setup.py
anarcat: don't tell me, say so in the code. :)

that said:

> The "magic" comment pinpoints that line: distutils's install_data.run() does a self.mkpath(dir) at this point, and we do self.mkdir(os.path.dirname(dir)).

don't we do `self.mkpath`?

and why don't we just call `copy_file()` ourselves? if we're going to coerce `install_data()` to try and fit a square peg in a round hole, why not just add a new `install_trans` command that would simply install translations in the expected location? because at this point it seems like we're reinventing `install_data` in everything but the name...

this, in turn, could eventually be shared with other projects more cleanly...

anarcat: oh wait we do call copy_file, nevermind. :)

anarcat:
ping any progress here?

mgedmin:
I wasn't sure if I should clean up this approach, or study the links you pasted and find a better way.  When I'm not sure, I procrastinate, and then forget about things...

What do you suggest?

anarcat:
let's prioritize cleaning up the code then. we can fix this the "right way" later, if we need to. just document the heck out of the code - the answers you sent me, add them as comments in the code. then let's just merge this...

anarcat:
gah. not sure i like this "update branch" thing - when we merge *one* pull request, we need to add a merge commit to *all* the other pending merge requests... i think i'll disable that now.

mgedmin:
I've updated the code, added a comment, and rebased on master.

PR #114, opened 2017-12-22
Support atom web feeds

reinholdfuereder:
This allows parsing and thus link checking Atom web feeds

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -90,6 +90,8 @@ class UrlBase (object):
         # It does not harm other URL schemes like HTTP since HTTP servers
         # should not send this content type. They send text/html instead.
         "application/x-httpd-php": "html",
+        # Allow parsing RSS feeds.
+        "application/atom+xml": "html",
         "text/css": "css",
         "application/x-shockwave-flash": "swf",
         "application/msword": "word",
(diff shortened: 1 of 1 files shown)


mgedmin:
Isn't the HTML content in an Atom feed encoded?  E.g. if I do view source on my blog's Atom feed, I see things like

```
&lt;p&gt;Ansible has &lt;a href=&#34;https://docs.ansible.com/ansible/intro.html&#34;&gt;great documentation&lt;/a&gt;,
```

reinholdfuereder:
@mgedmin First of all sorry for the delay. I was just discussing your comment/concern ad encoding with my colleague who implemented the Atom feed support, and here is the outcome:
Yes, you are right, it is using special encoding; and using `--debug=all` linkchecker cli argument shows that we are in a way just very lucky (details below) that this little change (pretending the Atom feed is HTML) is sufficient for our needs and that this is admittedly definitely not a full blown general solution though...

**Why are we lucky?**
Because with this change linkchecker checks exactly the feed's URLs of interest to us, and really nothing else; namely the URLs in `<link>` elements of `<entry>` elements. According to https://validator.w3.org/feed/docs/atom.html `<link>` elements are "only" recommended, whereas the sibling `<id>` elements are mandatory. And luckily in our case the are always pointing to the same URL...

anarcat:
okay, so if i understand this correctly, we would be "lying" by claiming RSS/Atom support at this stage? we'd certainly need to be parsing the feeds, and that's a whole different bag of tricks... i'm tempted to simply close this PR before we clarify what works and doesn't work in here...

reinholdfuereder:
@anarcat Yes, I would agree with your message => closing

PR #118, opened 2018-01-06
Remove third party libraries from repository

PetrDlouhy:
This removes third party libraries from the sources and uses them as dependency.


PetrDlouhy:
OK, I've added `dnspython` to `install_requires`.

anarcat:
okay well, i've merged this. fire in the hole! :)

Issue #26, opened 2017-02-05
Consider using BeautifulSoup instead of built-in HTML parser?


astrofrog:
I wonder whether it would be worth considering using an existing HTML parser such as BeautifulSoup to avoid having to include C code in the linkchecker package? This might lower the maintenance burden in the long term (since keeping C extensions working across platforms is not trivial).

anarcat:
i absolutely agree. we have tons of duct tape as well to parse that HTML that gives us weird results,  #23 for example.

seamang:
Definitely. I need an immediate fix for #23 just to be able to use linkchecker, but what we've added now is an ugly patch, and I can see us needing to add more of these with the current homegrown parser.

PetrDlouhy:
Much of my work in #40 is related to the HTML parser and there are still two remaining problems with that which cause failed tests on Python 3 and I am unable to solve them right now.

There would be ton of special cases, which are properly solved in more widely used parser, that might not be solved in the build-in parser.

So I think, that it would be huge benefit, if it gets implemented.

PetrDlouhy:
I have worked on this. See #119. It would require extensive testing an possibly some improvements, though.

cjmayo:
Good idea! Thanks Petr for showing it was possible.
Done.

PR #119, opened 2018-01-07
Htmlparser beatifulsoup

PetrDlouhy:
PR for #26

This completely removes the build-in parser and replaces it with BeatifulSoup.
The current implementation is mocking the API of the old parser. It means, that if somebody would like to use the old parser (maybe it is faster?), he could make standalone project from that and easily change the libraries.

So far this approach works, but there are few remaining failing tests and it would require a bit of cleaning.


PetrDlouhy:
I have updated this.
There are remaining problems:

- The failed tests are probably all due to differences between the parsers. But it would need to be checked and tested and the tests changed (or some removed in case of `test_parser.py`).
- I was not possible to get line numbers and position for the tag. As I was searching, `BeautifulSoup` doesn't provide this kind of information.

mgedmin:
I'd be very happy to see the custom C parser replaced with something external and maintained.

> I was not possible to get line numbers and position for the tag. As I was searching, BeautifulSoup doesn't provide this kind of information.

That's sad :(

Are there other HTML parsers we could consider?  Does lxml.html provide line number information?

Can we file a wishlist bug request to BeautifulSoup, or maybe even consider a PR?

Can we live without line numbers?  Search exists, as long as we know http://example.com is a broken link on some_page.html, we can find it.

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+class Parser(object):
+    html_doc = u""
+    handler = None
+    encoding = "iso8859-1"
+
+    def __init__(self, handler):
+        self.handler = handler
+
+    def feed(self, feed_text):
+        if type(feed_text) == bytes:
```
mgedmin: `if isinstance(feed_text, bytes):` would be more idiomatic.

I really hope the caller of this API won't ever feed us partial UTF-8 characters!

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+            feed_text = feed_text.decode(self.encoding)
+        self.html_doc += feed_text
+
+    def reset(self):
+        self.html_doc = u""
+
+    def flush(self):
+        soup = BeautifulSoup(self.html_doc, 'html.parser')
```
mgedmin: Oh, how about this instead: make self.html_doc a StringIO(), write to it in feed(), and here do html_doc = self.html_doc.getvalue(), and then do the if isinstance(html_doc, bytes): html_doc = html_doc.decode(self.encoding)?

Also, what if the document encoding is not actually ISO-8859-1?

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+        for tag in soup.find_all():
+            attrs = ListDict()
+            for k, v_list in tag.attrs.items():
+                if type(v_list) != list:
```
mgedmin: `if not isinstance(v_list, list):`

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+                    v_list = [v_list]
+                for v in v_list:
+                    if type(v) == str:
```
mgedmin: `if isinstance(v, bytes):`

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+                self.handler.characters(tag.text)
+            if (not tag.is_empty_element and
+                hasattr(self.handler, 'end_element')):
+                self.handler.end_element(tag.name)
+            if tag.comments:
+                for comment in tag.comments:
+                    self.handler.comment(comment)
+
+    def debug(self, text):
+        raise NotImplementedError("number is not implemented")
```
mgedmin: `number` -> `debug`?

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -0,0 +1,86 @@
...
+        return 0
+
+    def pos(self, text):
+        return 0
+
+    def peek(self, peek_len):
+        raise NotImplementedError("Peeking is deprecated in favor of 'handler.start_element' attribute 'element_text'")
+
+
+def parser(handler = None):
```
mgedmin: PEP-8 says no spaces around `=` here.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 # required:
+bs4
```
mgedmin: This also needs to go into `install_requires` in `setup.py`.

setup.py
```diff
@@ -463,7 +463,6 @@ def get_file_list (self):
         'linkcheck.configuration',
         'linkcheck.director',
         'linkcheck.htmlutil',
-        'linkcheck.HtmlParser',
```
mgedmin: This sort of thing is greatly simplified by using `setuptools.find_packages()`.

anarcat:
just a little note to send a big cheer down that lane: great initiative, and i hope we can make this work. it's utterly reckless to have our own C-based HTML parser...

In fact, I often wondered if we should just reuse those Javascript-based engines like Phantom or [whatever that DD ended up with](https://piware.de/post/2017-12-21-phantomjs-to-chromium/). ;)

can't we file a bug against beautiful soup to get line numbers reported or is that something completely out of scope there? progressive enhancement is good..

there's a conflict on the travis config file here, should be easy to fix, and i otherwise agree with the current review.

keep up the good work!

PetrDlouhy:
I filled the issue here: https://bugs.launchpad.net/beautifulsoup/+bug/1742921

I also tried to implement this with Python's `HTMLParser` here: https://github.com/PetrDlouhy/linkchecker-1/commit/b013538f8e23eafaa7a928407a9502e750694cbd
The implementation is even more straight-forward, but I wasn't able to get past the peeking. The `peek` function is speciality of the built in parser, and it's behaviour is not easy to simulate and couldn't be overpassed as easy as with `BeautifulSoup`.

I am not sure about the encoding handling either. I think, we should make some tests focused on the encoding and on the line numbers prior to pulling this.

linkcheck/HtmlParser/htmllib.py
```diff
@@ -186,7 +178,6 @@ def characters (self, data):
         @type data: string
         @return: None
         """
-        data = data.encode(self.encoding, "ignore")
```
anarcat: i agree encoding is an issue. we don't want to start crashing on weird encodings we find out there. and we don't need to be super fancy - from what i can tell from the previous source, we just drop weird chars... in fact, i wonder why those encoding lines are dropped here in the first place... isn't that where we would fix the encoding?

PetrDlouhy:
So I think, I solved the encoding. I removed the decoding parts and it is stored now in bytes (I will try to rewrite it to be `BytesIO`).
It seems, that now it works properly unlike in original Linkchecker. The charset tests, that are separately in #125 are passing with this implementation!

anarcat:
after getting confirmation that weird encodings don't even work now (#125), i retract my opposition to this PR. i think the concerns of @mgedmin still stand, but at least we shouldn't block on encoding.

thanks for doing all this research!

PetrDlouhy:
I made several corrections for this.
I fixed all tests, but I had to change `test_parser` massively to achieve this. The new parser is working different on edge cases, but I hope, closer to browsers in many cases. Anyway, this would require careful review to ensure, that it works properly.

This also doesn't include line numbers reporting or some other reference (except the link name and url, which Linkchecker gives).

mgedmin:
I've no energy to review this today.  Please ping me if I forget to come back to it.

anarcat:
@mgedmin ping :)

in retrospect, i think i'd be tempted to be more lax with a new beautifulsoup parser than anything else: rolling our own parser (written in C, nothing less!) is a really bad idea and i would be quite glad to get rid of it, even if it means a few more warts.

anarcat:
@PetrDlouhy i know it's been a long time, but @mgedmin finally reviewed your work, and while we could merge as is, it seems there are minor tweaks that would improve the result... could you take a look?

(i took the liberty of resolving conflicts after #147 - that could be rebased to remove the merge commit...)

mgedmin:
We still have one test failure because of the missing line numbers.

anarcat:
do we have an upstream issue for the line number stuff?

I would be ready to drop that feature in favor of having a non-C, non-housemade parser. what do you think @mgedmin ?

mgedmin:
> do we have an upstream issue for the line number stuff?

I don't know.  I never filed one.

> I would be ready to drop that feature in favor of having a non-C, non-housemade parser. what do you think @mgedmin ?

Oh, yes!  Only I'd like to do it properly (i.e. remove any line numbers from output instead of printing `0` in their place if we currently do so), and somebody ought to fix the test failure before we can merge this.

I don't have the energy currently to push this (or any other linkchecker issue), but I will try to provide code reviews.

anarcat:
update: upstream doesn't plan on adding the feature, but is open to a pull request.

so the next step here is either to rip out the line number code or patch beautifulsoup to include those.

@PetrDlouhy are you up for this?

cjmayo:
> Are there other HTML parsers we could consider? Does lxml.html provide line number information?
> 
> Can we file a wishlist bug request to BeautifulSoup, or maybe even consider a PR?
> 
> Can we live without line numbers? Search exists, as long as we know http://example.com is a broken link on some_page.html, we can find it.

lxml Element has a property `sourceline`:
> Original line number as found by the parser or None if unknown. 
https://lxml.de/api/lxml.etree._Element-class.html#sourceline

Don't know if that would work.

anarcat:
> Don't know if that would work.

that would be awesome, but at this point I think we're ready to ditch our homegrown parser in favor of beautifulsoup even if it means losing line numbers. it's too much of a liability.

cjmayo:
Beautiful Soup is using somebody else's parser, optionally lxml:
https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser
So, maybe we can have both.

cjmayo:
All the commits and code comments plus line number support are addressed in #337.

PR #120, opened 2018-01-08
Skip colon normalization to avoid 404 errors

hydrapolic:
https://github.com/linkcheck/linkchecker/issues/113


mgedmin:
Two tests are failing.

hydrapolic:
Yes,the PR fixed one thing,but broke the test. I don't really know what is the proper way to make it work for every use case.

PR #121, opened 2018-01-08
Execute parser test by parametrized

PetrDlouhy:
Separate parser tests, so it is easy to determine which are failing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ usedevelop = true
 deps =
     -rrequirements.txt
     pyftpdlib
+    parameterized
     pytest-xdist
     pytest-cov
 commands =
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -23,6 +23,8 @@ import linkcheck.HtmlParser.htmllib
 from cStringIO import StringIO
 import unittest
 
+from parameterized import parameterized
+
 
 # list of tuples
 # (<test pattern>, <expected parse output>)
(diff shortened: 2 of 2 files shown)


linkcheck/configuration/__init__.py
```diff
@@ -403,7 +403,7 @@ def make_userdir(child):
             # Windows forbids filenames with leading dot unless
             # a trailing dot is added.
             userdir += "."
-        os.mkdir(userdir, 0700)
+        os.makedirs(userdir, 0700)
```
mgedmin: Wasn't this part of a separate PR?

PetrDlouhy:
@mgedmin The `os.makedirs()` got there probably because the `master` branch changed. Now it is rebased.

tox.ini
```diff
@@ -6,6 +6,7 @@ usedevelop = true
 deps =
     -rrequirements.txt
     pyftpdlib
+    parameterized
```
anarcat: why do we need an extra dependency here? pytest has paremetrized tests already...

PetrDlouhy: The `parametrized` from `pytest` can't be used on class based tests, or at least I wasn't able to use it or find any solution.

anarcat: hmmm... i don't have experience with this myself, but it looks like it's supposed to work according to [docs](https://docs.pytest.org/en/latest/example/parametrize.html#a-quick-port-of-testscenarios). also, the parametrized module docs say:

> Why not use @pytest.mark.parametrize? Because spelling is difficult. Also, parameterized doesn't require you to repeat argument names, and (using param) it supports optional keyword arguments.

"spelling is hard" is not really a good argument in my opinion... ;)

PetrDlouhy: I don't get it, should I implement the `pytest_generate_tests`? I wasn't able to figure out how is that supposed to work after looking into the docs for quite a long time. And I think, it affect all methods on the class.
I think, that such implementation would very negatively affect code readability.

The `@pytest.mark.parametrize` decorator doesn't work on class methods.

And the dependency is only testing dependency. It shloudn't be part of the distribution package, so I think it is no such a big deal.

mgedmin: @anarcat: [py.test documentation](https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses) states:
> The following pytest features **do not** work, and probably never will due to different design philosophies:
> ...
> - Parametrization

mgedmin: Personally, I'd rather see an extra test-only dependency than a mysterious pytest generation function.

anarcat: ah. ouch. well nevermind then, i retract my objection. :)

anarcat:
sorry for the delay, @PetrDlouhy

PR #123, opened 2018-01-12
Add Python3 support - easiest changes

PetrDlouhy:
This is set of Python 3 compatibility changes.
It is the easiest changes - syntax and imports mostly.

### DIFF
--- a/linkcheck/checker/proxysupport.py
+++ b/linkcheck/checker/proxysupport.py
@@ -17,12 +17,14 @@
 """
 Mixin class for URLs that can be fetched over a proxy.
 """
-import urllib
-try:
-    import urlparse
+try:  # Python 3
+    from urllib import parse
+    from urllib import request
+    from urllib.parse import splitport
 except ImportError:
-    # Python 3
-    from urllib import parse as urlparse
+    from urllib import splitport
+    import urllib as request
+    import urlparse as parse
 import os
 from .. import LinkCheckerError, log, LOG_CHECK, url as urlutil, httputil
 
@@ -40,7 +42,7 @@ class ProxySupport (object):
         self.proxyauth = None
         if not self.proxy:
             return
-        proxyurl = urlparse.urlparse(self.proxy)
+        proxyurl = parse.urlparse(self.proxy)
         self.proxytype = proxyurl.scheme
         if self.proxytype not in ('http', 'https'):
             # Note that invalid proxies might raise TypeError in urllib2,
@@ -67,7 +69,7 @@ class ProxySupport (object):
 
     def ignore_proxy_host (self):
         """Check if self.host is in the $no_proxy ignore list."""
-        if urllib.proxy_bypass(self.host):
+        if request.proxy_bypass(self.host):
             return True
         no_proxy = os.environ.get("no_proxy")
         if no_proxy:
@@ -96,7 +98,7 @@ class ProxySupport (object):
 
 def parse_host_port (host_port):
     """Parse a host:port string into separate components."""
-    host, port = urllib.splitport(host_port.strip())
+    host, port = splitport(host_port.strip())
     if port is not None:
         if urlutil.is_numeric_port(port):
             port = int(port)
--- a/linkcheck/configuration/confparse.py
+++ b/linkcheck/configuration/confparse.py
@@ -16,7 +16,10 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Parse configuration files"""
 
-import ConfigParser
+try: # Python 3
+    from configparser import RawConfigParser
+except ImportError: # Python 2
+    from ConfigParser import RawConfigParser
 import os
 from .. import LinkCheckerError, get_link_pat, LOG_CHECK, log, fileutil, plugins, logconf
 
@@ -30,7 +33,7 @@ def read_multiline (value):
         yield line
 
 
-class LCConfigParser (ConfigParser.RawConfigParser, object):
+class LCConfigParser (RawConfigParser, object):
     """
     Parse a LinkChecker configuration file.
     """
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -18,14 +18,17 @@
 Aggregate needed object instances for checker threads.
 """
 import threading
-import thread
+try:  # Python 3
+    import _thread
+except ImportError:
+    import thread as _thread
+
 import requests
 import time
-try:
-    import urlparse
+try:  # Python 3
+    from urllib import parse
 except ImportError:
-    # Python 3
-    from urllib import parse as urlparse
+    import urlparse as parse
 import random
 from .. import log, LOG_CHECK, strformat, LinkCheckerError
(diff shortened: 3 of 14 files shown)


linkcheck/director/aggregator.py
```diff
@@ -18,7 +18,10 @@
 Aggregate needed object instances for checker threads.
 """
 import threading
-import thread
+try: # Python 3
+    import _thread as thread
+except ImportError:
+    import thread
```
mgedmin: In other modules you do `import thread as _thread` on Python 2.  I think I prefer that.

tests/__init__.py
```diff
@@ -288,9 +288,9 @@ def get_file (filename=None):
...
-    print "has msgfmt", has_msgfmt()
-    print "has POSIX", has_posix()
-    print "has proxy", has_proxy()
-    print "has X11", has_x11()
+    print("has clamav", has_clamav())
+    print("has network", has_network())
+    print("has msgfmt", has_msgfmt())
+    print("has POSIX", has_posix())
+    print("has proxy", has_proxy())
+    print("has X11", has_x11())
```
mgedmin: You want a `from __future__ import print_function` at the top of this file, otherwise you're printing tuples on Python 2.

tests/test_po.py
```diff
@@ -56,11 +56,8 @@ class TestGTranslator (unittest.TestCase):
     def test_gtranslator (self):
         """Test all pofiles for GTranslator brokenness."""
         for f in get_pofiles():
-            fd = file(f)
-            try:
+            with open(f, 'rb') as fd:
                 self.check_file(fd, f)
```
mgedmin: On Python 2 this changes newline interpretation, which doesn't matter for `check_file`.

On Python 3 this changes a NameError for `file` into a TypeError: startswith first arg must be bytes or a tuple of bytes, since `check_file()` assumes it'll get text, not bytes.

I suggest a couple of `b` string prefixes in check_file, and it's all good.

linkcheck/htmlutil/linkparse.py
```diff
@@ -82,11 +82,11 @@
...
 # matcher for <meta http-equiv=refresh> tags
-refresh_re = re.compile(ur"(?i)^\d+;\s*url=(?P<url>.+)$")
-_quoted_pat = ur"('[^']+'|\"[^\"]+\"|[^\)\s]+)"
-css_url_re = re.compile(ur"url\(\s*(?P<url>%s)\s*\)" % _quoted_pat)
+refresh_re = re.compile(r"(?i)^\d+;\s*url=(?P<url>.+)$")
+_quoted_pat = r"('[^']+'|\"[^\"]+\"|[^\)\s]+)"
+css_url_re = re.compile(r"url\(\s*(?P<url>%s)\s*\)" % _quoted_pat)
 swf_url_re = re.compile("(?i)%s" % urlutil.safe_url_pattern)
-c_comment_re = re.compile(ur"/\*.*?\*/", re.DOTALL)
+c_comment_re = re.compile(r"/\*.*?\*/", re.DOTALL)
```
mgedmin: This worries me a bit, since I remember that regexes in Python 2 have two modes (bytes vs unicode), but I don't remember how they differ exactly.

Let's hope our test coverage is adequate.

PetrDlouhy:
OK, I reworked this by @mgedmin suggestins and removed the regexp changes (I will try to add some charset tests first).

linkcheck/checker/proxysupport.py
```diff
@@ -17,12 +17,14 @@
...
-except ImportError:
-    # Python 3
+try: # Python 3
     from urllib import parse as urlparse
+    from urllib import request as urlrequest
+    from urllib.parse import splitport as urllib_splitport
+except ImportError:
+    from urllib import splitport as urllib_splitport
+    import urllib as urlrequest
+    import urlparse
```
anarcat: i feel this looks a little off... in the long term, shouldn't we be looking at using Python3 semantics? e.g. we would use `urllib.parse` instead of `urlparse`...

linkcheck/checker/proxysupport.py
```diff
@@ -96,7 +98,7 @@ def get_netloc(self):
 
 def parse_host_port (host_port):
     """Parse a host:port string into separate components."""
-    host, port = urllib.splitport(host_port.strip())
+    host, port = urllib_splitport(host_port.strip())
```
anarcat: i find this especially weird.

linkcheck/configuration/confparse.py
```diff
@@ -16,7 +16,10 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Parse configuration files"""
 
-import ConfigParser
+try: # Python 3
+    from configparser import RawConfigParser
+except ImportError: # Python 2
+    from ConfigParser import RawConfigParser
```
anarcat: ... whereas this is fine, for example.

linkcheck/director/aggregator.py
```diff
@@ -18,7 +18,10 @@
 Aggregate needed object instances for checker threads.
 """
 import threading
-import thread
+try: # Python 3
+    import _thread
+except ImportError:
+    import thread as _thread
```
anarcat: here's another example. above we do `from thread import error as thread_error` yet here we do `import thread as _thread`... it looks inconsistent - why don't we do the latter everywhere?

PetrDlouhy:
@anarcat I agree, that we should prefer Python 3 syntax. I have changed this, where possible (i.e. in case of `url.py` it is not possible, because it two locations in Python 2 are one location in Python 3).

requirements.txt
```diff
@@ -2,5 +2,6 @@
 requests<2.15,>=2.2
 pyxdg
 dnspython
+future
```
mgedmin: Actually, where is this being used?  You're not importing `future`, and you're not importing `builltins`, and you've try/except blocks for Python 2 vs Python 3 names in multiple places... What am I missing?

PetrDlouhy: `from __future__ import print_function`
But anyway, it will be required on many places in following Python 3 compatibility commits.

anarcat:
On 2018-01-15 05:30:16, PetrDlouhy wrote:
> +future
>
> `from __future__ import print_function`
> But anyway, it will be required on many places in following Python 3 compatibility commits.

I believe __future__ is in the standard library and doesn't require the
`future` third-party package.

PetrDlouhy:
OK, I removed the `future` import for this PR.

Makefile
```diff
@@ -32,8 +32,6 @@ TESTS ?= tests
 # set test options, eg. to "--verbose"
 TESTOPTS=
 PAGER ?= less
-# original dnspython repository module
-DNSPYTHON:=$(HOME)/src/dnspython-git
```
anarcat: shouldn't this have been done in #118? what's going on here?

PetrDlouhy:
@anarcat That was some mistake in rebasing. Now it is OK.

anarcat:
merged, whoohoo! thanks for your patience!

PR #125, opened 2018-01-12
(Not to be pulled) Charset tests

PetrDlouhy:
So about the encoding. It doesn't work in current implementation of Linkchecker at all.
I would expect, it translates the encoding from the original one to the encoding of the terminal. Or it translates the characters to the form, that one could easily read. (Like `Zlutoucky kun upel dabelske ody uuuu`). It does neither of that. It just outputs the link name in original form on the terminal, which will lead to unreadable characters.

This PR is not meant to be pulled, but it could serve as reference before implementing character handling fixes (i.e. during the parser change).

### DIFF
--- a/tests/checker/__init__.py
+++ b/tests/checker/__init__.py
@@ -200,8 +200,6 @@ class LinkCheckTest (unittest.TestCase):
 
     def fail_unicode (self, msg):
         """Print encoded fail message."""
-        # XXX self.fail() only supports ascii
-        msg = msg.encode("ascii", "replace")
         self.fail(msg)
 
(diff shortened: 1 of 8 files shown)


anarcat:
> So about the encoding. It doesn't work in current implementation of Linkchecker at all.

Ouch. Well, I guess that offsets my other concerns. ;) Good to know, thanks!!

cjmayo:
The charset tests commit is included in #337 (and has been very useful in developing that).

cjmayo:
Tests merged into master d1844a526ee712d. Thanks!

PR #126, opened 2018-01-14
Test for linenumbers and other parts of url_data

PetrDlouhy:
This adds a test for various parts of `url_data` variable in `Logger`.

I am not sure, if this works on all platforms.


anarcat:
ditto

PR #127, opened 2018-01-15
(WIP) add Appveyor configuration for Windows testing

PetrDlouhy:
I tried to add Windows testing. Unfortunately, `Travis-CI` doesn't support that. So I added configuration for `Appveyor`. Here are some results:
https://ci.appveyor.com/project/PetrDlouhy/linkchecker-1/history

There are some tests failing, so we will need to fix that before further development of Linkchecker to ensure it doesn't break anything on Windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,25 @@
+# AppVeyor.com is a Continuous Integration service to build and run tests under
+# Windows
+
+environment:
+    matrix:
+        - PYTHON_VERSION: "2.7"
+        - PYTHON_VERSION: "3.5"
+        - PYTHON_VERSION: "3.6"
+
+matrix:
+    allow_failures:
+        - PYTHON_VERSION: "3.5"
+        - PYTHON_VERSION: "3.6"
+
+install:
+    - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+    - "pip install tox coveralls"
+
+build: false
+
+test_script:
+    - "tox"
+
+after_test:
+  - "coveralls"
--- a/linkcheck/bookmarks/safari.py
+++ b/linkcheck/bookmarks/safari.py
@@ -16,8 +16,9 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import os
-import sys
 import plistlib
+import sys
+from xml.parsers.expat import ExpatError
 try:
     import biplist
     has_biplist = True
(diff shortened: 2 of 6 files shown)


appveyor.yml
```diff
@@ -0,0 +1,29 @@
+# AppVeyor.com is a Continuous Integration service to build and run tests under
+# Windows
+
+environment:
+
+    global:
+        CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
```
mgedmin: What does this do?  I don't see where %CMD_IN_ENV% is being used.

mgedmin:
The annoying thing about Appveyor is that repositories are tied to user accounts, and not to GitHub organisations.  AFAIU this means only one of the maintainers can tweak Appveyor settings (other than those settable via the appveyor.yml file) or force rebuilds.

I've an Appveyor account.  I can enable it for linkchecker, if @anarcat doesn't want to do it?

anarcat:
On 2018-01-15 13:46:41, Marius Gedminas wrote:
> I've an Appveyor account.  I can enable it for linkchecker, if @anarcat doesn't want to do it?

Please be my guest. An alternative would be to have a shared secret
somehow.

a.
-- 
I prefer the tumult of liberty to the quiet of servitude.
                        - Thomas Jefferson

appveyor.yml
PetrDlouhy: Probably nothing, I forgot to remove it after copying it from some similar configuration.

mgedmin:
Appveyor is enabled and configured to skip branches without an `appveyor.yml`.

I don't know how to ask it to build this PR -- if I click 'new build', it tries to build the latest commit from master, ignoring my "skip without appveyor.yml" setting.

We'll see how it works when you (or somebody else? ;) pushes some extra commits, e.g. Windows test fixes.

mgedmin:
@PetrDlouhy are you interested in reviving this branch?

git master dropped support for Python 2.7, added support for Python 3.5 through 3.8, and dropped all the C extension modules.

There are some unfortunate merge conflicts that would have to be resolved before we can see how the tests fare on Windows.

cjmayo:
File tests adapted (thanks!) and updated into #495 

Testing MS Windows with GitHub Actions added in #494 (avoids the external dependency).

For the exceptions with Safari bookmarks, I think we would be better served with logging warnings. But needs someone with actual experience of such problems.

PR #131, opened 2018-03-14
Adding Dockerfile

ruzickap:
Adding Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:2
+
+RUN set -x \
+    && apt-get update -qq \
+    && apt-get install -y -qq --no-install-recommends git \
+    && pip install --no-cache-dir git+https://github.com/linkcheck/linkchecker.git@master \
+    && apt-get -y -qq purge git \
+    && apt-get autoremove -y -qq \
+    && apt-get clean -y -qq all \
+    && rm -rf /var/lib/apt/lists/* /usr/share/doc /usr/share/locale /usr/share/man
+
+# /mnt enables linkchecker to access to access files on local machine if needed
+VOLUME /mnt
+
+WORKDIR /mnt
+
+ENTRYPOINT ["linkchecker"]


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:2
+
+RUN set -x \
+    && apt-get update -qq \
+    && apt-get install -y -qq --no-install-recommends gcc git musl-dev python-dnspython python-xdg \
```
anarcat: here in the debian package, the build-deps are:

```
Build-Depends: debhelper (>= 9),
               gettext,
               python-all-dev (>= 2.7),
               python-requests (>= 2.2.1),
               qt4-dev-tools
```

not sure why you need musl/gcc here?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    && pip install --no-cache-dir git+https://github.com/linkcheck/linkchecker.git@master \
+    && apt-get -y -qq purge gcc git musl-dev \
+    && apt-get autoremove -y -qq \
+    && apt-get clean -y -qq all \
+    && rm -rf /var/lib/apt/lists/* /usr/share/doc /usr/share/locale /usr/share/man
+
+VOLUME /mnt
+
+WORKDIR /mnt
```
anarcat: not sure why those are necessary?

Dockerfile
ruzickap: Fixed... Sorry - my bad.

Dockerfile
ruzickap: This will be used in case I would like to test "local files" using `linkchecker`.
I'll update documentation once we the `Dockerfile` will be merged to have it documented properly...

anarcat: ah okay, that makes sense... maybe add a comment explaining that? :)

ruzickap: Done...

anarcat:
maybe squash changes before merging, with a nicer commitlog explaining why we add a dockerfile :)

ruzickap:
Squash can be done by clicking the right button directly in GitHub - but I'll squash it for you...
But I'm not sure what to write to the "commit message"... "docker" people usually know what it means...

anarcat:
On 2018-03-14 09:58:15, Petr Ruzicka wrote:
> Squash can be done by clicking the right button directly in GitHub - but I'll squash it for you...
> But I'm not sure what to write to the "commit message"... "docker" people usually know what it means... 

oh this was mostly adressed at my co-maintainers.

And some people don't know what docker is. And even if they do, just
saying "Docker" doesn't say *why* you think this Dockerfile is a good
idea. A few suggestions:

 * we're not on pip yet
 * extra isolation to try out the package
 * testing new versions

etc.

ruzickap:
Ok...
Commit message changed - but I'll leave it to the co-maintainers to improve it more if necessary.

ruzickap:
Is there anything else I need to do to get it merged ?

anarcat:
you just did :p i guess i still wasn't happy with the commitlog, but that shouldn't be a blocker.

thanks

PR #137, opened 2018-03-25
Dockerfile fix, Documentation updated

ruzickap:
Dockerfile fix, Documentation updated

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,3 +36,18 @@ Usage
 ------
 Execute ``linkchecker http://www.example.com``.
 For other options see ``linkchecker --help``.
+
+Docker usage
+-------------
+
+If you do not want to install any additional libraries/dependencies you can use the Docker image.
+
+Example for external web site check:
+```
+docker run --rm -it -u $(id -u):$(id -g) linkchecker/linkchecker --verbose https://google.com
+```
+
+Local HTML file check:
+```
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -36,3 +36,18 @@ Usage
 ------
 Execute ``linkchecker http://www.example.com``.
 For other options see ``linkchecker --help``.
+
+Docker usage
+-------------
+
+If you do not want to install any additional libraries/dependencies you can use Docker image.
```
mgedmin: I'm not a native English speaker so I could be wrong, but I think there ought to be a "the" in front of "Docker image".

mgedmin:
BTW while testing this I noticed a warning

```
WARNING linkcheck.check 2018-03-25 11:09:02,555 MainThread Running as root user; dropping privileges by changing user to nobody.
```

Can we do something about that in the Dockerfile?

README.rst
ruzickap: I'm not a native speaker too... sorry. 
Fixed...

ruzickap:
I also fixed the "Warning" in my last commit...
Thanks for letting me know...

anarcat:
it would be nice if docker wouldn't run linkchecker as root, to be honest... is there anything we can do about that?

ruzickap:
Yes...
The `-u $(id -u):$(id -g)` parameter will ensure, that linkchecker will run as "non-root" user.
So if you use the "updated examples" from last commit - you should not see the "Running as root user" warning message...

anarcat:
good enough, thanks

PR #138, opened 2018-03-26
improve contrib guidelines

anarcat:
This creates an issue template so that we help people provide us with
reproducible bug reports.

It also moves the code and contrib guidelines to standard locations so
they get better visibility in the GitHub UI. I do not believe the code has visibility yet, but this is where it is expected in this page:

https://github.com/linkcheck/linkchecker/community

Note that there's no change to the code of conduct, it's just moved out to its own file. i previously prefered to include the code in the contribution guidelines, but now I prefer to refer to it, since many people might have read it already, (unfortunately) a bit like licenses files...


anarcat:
note that this document is special, as it denotes itself:

> As an exception to this rule, this specific document cannot be changed
> without the consensus of all administrators of the project.

so I ask for a review (positive or negative) for all maintainers of this organization. it will also be a good time to see if people are still involved/active in the project.

anarcat:
while i'm here, I'm merging the contrib guidelines document with an offsite copy I made in the [ecdsysis](https://ecdysis.readthedocs.io/en/latest/) project, where I maintain a canonical copy of basic contrib guidelines, documentation and code snippets for all projects I maintain. i think the changes can be useful here, but I'm of course totally open to amending them.

note that the work done on the contrib guidelines here is what spurred the creation of the ecdysis project: I figured I was tired of rewriting that document all the time... Since then, I stumbled upon a few other issues and improvements that I fixed in ecdysis and I figured it was time to contribute those back here.

anarcat:
so there are still things missing from the ecdysis project:

 * the [documentation section](https://ecdysis.readthedocs.io/en/latest/contribute.html#documentation). we don't have good, uniform documentation (yet). we should create a Sphinx-based book, but for now it's a bit of a mess, so don't try to encourage people to contribute to iit

 * the [release process](https://ecdysis.readthedocs.io/en/latest/contribute.html#release-process) we don't make many of those, so ... well, there isn't any for now.

anarcat:
okay, this is good to go now, sorry for the noise.

anarcat:
awesome @davejagoda :) is @seamang around?

anarcat:
@seamang ping

Issue #112, opened 2017-11-27
New website links to old website


SilkAndSlug:
[The new website](https://linkcheck.github.io/linkchecker/) has multiple links to [Wummel's website](https://github.com/wummel/linkchecker) in the sidebar, e.g. for Maintainer and Issue Tracker.

PR #143, opened 2018-04-12
replace old links on website

anarcat:
point to the new orga at least while we figure out namespace issues

Closes: #112


anarcat:
this changes only the source code, the website itself is changed in #144.

this was done with `git sed wummel linkcheck`.

PR #144, opened 2018-04-12
replace old links on website

anarcat:
point to the new orga at least while we figure out namespace issues

not sure this is the right way of doing this - i just did the search/replace here and did not regenerate the site from source.

Closes: #112

### DIFF
--- a/donations.html
+++ b/donations.html
@@ -24,7 +24,7 @@
   <div class="inner">
     <h1><a href="./index.html">LinkChecker</a></h1>
     <h2>Check websites for broken links</h2>
-    <a href="https://github.com/wummel/linkchecker"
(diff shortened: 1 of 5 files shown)


anarcat:
this only changes the website, the source change is in #143.

anarcat:
i think this branch is published on http://linkcheck.github.io/linkchecker/ but i must admit i have no clue how the hell this works.

http://linkcheck.github.io, i think, would be a separate repo under the linkcheck organization, named `linkcheck.github.io`. we don't have that, which is why it's a 404. if anything, we should have a blank site there there redirects to the website...

mgedmin:
Ah, yes, that makes sense.

mgedmin:
Sorry hit the wrong button

PR #147, opened 2018-04-13
Require requests >= 2.4

fd0:
This is already in the code in `linkcheck/__init__.py`, update the requirements and setup file so it can be installed via `pip`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # required:
-requests<2.15,>=2.2
+requests>=2.4
 pyxdg
 dnspython
 # optional:
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # required:
-requests>=2.4
+requests<2.15,>=2.4
 pyxdg
 dnspython
 # optional:
--- a/setup.py
+++ b/setup.py
@@ -501,7 +501,7 @@ args = dict(
     },
     # Requirements, usable with setuptools or the new Python packaging module.
     install_requires = [
-        'requests>=2.4',
+        'requests<2.15,>=2.4',
         'dnspython',
         'pyxdg',
     ],
--- a/setup.py
+++ b/setup.py
@@ -501,7 +501,7 @@ args = dict(
     },
     # Requirements, usable with setuptools or the new Python packaging module.
     install_requires = [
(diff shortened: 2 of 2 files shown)


anarcat:
and yes, this is a mess, sorry!

fd0:
It fails when installed via pip on Ubuntu Trusty (that's what Travis uses):

 * pip does not install requests, it's already installed: https://travis-ci.org/restic/restic.net/builds/366213777#L541
 * but running `linkchecker` fails: https://travis-ci.org/restic/restic.net/builds/366213777#L581

anarcat:
right okay. i took the liberty of editing your branch to restore the other limit, could you test the result and see if it works for you?

fd0:
Sure

fd0:
Seems to work, pip installs a newer version of requests :)

anarcat:
awesome, LGTM then.. :) glad to see you use linkchecker and thanks for the contribution!

anarcat:
( @mgedmin it would be nice to cleanup those requirements and specify them just in one place... do you have a quick trick up your sleeves on how to do that? i always wonder why we need the .txt file in the first place...)

fd0:
Cool, I like linkchecker, it does what it should. But I spent two hours fighting the old Python 2.7 (without SNI support *sad*) on Travis on Ubuntu Trusty. That was not fun at all :( (not your fault though, thanks for your work!)

anarcat:
yeah, this thing is really showing its age. this is only what you see as a user, as a developer there are much worse horrors lurking below, e.g. #119

fd0:
WTF? There's an HTML parser in this project? Written in **C**? Oh well. That was... unexpected...

mgedmin:
TBH I think `requirements.txt` is not necessary for us and can be safely deleted.

It is currently used in tox.ini (redundant, except for the 'argcomplete' dependency that tox wouldn't otherwise pick up as it's missing from install_requires), and mentioned in a couple of files in `doc/`.

anarcat:
@mgedmin let's remove it then, do you mind filing a pr?

mgedmin:
Then I would have to come up with documentation updates...  I don't have the energy for that at the moment. :(

PR #155, opened 2018-05-14
Remove unquote/quote on path

ddelporte:
Remove `parse.unquote` calls for path part.
Add char % to `_nopathquote_chars` to disable quote of %

It's a first step, for my part we should not modify host, path and querystring.


anarcat:
so i am concerned about the security implications of this. this quoting/unquoting stuff was done for a reason, and before this gets merged, we need to figure out why.

cjmayo:
Security question above noted.
LinkChecker always syntax checks links, deconstructing them and putting them back together. It would need a carefully constructed argument to justify changing that implementation.

Issue #139, opened 2018-03-26
[meta] how do members get demoted?


anarcat:
while reviewing our contribution guidelines in #138, I realized we had no formal way of removing disappeared or idle maintainers, at least not explicitly in the Membership section. and now I need approval from @davejagoda and @seamang to do what are after all rather trivial changes to the contribution guidelines. are you guys still around? :)

even if our folks are still active, this raises the question what should we do with inactive maintainers. I'm tempted to put a timeout on review requests after which those can be bypassed. i'm also tempted to remove the concensus clause from the guidelines completely to avoid this problem in the future without having to hardcode stuff about how to remove maintainers explicitely. 

thanks for your feedback!

mgedmin:
A timeout of reasonable length (somewhere on the order of 1 week to 1 month?) sounds like a good idea.

anarcat:
okay, let's propose 1 month timeout as a first guideline. should i bundle this in #138 or separately?

mgedmin:
Currently I assumed/hoped #138 was splitting the guidelines into two documents without making any semantic changes.  The way the diff gets computed would make those too hard to notice.

anarcat:
On 2018-03-26 23:20:41, Marius Gedminas wrote:
> Currently I assumed/hoped #138 was splitting the guidelines into two documents without making any semantic changes.  The way the diff gets computed would make those too hard to notice.

Makes sense.

anarcat:
the thing is we can't seem to merge #138 because @seamang is unresponsive. :) furthermore, because of the new org change related to the namespace issue (#136), @seamang is not actually part of the members here anymore. so effectively, we've gone through an informal 4-week timeout now, and I would propose we merge #138 as is now, and make another patch with an explicit timeout.

seamang:
Sorry - I left the job this account was linked to so have not been receiving mails etc. Can you remove me from the maintainers list? I might participate later using my personal account , but not for now. Sorry for holding things up.

anarcat:
@seamang you were already not on the new organisation here, so i just removed your invitation and removed you from the previous project. sorry to see you go, but thanks so much for responding and contributing what you did! :) you will always be welcome here.

anarcat:
so i think the next step here is another PR to amend the contribution guidelines to add a timeout.

PR #164, opened 2018-06-21
introduce demotion rules

anarcat:
While reviewing our contribution guidelines in #139, we found we had no formal way of removing disappeared or idle maintainers, at least not explicitly, in the Membership section. This was a problem to get changes done on the contributing document which requires consensus from the maintainers.

This change introduces a way to make changes to the document by using it as a "vote of confidence", in a way: if a change is proposed and an administrator does not respond within a given 1-month delay, the administrator can be demoted down to a regular maintainer and the document approved.

Closes: #139


anarcat:
> hmm... what's up with github/travis here? did we mess up CI in the project rename?

we didn't mess up the rename, I just had to grant Travis access to the new orga here... now done, and CI is running.

/me misses gitlab-ci :p

PR #175, opened 2018-08-02
Basic typo correction

regexaurus:
Just a basic typo correction.


regexaurus:
I'm also curious about the *send your thanks* URL `https://saythanks.io` in the *Positive feedback* section of CONTRIBUTING.mdwn. That URL directs one to the *Say Thanks* project home page, where you can either create or access your "thankfulness" inbox.  For this to work as expected, I think the URL to provide for users' feedback would be of the form https://saythanks.io/mailbox?id=ei34v9 or similar.

anarcat:
@regexaurus thanks for your contribution, and yes, it seems the URL might be a little too generic. :)

Issue #183, opened 2018-08-20
Version problems with pip on Mac


reagle:
## Summary

If I have to use python2 on Mac, the only version of requests available is 2.19.1 -- which I would think is greater than 2.2.0 .

## Steps to reproduce
```
╭─reagle@hom ~/tmp
╰─➤  pip2 install LinkChecker
Requirement already satisfied: LinkChecker in /usr/local/lib/python2.7/site-packages (9.3)
Requirement already satisfied: requests>=2.2.0 in /usr/local/lib/python2.7/site-packages (from LinkChecker) (2.19.1)
Requirement already satisfied: certifi>=2017.4.17 in /usr/local/lib/python2.7/site-packages (from requests>=2.2.0->LinkChecker) (2018.8.13)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in /usr/local/lib/python2.7/site-packages (from requests>=2.2.0->LinkChecker) (3.0.4)
Requirement already satisfied: urllib3<1.24,>=1.21.1 in /usr/local/lib/python2.7/site-packages (from requests>=2.2.0->LinkChecker) (1.23)
Requirement already satisfied: idna<2.8,>=2.5 in /usr/local/lib/python2.7/site-packages (from requests>=2.2.0->LinkChecker) (2.7)
╰─➤  linkchecker -v
This program requires Python requests 2.2.0 or later.
╰─➤  yolk -V requests
requests 2.19.1
```

## Environment
```
Darwin hom 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64 i386 iMac16,2 Darwin
Python 2.7.10 (default, Oct  6 2017, 22:29:07)
```

anarcat:
the linkchecker version on pip is outdated, see #108 , #136 , #4, etc.

the new readme says to use this to install linkchecker since #181:

    pip install git+https://github.com/linkchecker/linkchecker.git

reagle:
I recommend making the text `pip install git+https://github.com/linkchecker/linkchecker.git` also be in code style, otherwise it's less prominent than what you are telling people what not to do.

PR #185, opened 2018-08-24
Update readme

reagle:
As discussed in #183 .


reagle:
I apparently didn't get the formatting right for the code style for the
command. (I am not as familiar with restructure text as markdown.)

anarcat:
ah, doh. it's because this is an RST file, indeed... you need:

```
``foo``
```

(ie. double backquotes)

anarcat:
i did the same :)

PR #196, opened 2018-11-03
Require a recent Requests version

mgedmin:
Requests versions <= 2.19.1 are vulnerable to CVE-2018-18074:

> The Requests package through 2.19.1 before 2018-09-14 for Python sends
> an HTTP Authorization header to an http URI upon receiving a
> same-hostname https-to-http redirect, which makes it easier for remote
> attackers to discover credentials by sniffing the network.

Thank you, GitHub security alerts, for bringing this to my attention.

I do not recall why we were pinning requests to `< 2.15` and I hope Travis CI will remind me.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # required:
-requests<2.15,>=2.4
+requests >= 2.20.0
 pyxdg
 dnspython
 # optional:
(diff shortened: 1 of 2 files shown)


mgedmin:
Good: Travis says no problems exist with using a new requests version.

mgedmin:
Good point.

(I'm too lazy to remove the minimum before merging, but I won't object if anyone wants to do that.)

anarcat:
let's if see #197 still fixes the warning.

PR #197, opened 2018-11-03
lower requirements to allow install on older systems

anarcat:
PR #196 bumped this to 2.20 to fix a security warning about requests
in our repository, but I believe that warning can be fixed by removing
only the upper bound. This should make backporting linkchecker to
older systems easier, where requests presumably has the 2.20 fix
backported.


anarcat:
okay, let's see what this does - if the warning comes back, you have my approval to revert this of course.

PR #194, opened 2018-10-31
Do include anchors within cache_url,  mutex around the parsing

yarikoptic:
Closes #179

Please see details of the "investigation" in https://github.com/wummel/linkchecker/issues/557#issuecomment-434552215

- [ ] consider another approach (may be adding `full_url` record in addition) instead of populating anchor into `cache_url` which makes it to download them same page multiple times
- [ ] introduced test demonstrates an outstanding problem in threaded mode that in some runs anchor pointing to an element defined later in the document would be considered broken (I guess because parsing of the page is not finished yet whenever anchor is verified). It actually can manifest it in two ways:

<details>
<summary>printouts of failed runs</summary>

```
tests/checker/__init__.py:232: in fail_unicode
    self.fail(msg)
E   AssertionError: Differences found testing file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html (records might switch order which is ok)
E   @@ -2,7 +2,7 @@
E    cache key file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html#bad1
E    real url file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html
E    name first local bad one
E   -warning Anchor `bad1' not found. Available anchors: `anchor1'.
E   +warning Anchor `bad1' not found. Available anchors: -.
E    valid
E    url #bad2
E    cache key file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html#bad2
```
and
```
E   AssertionError: Differences found testing file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html (records might switch order which is ok)
E   @@ -1,8 +1,8 @@
E   -url #bad1
E   -cache key file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html#bad1
E   +url #anchor1
E   +cache key file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html#anchor1
E    real url file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html
E   -name first local bad one
E   -warning Anchor `bad1' not found. Available anchors: `anchor1'.
E   +name good anchor used before actual element is defined
E   +warning Anchor `anchor1' not found. Available anchors: -.
E    valid
E    url #bad2
E    cache key file:///home/yoh/proj/misc/linkchecker/tests/checker/data/anchors.html#bad2
E   @@ -10,6 +10,12 @@
E    name second local bad one
[...]

PR #198, opened 2018-11-06
Mutex around parse

yarikoptic:
Portion from #194 which is trivial and independent from that one which might take longer to fixup properly.

### DIFF
--- a/linkcheck/parser/__init__.py
+++ b/linkcheck/parser/__init__.py
@@ -133,7 +133,7 @@ def find_links (url_data, callback, tags):
         content = url_data.get_content()
         with parse_mutex:
             parser.feed(content)
-        parser.flush()
+            parser.flush()
     except linkparse.StopParse as msg:
         log.debug(LOG_CHECK, "Stopped parsing: %s", msg)
         pass
--- a/linkcheck/parser/__init__.py
+++ b/linkcheck/parser/__init__.py
@@ -17,10 +17,15 @@
 """
 Main functions for link parsing
 """
+import threading
+
 from .. import log, LOG_CHECK, strformat, url as urlutil
 from ..htmlutil import linkparse
 from ..HtmlParser import htmlsax
 from ..bookmarks import firefox
+
+# Is needed within find_links around non-threadsafe call
+parse_mutex = threading.Lock()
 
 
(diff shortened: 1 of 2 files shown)


linkcheck/parser/__init__.py
```diff
@@ -125,7 +130,9 @@ def find_links (url_data, callback, tags):
     handler.parser = parser
     # parse
     try:
-        parser.feed(url_data.get_content())
+        content = url_data.get_content()
+        with parse_mutex:
+            parser.feed(content)
         parser.flush()
```
mgedmin: I may be missing previous context, but it's unclear to me why `parser.feed()` is not thread-safe, but `parser.flush()` is fine?

Are we using the same parser instance in multiple threads?  (It doesn't look like we do.)

Is the parser using global state shared between multiple parser instances?

Is it perhaps GNU bison that's giving us a thread-unsafe parser?

Perhaps we should prioritize #119 that replaces our homegrown C parser with BeatifulSoup?

yarikoptic:
pushed that little patch offsetting flush to be under mutex as well.

Would be nice if there was a test demonstrating its need... unfortunately the one targeting anchors being added in #194 is not triggering it AFAIK, and originally I detected it while running on a ad-hoc test case I simulated in local collection of .htmls I generated by building https://github.com/bids-standard/bids-specification .

yarikoptic:
> I have no time/inclination to study the C sources generated by bison

who would? ;-)
Unfortunately I've not kept an original ad-hoc modified sample to test anchors situation and ATM without it cannot reproduce fluctuating reported links count to provide you undeniable evidence of positive effects.

yarikoptic:
what should be done to get at least this one merged? ;-)

mgedmin:
I would like to hear @anarcat's opinion.

anarcat:
@mgedmin i'll let you do the honors considering you actually herded this considerably more than i did. :)

mgedmin:
Ha!  You just want me to feel responsible if this ever backfires ;)

anarcat:
yes, of course, that too. :p

yarikoptic:
Thank you guys!  I still hope some time to get back to #194 -- feel welcome to beat me to it ;-)

PR #211, opened 2019-04-10
WIP: Python3: update makefiles to test in Python 3 in allow-failures mode

cjmayo:


linkcheck/HtmlParser/Makefile
```diff
@@ -2,7 +2,7 @@
 # reentrant bison parser support and uses features of bison >= 3.0.x
 LEX = flex
 YACC = bison
-PYINCLUDE=-I/usr/include/python2.7
+PYINCLUDE=-I/usr/include/python
```
mgedmin: FWIW my machine (Ubuntu 18.10) does not have a /usr/include/python.  It has /usr/include/python2.7 and /usr/include/python3.* for various values of *.

I wonder if we should maybe use something like `$(pkg-config --cflags python)` (which on my machine prints -I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7)?

(Or we could merge that PR that replaces linkcheck/HtmlParser with BeautifulSoup.)

cjmayo: I don't have /usr/include/python either, and nor does `pkg-config --cflags python` work. Supporting both Python 2 and 3 is complicated, and varies between distrubutions...

On the other hand, what does this file need to do? Supply a clean instruction? Run the awk?
Because setup.py does seem to do the actual building using Extension - which is the bit that needs the include.

cjmayo:
Suggest closing this one without merging. Project is building fine from `setup.py` without these changes and there is at least one other Makefile problem. Needs someone who has a use for the Makefiles to investigate.

anarcat:
i'd like to get rid of the Makefile altogether, if it's not useful. this can (and should) all be replaced by setup.py stuff, IMHO. yet this probably doesn't block py3 porting, so keeping this open until someone deals with the Makefile (or opens a more targeted issue to do so).

cjmayo:
This one also for the record:
https://github.com/linkchecker/linkchecker/blob/9d57bee16fd3137ffe7dc28a603285a068b103d1/po/Makefile#L5

mgedmin:
> Suggest closing this one without merging. Project is building fine from setup.py without these changes and there is at least one other Makefile problem. Needs someone who has a use for the Makefiles to investigate.

I like Makefiles, but I don't have the energy to investigate this.  +1 for closing (or even removing the Makefile entirely).

(IMHO in a Python project Makefiles are primarily a tool for maintainers.  They make sense when the maintainer finds them useful.  I find them useful: all of my personal Python projects have a Makefile that lets me do things like `make test` and `make release` without having to remember what particular tools or additional steps this one project has.)

anarcat:
> +1 for closing (or even removing the Makefile entirely).

Yep, done.

> (IMHO in a Python project Makefiles are primarily a tool for maintainers. They make sense when the maintainer finds them useful. I find them useful: all of my personal Python projects have a Makefile that lets me do things like make test and make release without having to remember what particular tools or additional steps this one project has.)

While I understand where this is coming from, adding Make as a dependency to a project, even if only for maintenance, adds yet another hurdle to adoption for new people. They need to learn another tool that's not part of the standard programming environment. Python (and especially distutils and friends) is complicated enough without having people figure out Make on top of that...

I would actually agree if Makefiles were used as tiny wrappers on top of `setup.py` to remind people of the right targets to use. But looking at our own Makefile, it's not really the case: it's 200 lines of business logic, doing everything from a range of linting, unit tests, debian package build (!), homepage generation. It's a mess! And that's often what happens in those cases: the Makefile is there and then it's easy to expand it to do all sorts of random things and it grows and grows until it supplants `setup.py` for even stuff that it's supposed to be doing (like `setup.py test`).

This is arguably because packaging is too hard in Python. But that's being improved and we should follow that instead of hiding it under a layer of Makefile-ing.

So while I also *do* like Makefiles and use them often, I think that, in our case, we should totally get rid of it and do everything in setup.py, like good Python citizens. ;)

But until we go there, i think this can be closed as well. :)

PR #212, opened 2019-04-10
{python3_02} Python3: fix regexp format

cjmayo:


linkcheck/htmlutil/linkparse.py
```diff
@@ -82,11 +82,11 @@
...
 # matcher for <meta http-equiv=refresh> tags
-refresh_re = re.compile(ur"(?i)^\d+;\s*url=(?P<url>.+)$")
-_quoted_pat = ur"('[^']+'|\"[^\"]+\"|[^\)\s]+)"
-css_url_re = re.compile(ur"url\(\s*(?P<url>%s)\s*\)" % _quoted_pat)
+refresh_re = re.compile(r"(?i)^\d+;\s*url=(?P<url>.+)$")
+_quoted_pat = r"('[^']+'|\"[^\"]+\"|[^\)\s]+)"
+css_url_re = re.compile(r"url\(\s*(?P<url>%s)\s*\)" % _quoted_pat)
 swf_url_re = re.compile("(?i)%s" % urlutil.safe_url_pattern)
-c_comment_re = re.compile(ur"/\*.*?\*/", re.DOTALL)
+c_comment_re = re.compile(r"/\*.*?\*/", re.DOTALL)
```
mgedmin: On Python 2 the type of the string passed to re.compile (unicode vs str) is significant.  I'm not quite sure what the implications are, but I worry about breaking things that might not be caught by the test suite.

Of course, on Python 3 the `ur"..."` literals are invalid syntax, but we could double the backslashes and use `u"..."` literals.

cjmayo: Doubled all `\` except those escaping `"`.
Prefixed all the strings with `u`, including swf_url_re. This does seem the right thing to do in general and not just for re:
https://python-future.org/futurize.html#separating-text-from-bytes

mgedmin: TBH I no longer trust my judgement - I tried to reproduce the kinds of problems I seemed to remember you can get from mixing up bytes vs unicode strings and regexps on Python 2.  I couldn't get any breakage.

cjmayo: Is the conclusion to go back to the original patch with `r"` (in all instances)?

mgedmin: Yeah, sorry!

cjmayo: No problem. Just cherry-picked the original commit in.

PR #213, opened 2019-04-10
{python3_03} Python3: fix HtmlParser

cjmayo:


linkcheck/HtmlParser/htmlparse.c
```diff
@@ -2460,35 +2460,36 @@ MOD_INIT(htmlsax) {
...
         Py_DECREF(resolve_entities);
         Py_DECREF(set_encoding);
-        Py_DECREF(m);
+        Py_DECREF(h);
         return MOD_ERROR_VAL;
     }
-    Py_DECREF(m);
-    if ((u_meta = PyString_Decode("meta", 4, "ascii", "ignore")) == NULL) {
+    Py_DECREF(h);
+    if ((u_meta = PyUnicode_Decode("meta", 4, "ascii", "ignore")) == NULL) {
```
mgedmin: I would suggest `PyUnicode_FromStringAndSize("meta", 4)` as a slight simplification.

linkcheck/HtmlParser/htmlparse.c
```diff
@@ -2460,35 +2460,36 @@ MOD_INIT(htmlsax) {
         /* init error */
         PyErr_Print();
     }
-    if ((m = PyImport_ImportModule("linkcheck.HtmlParser")) == NULL) {
+    PyObject* h = NULL;
+    if ((h = PyImport_ImportModule("linkcheck.HtmlParser")) == NULL) {
```
mgedmin: What is the reason for renaming `m` to `h` in the rest of this function?

linkcheck/checker/mailtourl.py
```diff
@@ -57,8 +57,8 @@ def is_literal (domain):
     return domain.startswith(u'[') and domain.endswith(u']')
 
 
-_remove_quoted = re.compile(ur'\\.').sub
-_quotes = re.compile(ur'["\\]')
+_remove_quoted = re.compile(r'\\.').sub
+_quotes = re.compile(r'["\\]')
```
mgedmin: Like mentioned on #212, I'm not entirely sure this is safe on Python 2.

linkcheck/HtmlParser/htmlparse.c
cjmayo: Done.

linkcheck/HtmlParser/htmlparse.c
cjmayo: To do with the return value?:

https://github.com/linkchecker/linkchecker/blob/7c0d8603368b33c4c409a30434ba9f9e93b4d898/linkcheck/HtmlParser/htmlparse.c#L2494

linkcheck/checker/mailtourl.py
cjmayo: Made `u` as #212 with doubled `\`.

linkcheck/HtmlParser/htmlparse.c
mgedmin: Thank you!

I see, on Python 2 MOD_SUCCESS_VAL(m) is defined to be empty, so we need to preserve the actual module we're defining and return it on Python 3 only.  Which explains why the old code was not buggy on Python 2, and why it needed to be changed to support Python 3.

linkcheck/checker/mailtourl.py
cjmayo: Reverted back to:
```py
_remove_quoted = re.compile(r'\\.').sub
_quotes = re.compile(r'["\\]')
```

linkcheck/HtmlParser/htmlparse.y
```diff
@@ -988,35 +988,36 @@ MOD_INIT(htmlsax) {
...
         Py_DECREF(resolve_entities);
         Py_DECREF(set_encoding);
-        Py_DECREF(m);
+        Py_DECREF(h);
         return MOD_ERROR_VAL;
     }
-    Py_DECREF(m);
-    if ((u_meta = PyString_Decode("meta", 4, "ascii", "ignore")) == NULL) {
+    Py_DECREF(h);
+    if ((u_meta = PyUnicode_Decode("meta", 4, "ascii", "ignore")) == NULL) {
```
mgedmin: This could also be PyUnicode_FromStringAndSize().

cjmayo: Changed as well.

PR #215, opened 2019-04-10
{python3_05} Python3: fix masks, logging and range

cjmayo:


cjmayo:
Rebased onto python3_04 to attempt to pass tests (missing builtins module). I will wait to fix the other PR test failures until this one is working.

linkcheck/configuration/__init__.py
```diff
@@ -403,7 +403,7 @@ def make_userdir(child):
             # Windows forbids filenames with leading dot unless
             # a trailing dot is added.
             userdir += "."
-        os.makedirs(userdir, 0700)
+        os.makedirs(userdir, 0o700)
```
mgedmin: This part is great!

linkcheck/decorators.py
```diff
@@ -36,12 +36,16 @@ def h ():
...
 from __future__ import print_function
+
+import codecs
 import warnings
 import signal
 import os
 import sys
 import time
 
+from builtins import bytes
```
mgedmin: This part is a bit confusing.  (Doesn't Python 2 already have `bytes`?)

linkcheck/decorators.py
```diff
@@ -125,9 +129,9 @@ def newfunc (*args, **kwargs):
         res = func(*args, **kwargs)
         duration = time.time() - t
         if duration > limit:
-            print(func.__name__, "took %0.2f seconds" % duration, file=log)
-            print(args, file=log)
-            print(kwargs, file=log)
+            log.write(bytes("%s took %0.2f seconds\n" %(func.__name__, duration), "utf8"))
+            log.write(bytes("%s\n" % str(args), "utf8"))
+            log.write(bytes("%s\n" % str(kwargs), "utf8"))
```
mgedmin: This part is very confusing?  What's going on here?

linkcheck/decorators.py
```diff
@@ -36,12 +36,16 @@ def h ():
 
 """
 from __future__ import print_function
+
+import codecs
```
mgedmin: I don't see where this new import is getting used?

linkcheck/logger/blacklist.py
```diff
@@ -90,7 +90,7 @@ def write_blacklist (self):
         """
         Write the blacklist.
         """
-        oldmask = os.umask(0077)
+        oldmask = os.umask(0o077)
```
mgedmin: This part is great!

tests/test_decorators.py
```diff
@@ -42,4 +42,5 @@ def f ():
             time.sleep(1)
             return 42
         self.assertEqual(f(), 42)
-        self.assertTrue(log.getvalue())
+        log_value = log.getvalue()
+        self.assertEquals(log_value, b'f took 1.00 seconds\n()\n{}\n')
```
mgedmin: assertEqual**s** emits a deprecation warning; please use `assertEqual`.

(And, again, I don't understand why this needs to be bytes?  Especially given that it's writing to a StringIO and not a BytesIO?)

cjmayo: It was using:
`log = BytesIO()`
Changed to StringIO in python3_05c.

linkcheck/decorators.py
cjmayo: Removed in python3_05c.

cjmayo:
Split into #233 , #234 and #235.

PR #216, opened 2019-04-10
{python3_06} Python3: fix tests init - exceptions and string

cjmayo:


tests/__init__.py
```diff
@@ -24,6 +24,7 @@
 import pytest
 from contextlib import contextmanager
 from linkcheck import LinkCheckerInterrupt
+from builtins import str
```
mgedmin: I'm not familiar with `future`.  Is this equivalent to `str = unicode` on Python 2?

I'm not sure I like that, it seems very subtle.  I much prefer the approach `six` took where you have a new `six.text_type` to mean `unicode` on Python 2 and `str` on Python 3.

(Still, this is test code, I'm not going to veto anything.)

cjmayo:
Rebased on master for build checks.

cjmayo:
Rebased again and updated for `as str_text`

tests/__init__.py
```diff
@@ -90,7 +91,7 @@ def newfunc (*args, **kwargs):
             if not testfunc():
                 pytest.skip("%s is not available" % name)
             return func(*args, **kwargs)
-        newfunc.func_name = func.func_name
+        newfunc.__name = func.__name__
```
mgedmin: Typo: `__name` should be `__name__`.

In fact I suggest using `@functools.wraps(func)` to decorate the `def newfunc` instead of copying metadata by hand. It'll also take care to copy `__doc__` and `__module__` and all sorts of other important things.

cjmayo: Good to know. Added `wraps`.

PR #217, opened 2019-04-10
{python3_07} Python3: use BytesIO instead of StringIO

cjmayo:


linkcheck/logger/csvlog.py
```diff
@@ -69,7 +69,7 @@ def start_output (self):
         else:
             # write empty string to initialize file output
             self.write(u"")
-        self.queue = StringIO()
+        self.queue = BytesIO()
         self.writer = csv.writer(self.queue, dialect=self.dialect,
                delimiter=self.separator, lineterminator=self.linesep,
                quotechar=self.quotechar)
```
mgedmin: TBH I'm surprised this works on Python 3.

(... does it work on Python 3?  I thought csv.writer wants to do I/O using native `str` objects, i.e. bytes on Python 2 but unicode strings on Python 3.)

PR #218, opened 2019-04-10
{python3_08} Python3: use str and basestring from builtins

cjmayo:


tests/test_dummy.py
```diff
@@ -60,4 +62,4 @@ def test_indexes (self):
         del dummy[2:3]
         str(dummy)
         repr(dummy)
-        unicode(dummy)
+        str(dummy)
```
mgedmin: The test is now doing `str(dummy)` twice.  I suspect one of those should be changed to `bytes(dummy)`?

cjmayo: Done.

cjmayo: Wasn't quite that simple. With Python 3:
```
>       bytes(dummy)
E       TypeError: 'Dummy' object cannot be interpreted as an integer
```
Updated to:
```py
bytes(str(dummy), "utf-8")
```
and importing bytes from builtins.

mgedmin: No, that just tests that str(dummy) works again.

I'm, as usual, hampered by my complete unfamiliarity with the source code.  What is the Dummy class used for?  It had tests for converting both to bytes and to unicode on Python 2.  Should it have tests for converting both to bytes and to unicode on Python 3?  If yes, then it would need an explicit `__bytes__` method.  If no, then the extra test can be simply removed (on Python 3 only).

cjmayo: Dose anyone know what Dummy is for?

In the meantime I've updated to:
```py
-        unicode(dummy)
+        if sys.version_info.major == 2:
+            unicode(dummy)
```

mgedmin: "A dummy object ignores all access to it. Useful for testing." says the docsting.

It's actually used in _Logger.start_fileoutput() and _Logger.write(), for which only Dummy().flush(), Dummy().write('...') and Dummy().close() should work and do nothing.  (This is handled by Dummy.__getattr__() returning self and Dummy.__call__() doing nothing.)

Dummy does define both `__str__` and `__unicode__` on Python 2, so for conceptual coherence it should be defining `__bytes__` and `__str__` on Python 3.  In practice it doesn't matter as nothing ever calls any of those.

Long-term I think we should drop the entire class and use a simpler null object for the do-nothing logging case (and use mock.Mock() if we ever need a dummy object in tests).

PR #223, opened 2019-04-10
{python3_13} Python3: fix imports in test_noproxy

cjmayo:


tests/checker/test_noproxy.py
```diff
@@ -17,8 +17,11 @@
 """
 Test proxy handling.
 """
-import httpserver
-from test.test_support import EnvironmentVarGuard
+from .import httpserver
```
mgedmin: I think PEP-8 wants this to be `from . import httpserver`.

cjmayo: Done.

cjmayo:
@mgedmin please confirm OK.

PR #227, opened 2019-04-10
{python3_17} Python3: fix unicode in url.py

cjmayo:


linkcheck/url.py
```diff
@@ -28,6 +28,7 @@
     import urlparse
 
 import requests
+from builtins import str as text
```
mgedmin: (I would even suggest `text_type`, but that might be just that I'm so used to [`six`](https://pypi.org/project/six/))

cjmayo:
Rebased on master for build checks.

cjmayo:
Updated for `str_text` and rebased.

PR #228, opened 2019-04-10
{python3_18} Python3: fix unicode in urlbase

cjmayo:


linkcheck/checker/urlbase.py
```diff
@@ -39,6 +39,8 @@
 except ImportError:
     # Python 3
     from io import StringIO
+from builtins import str as str_text
```
mgedmin: I think we should be consistent with the aliases we give to the unicode text type in various source files.  So far I've seen `str`, `text` and now `str_text`.

My personal preference would be `text_type`.

anarcat: what's wrong with `str`?

mgedmin: When I see `str`, I'm going to assume it's the native string type (bytes on Python 2, unicode on Python 3).

Importing `str` from `builtins` (which _looks_ deceptively like an [stdlib module](https://docs.python.org/3/library/builtins.html), but is actually a 3rd-party module provided by [future](https://pypi.org/project/future/)) will change that to be unicode on both Python 2 and 3, on a per-file basis.

anarcat: oh wow, that *is* truly confusing... so `builtins` from future shadows the stdlib module? that seems kind of wrong... 

or maybe it's deliberate? hmm...

mgedmin: TBH I've no idea what `future` does on Python 3.  I'd hope it has a mechanism to avoid shadowing the real builtins, but I've never checked?
.

cjmayo: Yes in builtins `__init__.py`:
```py
if PY3:
    import builtins
    bytes = builtins.bytes
    dict = builtins.dict
    int = builtins.int
    list = builtins.list
    object = builtins.object
    range = builtins.range
    str = builtins.str
    __all__ = []
else:
    from future.types import (newbytes as bytes,
                              newdict as dict,
                              newint as int,
                              newlist as list,
                              newobject as object,
                              newrange as range,
                              newstr as str)
```

I think the future philosophy is enabling Python 3 code that runs on Python 2. So, `from builtins import str` would be the preferred option.
https://python-future.org/futurize.html#separating-text-from-bytes

Maybe that's not going to be possible (straightforward) in every file.

anarcat: okay then maybe `str` is perfectly reasonable?

mgedmin: > okay then maybe str is perfectly reasonable?

￼I suppose, if we do it consistently?

What I don't like is that all it takes is forgetting to do the import in a single source module and suddenly `str` in one file means a different thing from `str` in another file.

I don't like it.

anarcat: think of it like this: in a year, Python 2 will be unsupported and we will start removing support from it altogether, and this will be all moot as `str` will mean the same everywhere. in python3, where it matters, it *will* mean the same everywhere regardless of whether the import is there or not.

so, long term, this is not a problem. short term, unit tests should catch it. if it's not, we have a bug, in python2, which will be unsupported in a year. :)

so i'd go with `str`.

(really, i'm not even sure we should bend over backwards so much to support both py2 and py3, to be honest. but since we've taken that route, I think it's great we're doing it. but i wouldn't block on stuff like that.)

mgedmin: > so i'd go with str.

Okay.

> (really, i'm not even sure we should bend over backwards so much to support both py2 and py3, to be honest. but since we've taken that route, I think it's great we're doing it. but i wouldn't block on stuff like that.)

Speaking from experience, supporting both 2 and 3 is the _easier_ route.

If you break Python 2 while implementing support for Python 3, you'll have to do the conversion all in one go, which is _hard_ for a larger codebase.  And then you cannot merge your changes until you fix everything, which takes _time_, and then then new bugfixes get merged to master and the code diverges and you get merge conflicts...

Whereas if you migrate from 2-only to 2 and 3, you can make the conversion piecemeal, while having a working program all of the time.  (And then you can drop Python 2 compatibility bits once the software's been sufficiently tested on Python 3.)

mgedmin:
#214 was merged so this one probably needs a rebase to make the tests pass

linkcheck/checker/urlbase.py
cjmayo: I think go with `str` where the code is obviously going to be good for Python 3, if unsure use `text` so they can be easily found when moving to Python 3 only.

This affects #216, #227, #229 and #230 and also means reviewing a few changes we have already made:
```
$ grep -rn builtins | grep str
linkcheck/configuration/confparse.py:25:from builtins import str
tests/configuration/test_config.py:24:from builtins import str
tests/test_dummy.py:24:from builtins import bytes, str
tests/test_updater.py:24:from builtins import str
```

This is based on an overall strategy of continuing what we have been doing, piece by piece converting to Python 2 + 3, because:
1) It seems to be going OK
2) It won't be possible to switch to Python 3 without replacing the parser. This way in theory a new Python 2 release could be made from master if needed.

Does that make sense?

cjmayo: After starting on this `as text` doesn't seem like a good idea - text is used as a variable name in a few places. `str_text` would cause me the least problems when trying to manage the python3 branch.

 #245 is an example of code that can be removed when Python 2 is no longer supported.

cjmayo:
Rebased and updated for str_text.

linkcheck/checker/urlbase.py
mgedmin: > After starting on this `as text` doesn't seem like a good idea - text is used as a variable name in a few places.

Which is why I suggested `text_type` ;)

> `str_text` would cause me the least problems when trying to manage the python3 branch.

That's fine also.

linkcheck/strformat.py
```diff
@@ -54,10 +56,10 @@ def unicode_safe (s, encoding=i18n.default_encoding, errors='replace'):
     @rtype: unicode
     """
     assert s is not None, "argument to unicode_safe was None"
-    if isinstance(s, unicode):
+    if isinstance(s, str_text):
         # s is already unicode, nothing to do
         return s
-    return unicode(str(s), encoding, errors)
+    return str_text(bytes(s), encoding, errors)
```
mgedmin: This feels wrong to me.  Previously, `unicode_safe(42)` would return `u'42'`.  Now it will return a string with 42 NUL bytes.

linkcheck/strformat.py
```diff
@@ -69,7 +71,7 @@ def ascii_safe (s):
     @return: encoded ASCII version of s, or None if s was None
     @rtype: string
     """
-    if isinstance(s, unicode):
+    if isinstance(s, str_text):
         s = s.encode('ascii', 'ignore')
     return s
```
mgedmin: I only just noticed all those `@rtype: string` annotations that should maybe be mass-replaced with `@rtype: bytes` where it makes sense?  (In a separate PR maybe, this one's already large enough.)

(I'm not 100% convinced it makes sense for ascii_safe() to return bytes on Python 3, but then I haven't actually looked at where it's used.)

cjmayo:
Branched off ascii_safe() and unicode_safe() into PR #251.

linkcheck/strformat.py
cjmayo: I'm tending to think we park this in the new PR and revisit after going Python 3 only.

PR #229, opened 2019-04-10
{python3_19} Python3: fix unicode in fileurl

cjmayo:


linkcheck/checker/fileurl.py
```diff
@@ -135,7 +136,10 @@ def build_base_url(self):
             base_url = re.sub("^file://(/?)([a-zA-Z]):", r"file:///\2|", base_url)
             # transform file://path into file:///path
             base_url = re.sub("^file://([^/])", r"file:///\1", base_url)
-        self.base_url = unicode(base_url)
+        try:
+            self.base_url = unicode(base_url)
+        except NameError:
+            self.base_url = base_url
```
mgedmin: Why not `text(base_url)`?

cjmayo: It causes test failures with Python 2.7.

cjmayo:
Rebased and updated for str_text.

linkcheck/checker/fileurl.py
mgedmin: I would like to understand why.  Isn't `str_text` an alias for `unicode` on 2.7?  Or is it future's backport of the Python 3 str that differs from 2.7's `unicode` somehow? (How?)

*checks [documentation](http://python-future.org/what_else.html#str)*

Ah, ok, so `str_text` differs from `unicode` in that it won't allow mixing with `bytes` under some circumstances, AFAIU.

I don't know that I want to know the precise details where things break in our test suite.

PR #232, opened 2019-04-13
Remove leftovers from introduction of requests

cjmayo:
Things that weren't removed in 7b34be590b.


anarcat:
dafuq? thanks for spotting this.

mgedmin:
> (whyyyy did linkchecker even have a copy of the stdlib's gzip.py???)

That was rhetorical but then I got curious and the answer is: to work around an upstream bug (something to do with incomplete gzip streams) in https://github.com/linkchecker/linkchecker/commit/e7ca18403a1d72cfd7f8d50add3b6776dae7b50e.

It was no longer used since the switch to `requests` in https://github.com/linkchecker/linkchecker/commit/7b34be590b52a85b207fae335d0885c85eb7b0aa.

PR #235, opened 2019-04-13
{python3_05c} Python3: fix logging

cjmayo:
Split from #215.

### DIFF
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -20,7 +20,13 @@ Test decorators.
 
 import unittest
 import time
-from io import BytesIO
+
+try:
+    from cStringIO import StringIO
+except ImportError:
+    # Python 3
+    from io import StringIO
+
 import linkcheck.decorators
 
 
@@ -36,10 +42,10 @@ class TestDecorators (unittest.TestCase):
         self.assertEqual(f(), 42)
 
     def test_timeit2 (self):
-        log = BytesIO()
+        log = StringIO()
         @linkcheck.decorators.timed(log=log, limit=0)
         def f ():
             time.sleep(1)
             return 42
         self.assertEqual(f(), 42)
-        self.assertTrue(log.getvalue())
+        self.assertEqual(log.getvalue(), 'f took 1.00 seconds\n()\n{}\n')


cjmayo:
More checking needed on this one before merging. It's the log in the test I've changed to StringIO but maybe everywhere else it is bytes.

linkcheck/decorators.py
```diff
@@ -125,9 +125,9 @@ def newfunc (*args, **kwargs):
         res = func(*args, **kwargs)
         duration = time.time() - t
         if duration > limit:
-            print(func.__name__, "took %0.2f seconds" % duration, file=log)
-            print(args, file=log)
-            print(kwargs, file=log)
+            log.write("%s took %0.2f seconds\n" % (func.__name__, duration))
+            log.write("%s\n" % str(args))
+            log.write("%s\n" % str(kwargs))
```
mgedmin: Is this change necessary?

tests/test_decorators.py
```diff
@@ -20,7 +20,13 @@
 
 import unittest
 import time
-from io import BytesIO
+
+try:
+	from cStringIO import StringIO
+except ImportError:
+	# Python 3
+	from io import StringIO
```
mgedmin: The indentation here is doubled (did you use a tab character by accident)?

tests/test_decorators.py
```diff
@@ -36,10 +42,10 @@ def f ():
...
     def test_timeit2 (self):
-        log = BytesIO()
+        log = StringIO()
         @linkcheck.decorators.timed(log=log, limit=0)
         def f ():
             time.sleep(1)
             return 42
         self.assertEqual(f(), 42)
-        self.assertTrue(log.getvalue())
+        self.assertEqual(log.getvalue(), 'f took 1.00 seconds\n()\n{}\n')
```
mgedmin: This looks right to me: the default value of `log` is `sys.stderr`, which is definitely a text stream and not a byte stream on Python 3.

(The decorator itself looks like debug code to be inserted while developing and currently has no uses outside of the test suite, so getting it wrong won't cause any harm to users.)

linkcheck/decorators.py
cjmayo: Reverted.

tests/test_decorators.py
cjmayo: Yes, more by habit than accident...

cjmayo:
That should be good now.

PR #238, opened 2019-04-14
[WIP] Avoid segfaults while testing on Python 3.6

cjmayo:
Default memory allocator changed in Python 3.6 to pymalloc, where available, meaning:

> The GIL must now be held when allocator functions of PYMEM_DOMAIN_OBJ (ex: PyObject_Malloc()) and PYMEM_DOMAIN_MEM (ex: PyMem_Malloc()) domains are called.

https://docs.python.org/3.6/whatsnew/3.6.html#build-and-c-api-changes

The allocator can be set from the C API, but it's a bit complex for me. Setting PYTHONMALLOC is a workaround to get the tests running.

Obviously this doesn't just affect testing though. Another reason to replace the parser?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,7 @@ matrix:
    allow_failures:
       - python: "3.5"
       - python: "3.6"
+env:
+  global:
+    # workaround Python 3.6 incompatibility #238
+    - PYTHONMALLOC=malloc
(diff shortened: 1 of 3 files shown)


mgedmin:
> Obviously this doesn't just affect testing though. Another reason to replace the parser?

Oh yes!

cjmayo:
Not needed with bs4.

PR #244, opened 2019-04-19
Fix mistakes in changes to test_dummy.py and test_updater.py in 8f4acc31

cjmayo:
In test_dummy.py bytes import was left over from previous version of the change
In test_updater.py an instance of unicode() was left unchanged.

### DIFF
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -21,7 +21,7 @@ Test dummy object.
 import sys
 import unittest
 
-from builtins import bytes, str
+from builtins import str
 
 import linkcheck.dummy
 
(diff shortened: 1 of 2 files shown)


tests/test_dummy.py
```diff
@@ -21,7 +21,7 @@
 import sys
 import unittest
 
-from builtins import bytes, str
+from builtins import str
```
mgedmin: I thought the plan was to rename this to `str_text` everywhere?

anarcat: shouldn't this be fixed then?

anarcat:
the `str/text_str` change wasn't made in this patch, so screw it, it can be done later/separately.

PR #245, opened 2019-04-19
Import str as str_text from builtins when supporting transition

cjmayo:
Expected to be removed when the project moves to Python 3 only.

### DIFF
--- a/tests/configuration/test_config.py
+++ b/tests/configuration/test_config.py
@@ -21,15 +21,15 @@ Test config parsing.
 import unittest
 import os
 import linkcheck.configuration
-from builtins import str
+from builtins import str as str_text
 
 
 def get_file (filename=None):
     """Get file name located within 'data' directory."""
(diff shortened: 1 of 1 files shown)


anarcat:
okay, i'm confused by this one. i thought we agreed on `str`?

cjmayo:
Well, I was suggesting both in #228. Because I expect os.path.join() to return a string this one I think is a clear example of code that can be removed when Python 2 support is removed. It will be obvious to remove the builtin imports but working out which str() is needed in Python 3 and which can go could be more work, this way it should be clear and if not an error will be generated when the import is removed.

cjmayo:
> okay, i'm confused by this one. i thought we agreed on `str`?

Happy to discuss further (and change if needed). I've updated all the commits now, so hopefully the intent of the code is clear even if my explanation isn't.

mgedmin:
> i thought we agreed on str?

Personally, I was never happy with `str`.

PR #247, opened 2019-04-21
Make TestRobotsTxt Python 3.7 compatible

cjmayo:
urllib.parse.quote() moved from RFC 2396 to RFC 3986 for quoting URL
strings. “~” is now included in the set of reserved characters.

### DIFF
--- a/tests/test_robotstxt.py
+++ b/tests/test_robotstxt.py
@@ -18,6 +18,7 @@
 Test robots.txt parsing.
 """
 
+import sys
 import unittest
 import linkcheck.robotparser2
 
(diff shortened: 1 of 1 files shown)


tests/test_robotstxt.py
```diff
@@ -224,7 +225,8 @@ def test_access4 (self):
             "Disallow: /tmp",
             "Disallow: /a%3Cd.html",
             "Disallow: /a/b.html",
-            "Disallow: /%7Ejoe/index.html",
+            "Disallow: /%7Ejoe/index.html" if sys.version_info < (3, 7) \
+            else "Disallow: /~joe/index.html",
```
mgedmin: You don't need the `\` to continue the line, the surrounding `[`...`]` take care of that.

tests/test_robotstxt.py
```diff
@@ -249,7 +251,8 @@ def test_access5 (self):
             "Disallow: /tmp/",
             "Disallow: /a%3Cd.html",
             "Disallow: /a/b.html",
-            "Disallow: /%7Ejoe/index.html",
+            "Disallow: /%7Ejoe/index.html" if sys.version_info < (3, 7) \
+            else "Disallow: /~joe/index.html",
```
mgedmin: Likewise.

anarcat:
i've just merged this regardless of the cosmetic changes suggested by @mgedmin, those can be fixed separately.

PR #230, opened 2019-04-10
{python3_20} Python3: decode parts before submitting them to urllib.quote()

cjmayo:

PR #210, opened 2019-04-10
[WIP] Python3 Support

cjmayo:
**Do not merge - for tracking PRs only**

This branch is a rebase of that by @PetrDlouhy in #40, needed because master has moved on.

These commits were dropped because they are already in master:
* Python3: fix imports
* fixes for Python 3: fix proxysuport

These needed slight editing to apply because of commits that have since been made to master (code changes are the same):
* Python3: decode parts before submitting them to urllib.quote()
* fixes for Python 3 + Travis test: fix cookies

Typos in titles were fixed for:
* Python3: update makefiles to test in Pyhon3 in allow-failures mode
* Python3: fix url.enswith in url.py
* Python3: fix test_parser test for removing invalit entity
* Python3: fix default encoding in htmllib to allow parsig of entities

The commits in this branch are being submitted in separate PRs against master. PR titles start with {python3_nn} reflecting their branch name.  The submission of PRs is limited by the commits that will apply on the current master. Most PRs are single commits, when a commit will not apply on master an attempt is made to add it to the previous branch before stopping.

As well as tracking the PRs for merging hopefully this PR will demonstrate the final state of the CI tests. There are some additional commits that fix various warnings.

PR #249, opened 2019-04-21
[WIP] Convert HtmlParser to use Beautiful Soup

cjmayo:
**For merging after python3 branch**

This is derived from the work by @PetrDlouhy in #119 and like that PR also the work in #125.

It is based on the python3 branch in #210, and indeed does not work on Python 2. A CI test pass here indicates success on Python 3.5 and Python 3.6. Python 3.7 results are included but telnet tests will likely fail due to miniboa not being Python 3.7 compatible.

Currently there is a test fail in the python3 branch #210 described in #230.
However,  I have been able to install and run this under Python 3.7.


anarcat:
this begs the question of how long we actually want to support python 2. or is your goal eventually to backport this to py2 as well?

thanks for your work!

cjmayo:
> this begs the question of how long we actually want to support python 2. or is your goal eventually to backport this to py2 as well?

No, I think this is Python 3 only. There shouldn't be too many commits to review when it is all tidied up.
For the python3 branch I think (at the moment!) we still have to go with Python 2 & 3, else we will have no CI testing per commit.

I'm going to have more of a look at the parsing. This uses Beautiful Soup which is using an HTML parser and then creates a parser for the Beautiful Soup output. Maybe there can be fewer layers and get back the line numbers.

mgedmin:
> this begs the question of how long we actually want to support python 2.

I personally plan to stop supporting Python 2 when it gets EOLed by upstream, which AFAIR is going to happen on January 1st 2020.

cjmayo:
I've got line numbers working. Patch to Beautiful Soup submitted here:
https://bugs.launchpad.net/beautifulsoup/+bug/1742921

I'll update this branch in the coming days.

cjmayo:
Well, working with my patched Beautiful Soup (Python 3.7 Travis run), cloned to GitHub for use with tox, and easier to view:
https://github.com/cjmayo/beautifulsoup/commit/21f6f34307129d12f686b72e1c8b565fff4af922

Need to look at compatibility with current Beautiful Soup.

cjmayo:
Tests passing. I've also addressed the outstanding comments in #119, in a commit on the end for now.

I think that's it for this one for the time being, need to see what happens with Beautiful Soup line numbers and get the python3 branch merged.

There will be tidying up and reviewing to do. I've also selected the patched Beautiful Soup for Python 3.7 in requirements.txt, because it was easy to do. If Beautiful Soup is modified there will need to be another way of testing with old and new versions for a while.

anarcat:
here we need to decide what we do with our patched beautifulsoup. it's the only reason why the test suite is passing and i don't know if we want to carry a fork of this in the long term. upstream seemed open to the feature but hasn't responded to the patch since april, so i'm worried.

SchoolGuy:
@anarcat I could do a plus one and comment on that patch to bring it to their attention list. Just ping me the PR.

anarcat:
it's listed above, but hidden behind a "load more" button, here it is again: https://bugs.launchpad.net/beautifulsoup/+bug/1742921

cjmayo:
Beautiful Soup now supports line numbers (not yet released):
https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/changes/516?start_revid=516

Code on this PR tweaked to support it.

anarcat:
oh my, that's amazing! now the question is when will this be released in bs4, and do we need to wait for this to happen...

why do tests pass here? is it because we patch to use the latest bs4 from bzr upstream? or that we took out the line numbers from the test suite?

it would be nice to still be able to run with older bs4 releases, is that possible? that way we could release with this change rightaway without either breaking the test suite or having to wait for the upstream release....

cjmayo:
I've created a GitHub mirror with the latest bs4 code + a modified version number, tox is using that for Python 3.7 tests:
https://github.com/linkchecker/linkchecker/blob/24e8b89c499d8850579750c03794f3bdf87c9c2b/tox.ini#L26

but not  Python 3.5 or 3.6 (hence the modified release number so that you can tell from the Travis log).

test_all_parts.py is checking whether bs4 supports line numbers:
https://github.com/linkchecker/linkchecker/blob/24e8b89c499d8850579750c03794f3bdf87c9c2b/tests/checker/test_all_parts.py#L26

and using the appropriate results file by skipping a test. Maybe there is a better way to code that.

The results file used if bs4 does not have line number support shows that in this case you get values of None :
https://github.com/linkchecker/linkchecker/blob/9f824cf40347aa1c6b0e1809340c3eb076efcbad/tests/checker/data/all_parts.html.result#L6

anarcat:
this looks pretty good, although it would be better to source directly from the bs4 source tree than your fork... do they mirror on github or provide dev snapshots in some way?

awesome work.

cjmayo:
I haven't researched, the issue being that bzr still has:
```py
__version__ = "4.8.0"
```
so you wouldn't be able to tell from the Travis log which version you had downloaded.

Cheers.

cjmayo:
OK #230 in. Votes please for one of:
1) lots more individual PRs
2) tidying up this one and merging in
3) tidying up #210 and merging before this one

anarcat:
not sure i know very well the impact of each enough for my vote to be relevant but...

wouldn't merging this before py3 make the latter easier to do since we'd be removing a shit ton of code?

then again, bs4 is not absolutely necessary for our survival, in the short term: py3 is, however. so i'd be tempted to do that first... 

so I'd go #210.

mgedmin:
I like lots of tiny PRs, those are fun and easy to review.

mgedmin:
> I like lots of tiny PRs, those are fun and easy to review.

(In fact I feel tempted to create some by taking a test that fails on Python 3 and trying to make it pass, but I don't want to duplicate work unnecessarily.)

cjmayo:
> wouldn't merging this before py3 make the latter easier to do since we'd be removing a shit ton of code?

It's an interesting challenge. But this one only works on Python 3, see https://github.com/linkchecker/linkchecker/pull/249#issue-272234457. I must admit I have already forgotten why.

> then again, bs4 is not absolutely necessary for our survival, in the short term: py3 is, however. so i'd be tempted to do that first...

And we have to switch out the html parser to get running on Python 3.

> so I'd go #210.

I've updated #210, based on the current master (the testing only passes there because the HtmlParser tests are skipped).

cjmayo:
> > I like lots of tiny PRs, those are fun and easy to review.
> 
> (In fact I feel tempted to create some by taking a test that fails on Python 3 and trying to make it pass, but I don't want to duplicate work unnecessarily.)

As mentioned I have updated #210. Do you want to take a look at that and work out whether you think it is viable to follow the test based approach instead?

cjmayo:
Dropped:
9f286807 ("fix flush tests after changing the parser", 2018-01-15)
Tests pass fine as they are.

cjmayo:
Tidied and updated, based on and dependent on how we go with #308 and #311. Also includes #312 and #313

cjmayo:
Rebased on the latest master, and because Beautiful Soup 4.8.1 with line number support has been released, removed the use of my bs4 copy repository. Python 3.5 tests are running against an older release to check we still work without line numbers. Python 3.6 and Python 3.7 use the latest bs4.

Next we need to confirm whether the new test in #311 (which is included here) reflects the way we want to go with quoting - both the behaviour and the use of encoding with unquote() which makes us Python 3 only.

tests/checker/test_mail_good.py
```diff
@@ -145,15 +145,15 @@ def test_valid_mail4 (self):
 
     @need_network
     def test_unicode_mail (self):
-        mailto = u"mailto:�lvin@users.sourceforge.net"
+        mailto = b"mailto:\xf6lvin@users.sourceforge.net".decode("iso-8859-1")
```
mgedmin: Likewise.

tests/test_parser.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: iso-8859-1 -*-
+# -*- coding: utf8 -*-
```
mgedmin: :sparkles: :sparkle: :sparkles:

tox.ini
```diff
@@ -1,17 +1,36 @@
...
+    miniboa >= 1.0.8
     biplist
+
+[testenv]
+usedevelop = true
+deps =
+    -rrequirements.txt
+    {[base]deps}
 commands =
     py.test {posargs:--tb=short --cov=linkcheck tests}
```
mgedmin: IIRC new versions of pytest drop the `py.test` executable and want you to use `pytest`.

(Are we pinning an old version of pytest in requirements.txt, or have I misremembered and the old py.test command alias still works in recent versions?)

tox.ini
```diff
@@ -1,17 +1,36 @@
...
 setenv =
     LC_ALL=en_US.utf-8
+
+[testenv:py35]
+# test bs4 without line number support
```
mgedmin: I would rather use a separate `py35-oldbs4` for this.  Time will come when we drop support Python 3.5, and I don't know if we want to tie dropping of older beautifulsoup versions with that.

You could use tox's [factors](https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings) to adjust dependencies for *-oldbs4 environments.

tox.ini
```diff
@@ -1,17 +1,36 @@
...
+    beautifulsoup4 < 4.8.1
+    requests >= 2.4
+    pyxdg
+    dnspython
+    future
+    {[base]deps}
+commands =
+    {[testenv]commands}
+setenv =
+    {[testenv]setenv}
```
mgedmin: commands/setenv/usevelop will be inherited from [testenv] by default so you don't need to specify those again.

(And if you go with my suggestion to use factors, you won't need a separate [testenv:py35] section either.)

tox.ini
mgedmin: Also, if we care about this, we should have a

```yaml
matrix:
  include:
    - name: old BeautifulSoup4
      python: "3.7"
      script: tox -e py37-oldbs4
```

or similar to add a Travis job for this as well.

mgedmin:
My experiments show that

```
>>> from future.moves.urllib import parse
>>> help(parse.unquote)
```
gives me the wrong unquote (no encoding argument), but

```
>>> from future.standard_library import install_aliases
>>> install_aliases()
>>> from urllib import parse
>>> help(parse.unquote())
```

works.  I'm confused.

mgedmin:
install_aliases() reduces the number of failures but is not a panacea: the backported version doesn't like  byte strings.

Makefile
```diff
@@ -76,9 +75,7 @@ locale:
 
 # to build in the current directory
 localbuild: MANIFEST locale
-	$(MAKE) -C linkcheck/HtmlParser
 	$(PYTHON) setup.py build
-	cp -f build/lib.$(PLATFORM)-$(PYVER)*/linkcheck/HtmlParser/htmlsax*.so linkcheck/HtmlParser
 	cp -f build/lib.$(PLATFORM)-$(PYVER)*/linkcheck/network/_network*.so linkcheck/network
```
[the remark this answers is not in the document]
cjmayo: Hopefully it can be removed but needs investigation.

linkcheck/HtmlParser/htmllib.py
```diff
@@ -43,7 +45,8 @@ def _print (self, *attrs):
         @type attrs: tuple
         @return: None
         """
-        print >> self.fd, self.mem, attrs
+        self.fd.write(self.mem)
+        self.fd.write(str_text(attrs))
```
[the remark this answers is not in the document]
cjmayo: It's only in used in a debug script, and a test which compares the output of two uses of the HtmlPrinter. So, hopefully we are OK with this slight change in behaviour.

cjmayo:
Yes, a lot of "valid" results get marked as "error".

tests/checker/test_charsets.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def test_utf8(self):
+        self.file_test("charsets/utf8.html")
+
+    def test_iso8859_2(self):
+        self.file_test("charsets/iso8859-2.html")
+
+    def test_cp1250(self):
+        self.file_test("charsets/cp1250.html")
```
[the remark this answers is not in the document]
cjmayo: The output is compared automatically with the file:
https://github.com/linkchecker/linkchecker/blob/9fdd7e7197c663ffb4eda8fba4191e1a5cc0aeed/tests/checker/data/charsets/cp1250.html.result

Makefile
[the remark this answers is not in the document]
cjmayo: Created #315 and included here for now to speed up testing.

linkcheck/plugins/anchorcheck.py
```diff
@@ -47,7 +49,7 @@ def check_anchor(self, url_data):
         A warning is logged and True is returned if the anchor is not found.
         """
         log.debug(LOG_PLUGIN, "checking anchor %r in %s", url_data.anchor, self.anchors)
-        enc = lambda anchor: urlutil.url_quote_part(anchor, encoding=url_data.encoding)
+        enc = lambda anchor: parse.quote(anchor)
```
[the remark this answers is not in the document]
cjmayo: I've looked again and removed the lambda, it was only being used once and is now very simple.

tests/checker/test_mail_good.py
```diff
@@ -88,15 +88,15 @@ def test_good_mail (self):
     def test_warn_mail (self):
         # some mailto addrs with warnings
         # contains non-quoted characters
-        url = u"mailto:calvin@users.sourceforge.net?subject=���"
-        qurl = self.norm(url)
+        url = b"mailto:calvin@users.sourceforge.net?subject=\xe4\xf6\xfc".decode("iso-8859-1")
```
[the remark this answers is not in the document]
cjmayo: Changed as suggested.

tests/checker/test_mail_good.py
cjmayo: Changed as suggested.

tox.ini
cjmayo: No pinning here. `py.test` still works with pytest 4.5.0. Tempting to just change it, but it seems like something extra.

tox.ini
cjmayo: > I would rather use a separate `py35-oldbs4` for this. Time will come when we drop support Python 3.5, and I don't know if we want to tie dropping of older beautifulsoup versions with that.

Add a separate py35-oldbs4.

> You could use tox's [factors](https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings) to adjust dependencies for *-oldbs4 environments.

The complication is including requirements.txt. You can't override dependency versions by specifying the same package with a different version in the tox.ini.

cjmayo: > Also, if we care about this, we should have a
> 
> ```yaml
> matrix:
>   include:
>     - name: old BeautifulSoup4
>       python: "3.7"
>       script: tox -e py37-oldbs4
> ```
> 
> or similar to add a Travis job for this as well.

The way it is now py35 and py35-oldbs4 get run by Travis as the Python 3.5 test. Would adding a matrix help? e.g. to split them into two separate tests?

tox.ini
cjmayo: Removed redundant commands and setenv as suggested.

tox.ini
cjmayo: > The complication is including requirements.txt. You can't override dependency versions by specifying the same package with a different version in the tox.ini.

I see setup.py has install_requires as well. So perhaps OK to not use requirements.txt in tox.ini. (Does seem awkward having to maintain both setup.py and requirements.txt).

tox.ini
mgedmin: I probably misread the "Removed all py.test-X* entry points" note in pytest 3.0.0's changelog and ended up with the wrong impression.

tox.ini
mgedmin: > The complication is including requirements.txt. You can't override dependency versions by specifying the same package with a different version in the tox.ini.

Oh well, scratch that idea, I guess.

+1 for not using requirements.txt in tox.ini (in a separate PR).  Our users (if we ever get any) will install linkchecker with pip install, which will totally ignore our requirements.txt and go along with install_requires from setup.py.  We should test what our users are getting instead of whatever we happened to pin when we found a working set of versions.

cjmayo:
Bit of a status update. I believe all the comments have been addressed (except Python 2 compatibility). It has been rebased on the latest master (I like the `-ra` option to pytest and have started using that for my local testing too).
At that point it turned out that PdfParser didn't work on Python 3, hence #332 which I have included here, but it's broken Travis because the old version of pdfminer work on Python 2 and not Python 3, something for tomorrow...

mgedmin:
(The annoying thing -- for me -- is that I don't remember if I reviewed all this code, or if I just glanced at it here and there, given the WIP in the title and the dropped 2.7 compat.  I do want to give 2.7 another try anyway.)

mgedmin:
@cjmayo the following patch makes tests work on Python 3:

```diff
diff --git a/tox.ini b/tox.ini
index 201d9fc7..f0108da2 100644
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,8 @@ envlist = py{35,36,37,35-oldbs4}
 deps =
     pyftpdlib
     parameterized
-    pdfminer < 20191010
+    py27: pdfminer < 20191010
+    !py27: pdfminer
     pytest-xdist
     pytest-cov
     miniboa >= 1.0.8
```

with the exception for the two tests/test_clamav.py tests that are (correctly) broken, but will probably be skipped on your machine if you don't have clamav-daemon installed and running.

We've added you as a member to the linkchecker organization on GH a while back; does that give you enough permissions to push to this repo?  If so, it would be convenient to have you PRs based on branches here, so we could all collaborate.  If not, then that's something we ought to fix.

mgedmin:
I've fixed the clamav tests (and the pdfminer issue) in https://github.com/linkchecker/linkchecker/pull/333.

cjmayo:
It seemed like it was going to so simple to rebase on master after the latest changes including the pdfminer tox fix, but alas enabling the sitemap test resulted in another failure. My available time has been taken up with #335 (which is included here to get the green tick on this one) - hopefully I will be able to respond to other public and private questions before too long.

linkcheck/parser/sitemap.py
```diff
@@ -29,22 +29,26 @@ def __init__(self, tag):
...
         """Parse XML URL data."""
         self.url_data = url_data
-        self.loc = False
+        self.in_tag = False
         self.url = u""
-        data = url_data.get_raw_content()
+        data = url_data.get_content()
         isfinal = True
         try:
             self.parser.Parse(data, isfinal)
```
mgedmin: This explodes on Python 2 as soon as the sitemap.xml contains any non-ASCII characters.  Try this:

```diff
diff --git a/tests/checker/data/sitemap.xml b/tests/checker/data/sitemap.xml
index 30e23048..b0ae9a20 100644
--- a/tests/checker/data/sitemap.xml
+++ b/tests/checker/data/sitemap.xml
@@ -6,4 +6,10 @@
       <changefreq>monthly</changefreq>
       <priority>0.8</priority>
    </url>
+   <url>
+      <loc>http://www.example.com/?ascii=nø</loc>
+      <lastmod>2005-01-01</lastmod>
+      <changefreq>monthly</changefreq>
+      <priority>0.8</priority>
+   </url>
 </urlset> 
```

cjmayo: > Apologies for creating extra work for you, but the regression test I thought I enabled wasn't actually testing the thing I thought it was testing -- passing unicode data to the expat parser on Python 2 "works" as long as that unicode data contains only characters in the ASCII subset.

Enabling more tests is great (and has proved its value). I came here to help get Python 3 supported, it's all part of the same effort.

cjmayo:
> We've added you as a member to the linkchecker organization on GH a while back; does that give you enough permissions to push to this repo? If so, it would be convenient to have you PRs based on branches here, so we could all collaborate. If not, then that's something we ought to fix.

I tried to push to https://github.com/linkchecker/linkchecker.git and got a permission denied response.

I'm keen to not be the sole "keeper" of these big branches (and rate limiter). So am happy to store them here. To date there has been a fair bit of rebasing going on which could have made sharing difficult but I think I am pretty much done on this one now (there is hope!).

mgedmin:
> I tried to push to https://github.com/linkchecker/linkchecker.git and got a permission denied response.

So, being a member of the organisation does not grant one write access to the org's repos!  Interesting.

I've created a new team ('developers'), invited you and @anarcat, and granted all team members write access to this repository.

cjmayo:
> I've created a new team ('developers'), invited you and @anarcat, and granted all team members write access to this repository.

That worked. I've done one last rebase to remove the sitemap commit that is now in master and pushed the branch to https://github.com/linkchecker/linkchecker.git.

What to do with this PR? It is not possible to change the source branch? I haven't updated it. Is a new PR needed to run CI on the new branch? (GitHub does show a couple of links to this PR when you select the new branch).

mgedmin:
Yes, a new PR is needed.  GitHub lets you change the target branch of the PR, but not the source branch/repo.

cjmayo:
OK, created #337 and closing this one.

PR #252, opened 2019-04-25
Make test_all_parts TestLogger import Python 3 compatible

cjmayo:
tests/checker/test_all_parts.py:21: in <module>
    import __init__ as init
E   ModuleNotFoundError: No module named '__init__'

testWarning: cannot collect test class 'TestLogger' because it has a
__init__ constructor

### DIFF
--- a/tests/checker/__init__.py
+++ b/tests/checker/__init__.py
@@ -36,6 +36,9 @@ class TestLogger (linkcheck.logger._Logger):
     """
     Output logger for automatic regression tests.
     """
+    
+    # don't attempt to collect this class because it has an __init__()
+    __test__ = False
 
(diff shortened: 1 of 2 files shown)


tests/checker/__init__.py
```diff
@@ -36,6 +36,9 @@ class TestLogger (linkcheck.logger._Logger):
     """
     Output logger for automatic regression tests.
     """
+    
+    # don't attempt to collect this class because it has an __init__()
+    __test__ = False
```
mgedmin: Nice!

anarcat:
i frankly have no idea what's going on here, but i'll trust this will not backdoor my machine. :p

PR #253, opened 2019-04-25
Fix a hanging test on Python 3

mgedmin:
I'm not entirely sure why the test is hanging, but this seems clear
enough:

- the test setup spawns a (non-daemon) background thread that runs
  forever, or until it is told to quit by receiving a TCP packet on a
  certain port
- the test teardown tries to tell the background thread to quit (which
  doesn't work) and waits for that to happen
- as a result the entire test run hangs forever

This commit adds a timeout as an extra safety net so that the test run
will complete even if the clean shutdown procedure fails for some
reason.

### DIFF
--- a/tests/checker/telnetserver.py
+++ b/tests/checker/telnetserver.py
@@ -50,7 +50,8 @@ class TelnetServerTest (LinkCheckTest):
     def setUp (self):
         """Start a new Telnet server in a new thread."""
         deadline = time.time() + 10 * TIMEOUT
-        self.port = start_server(self.host, 0, deadline=deadline)
+        self.port = start_server(self.host, 0, deadline=deadline,
+                                 testcase=self.id())
         self.assertFalse(self.port is None)
 
     def tearDown(self):
@@ -61,7 +62,7 @@ class TelnetServerTest (LinkCheckTest):
             pass
 
 
-def start_server (host, port, deadline):
+def start_server (host, port, deadline, testcase):
     # Instantiate Telnet server class and listen to host:port
     clients = []
     def on_connect(client):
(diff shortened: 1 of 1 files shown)


mgedmin:
> I'm not entirely sure why the test is hanging

This is probably because telnetlib.Telnet.write() expects a byte string but receives a plain str instead.  I'm sure a fix for that is waiting somewhere in the huge series of patches that comprise #210 and we'll get to it eventually, but I'd rather have a reliable test suite that doesn't hang even if bugs are present in the code.

tests/checker/telnetserver.py
```diff
@@ -49,7 +49,8 @@ def get_url(self, user=None, password=None):
 
     def setUp (self):
         """Start a new Telnet server in a new thread."""
-        self.port = start_server(self.host, 0)
+        deadline = time.time() + 2 * TIMEOUT
```
anarcat: is that 10 seconds? isn't it a little short for long running tests?

mgedmin: Turns out it is!

mgedmin: Bumped to 50 seconds.

mgedmin: The test should not be "long running": all it does is connect to localhost three times and emit an "exit" command each time. This should take milliseconds!  But that's another issue.

mgedmin: Oh, duh!  Linkchecker's telnet:// URL verification logic is
- wait for the string 'login: ' to appear with a 10-second timeout, then send the username
- wait for the string 'Password: ' to appear with a 10-second timeout, then send the password

and the integration test spawns a local telnet server that never sends the login/password prompts, forcing linkchecker to time out.

mgedmin: https://github.com/linkchecker/linkchecker/pull/254

anarcat:
hum. i thought it was either this PR or #254 but not both, did i miss something?

either way, thanks for your work :)

mgedmin:
In the end I didn't squash #254, so all the commits that were part of #253 got merged too...

PR #272, opened 2019-09-06
{python3_22} Python3: fix decode_parts function

cjmayo:


linkcheck/url.py
```diff
@@ -350,11 +350,14 @@ def url_norm (url, encoding=None):
...
-    if url.endswith('#') and not urlparts[4]:
+    if decode_for_unquote(url).endswith('#') and not urlparts[4]:
         # re-append trailing empty fragment
         res += '#'
     if encode_unicode:
-        res = unicode(res)
+        try:
+            res = unicode(res)
+        except NameError:
+            pass
```
mgedmin: Not a fan, but OK.

A comment next to the `encode_unicode:` check, explaining that this flag is used only for Python 2, because on Python 3 everything's already in unicode, would be nice.

anarcat: agreed. do have have a greppable marker for Python2 code that should be removed?

mgedmin: I don't remember, if I would git grep for `Python 2`, `PY2`, and `sys.version`, as well as some Python-2-only builtin names such as `unicode`, or, maybe better, `except NameError:`, to catch backwards compatibility code, if I wanted to remove it.

And thus I've accidentally conceded that no comment is necessary because `except NameError` is sufficiently unusual in Python code that it essentially only shows up in Python 2 compatibility code.

anarcat: okay, that makes sense, resolving, and please merge if you're okay with the result.

PR #273, opened 2019-09-06
{python3_23} fixes for Python 3 + Travis test: fix linkchecker tests

cjmayo:


linkcheck/director/console.py
```diff
@@ -22,15 +22,16 @@
 import os
 import time
 from .. import i18n, configuration, strformat, better_exchook2
+from builtins import str as str_text
 
 # Output to stdout and stderr, encoded with the default encoding
-stderr = i18n.get_encoded_writer(out=sys.stderr)
-stdout = i18n.get_encoded_writer()
+stderr = sys.stderr
+stdout = sys.stdout
```
anarcat: i don't get this change at all, won't this break in py2?

mgedmin: It might!

While Python 2's stdout supports Unicode output, any conversion errors get raised as exceptions (e.g. try print u'\u263b' after setting LC_ALL=C).  Linkchecker's i18n.get_encoded_writer() uses codecs.getwriter(encoding)(sys.stdout, errors='replace'), so instead of an error the user would see question marks instead of an error.

anarcat: so in other words this will break with non-UTF-8 locales showing UTF-8 contents...

why are we doing this then?

mgedmin: I would guess that Python 3 is unhappy if you try to wrap sys.stdout with a codecs.getwriter(encoding): the codes writer will try to write bytes into sys.stout, which expects (unicode) strings.

Suggestion: do a version check?

```python
if six.PY3:
    stdout = sys.stdout
    stderr = sys.stderr
else:
    stdout = i18n.get_encoded_writer()
    stderr = i18n.get_encoded_writer(out=sys.stderr)
```

(except we're not using six, are we?  does the `future` package provide some standartized helper to avoid writing `if sys.version_info()[0] >= 3:` everywhere?)

mgedmin: also, if the desired behavior is to write strings to stdou/stderr while ignoring encoding problems, then maybe i18n.get_encoded_writer() should be fixed to work on python 3.

All it would take, imho, is to `getattr(out, 'buffer', out)` so it would use sys.stdout.buffer/sys.stderr.buffer on python 3, but sys.stdout/sys.stderr on python 2.

anarcat: okay, then let's clarify this, if only through comments in the source code so we don't have to wrangle again this later.

mgedmin: I wonder: are there any tests that test this writing-despite-encoding-problems feature?  I'll experiment (on Python 2)

mgedmin: No tests fail if I use `stdout=sys.stdout; stderr=sys.stderr` on Python 2.

*sigh*

mgedmin: I've implemented my suggested fix in https://github.com/linkchecker/linkchecker/pull/new/python3_23_alt

It fixes the failures in tests/test_console.py that I think this PR was aiming to fix.

anarcat:
closed in favor of #277

PR #274, opened 2019-09-06
{python3_24} Python3: fix logger

cjmayo:


linkcheck/logger/__init__.py
```diff
@@ -323,7 +324,7 @@ def writeln (self, s=u"", **args):
         """
         Write string to output descriptor plus a newline.
         """
-        self.write(u"%s%s" % (s, unicode(os.linesep)), **args)
+        self.write(u"%s%s" % (s, str_text(os.linesep)), **args)
```
mgedmin: TBH `u"%s%s" % (s, os.linesep)` should work just as well.

anarcat: but not in py2?

mgedmin: no, it should work in all versions of Python.

On Python 2

- `u"...%s..." % str_containing_pure_ascii` is safe and returns Unicode.

- `u"...%s..." % str_containing_arbitrary_user_supplied_value` is dangerous and might raise a UnicodeDecodeError.

- (and `u"...{}...".format(str_containing_pure_ascii)` would always raise an exception.)

anarcat: okay, then let's just do that instead here, no?

mgedmin: Eh, it's a very minor thing and I already regret bringing it up.

mgedmin: (It's important to not use `unicode(os.linesep)` because `unicode` is not a defined name on Python 3.)

PR #277, opened 2019-09-09
Fix test_console.py on Python 3

mgedmin:
This is a alternative fix I suggested in the comments on PR #273.

### DIFF
--- a/linkcheck/i18n.py
+++ b/linkcheck/i18n.py
@@ -202,4 +202,6 @@ def get_encoded_writer (out=sys.stdout, encoding=None, errors='replace'):
     if encoding is None:
         encoding = default_encoding
     Writer = codecs.getwriter(encoding)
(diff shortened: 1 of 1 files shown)


cjmayo:
I have tested on Python 3 and see this fix failures with both test_console.py  and test_linkchecker.py.

Without this when test_linkchecker() calls the linkchecker executable with -V option, it results in:
```
  File "linkchecker", line 429, in <module>
    print_version()
  File "linkcheck/cmdline.py", line 41, in print_version
    console.print_version()
  File "linkcheck/director/console.py", line 161, in print_version
    configuration.ReleaseDate, file=out)
  File "/usr/lib64/python3.6/codecs.py", line 377, in write
    self.stream.write(data)
TypeError: write() argument must be str, not bytes
```

LGTM.

PR #278, opened 2019-09-09
{python3_27} Python3: fix csvlog

cjmayo:


linkcheck/logger/csvlog.py
```diff
@@ -19,7 +19,10 @@
 """
 import csv
 import os
-from io import BytesIO
+try:
+    from StringIO import StringIO
```
mgedmin: I would make that `from cStringIO import StringIO`, but it probably doesn't matter.

linkcheck/logger/csvlog.py
```diff
@@ -119,12 +122,21 @@ def log_url (self, url_data):
         self.writerow(map(strformat.unicode_safe, row))
         self.flush()
 
+    def encode_row_s(self, row_s):
+        try:
+            return row_s.encode("utf-8", self.codec_errors)
+        except(TypeError, AttributeError):
+            return row_s
```
mgedmin: This is one of those rare occasions where there should be a space in front of the `(`.

Why are you catching both TypeError and AttributeError?  On Python 3 I expect an AttributeError: str has no attribute encode.  On Python 2 I expect no error at all (the old code did not have any try/except).  When does the TypeError happen?

(Personally I would prefer a type check instead of try/except:

```python
     if isinstance(row_s, str):
        return row_s  # Python 3
    else:
        return row_s.encode(...)  # Python 2
```
but it's a matter of personal style, I think.)

linkcheck/logger/csvlog.py
cjmayo: That's on me. It was cStringIO but I changed it because I had found problems elsewhere (will post later) because cStringIO can't handle all Unicode. I think we probably should change this one back for now.

cjmayo: Another option: Python 2 now has io.StringIO, so we could use the same for both?

mgedmin: This is passed to csv.writer, right?  csv is a nasty module: it wants native strings (i.e. bytes on Python 2 and str on Python 3).  You can't write bytes to io.StringIO() on Python 2.  You might get away with using io.BytesIO() on Python 2 and io.StringIO() on Python 3.

I'm very surprised to hear that you've discovered a difference between StringIO and cStringIO.  IIRC both handle a mixture of bytes and unicode as long as that mixture is ASCII-only, but as soon as you stray off the safe path, both explode.

cjmayo: I've put from `cStringIO import StringIO` back.

linkcheck/logger/csvlog.py
cjmayo: Don't know why both exceptions are being caught (didn't modify this one - maybe I merged two together).
I've implemented the isinstance alternative as suggested.

PR #279, opened 2019-09-09
{python3_28} Python3: fix robotparser

cjmayo:


cjmayo:
No that's fine, do keep pointing them out. I did merge some already but there are a lot...

anarcat:
> No that's fine, do keep pointing them out. I did merge some already but there are a lot...

wait, actually those are *all* identical

cjmayo:
Ah yes, something went wrong in the script there, some quick force pushes and then back to checking the script...

cjmayo:
Hopefully, that is better - keep checking.

linkcheck/robotparser2.py
```diff
@@ -158,13 +159,13 @@ def parse (self, lines):
                     entry = Entry()
                     state = 0
             # remove optional comment and strip line
-            i = line.find('#')
+            i = decode_for_unquote(line).find('#')
             if i >= 0:
                 line = line[:i]
```
mgedmin: This is wrong:

```
$ python3
>>> line = 'ąčę# comment'.encode('utf-8')
>>> i = line.decode('utf-8').find('#')
>>> line[:i].decode('utf-8')
```

You would expect line[:i] to be everything until the # mark (i.e. 'ąčę'), but what you get is a UTF-8 character cut in half because the index is computed in Unicode code points, but the slicing is in bytes.

Suggested fix: `i = line.find(b'#')`

linkcheck/robotparser2.py
```diff
@@ -158,13 +159,13 @@ def parse (self, lines):
...
             line = line.strip()
             if not line:
                 continue
-            line = line.split(':', 1)
+            line = decode_for_unquote(line).split(':', 1)
```
mgedmin: This looks fine, although I question why the robots.txt files gets bytes as the input, and why is it hardcoding the charset to UTF-8 instead of relying on the Content-Type HTTP header.

linkcheck/robotparser2.py
cjmayo: Replaced as suggested.

linkcheck/robotparser2.py
cjmayo: Is this as simple as passing `decode_unicode=True` to iter_lines() in RobotFileParser.read()?

linkcheck/robotparser2.py
cjmayo: It breaks the Beautiful Soup branch:

```
>           i = line.find(b'#')
E           TypeError: must be str, not bytes

linkcheck/robotparser2.py:162: TypeError
```

cjmayo: Maybe a version of:
```
sep = b"?" if isinstance(urlparts[2], bytes) else u"?"
```
?

mgedmin: yeah, that seems fine!

linkcheck/robotparser2.py
mgedmin: yes, great idea!

linkcheck/robotparser2.py
cjmayo: But, all unnecessary once the input to parse() is always Unicode after decode_unicode=True.

My tests on Python 2 suggest find'ing a string in a Unicode string works fine.

cjmayo:
Converting to Unicode then hit a problem with Python 2 urllib.quote() so I added the below. Tests now pass, a bit concerned there might be others which aren't being tested, but don't have time to check now.

```diff
--- a/linkcheck/robotparser2.py
+++ b/linkcheck/robotparser2.py
@@ -32,7 +32,7 @@ import time
 
 import requests
 
-from . import log, LOG_CHECK, configuration
+from . import log, LOG_CHECK, configuration, url as urlutil
 
 __all__ = ["RobotFileParser"]
 
@@ -281,7 +281,7 @@ class RuleLine (object):
             # an empty value means allow all
             allowance = True
             path = '/'
-        self.path = parse.quote(path)
+        self.path = urlutil.url_quote_part(path)
         self.allowance = allowance
 
     def applies_to (self, path):
```

PR #40, opened 2017-02-13
Add Python 3 support

PetrDlouhy:
This PR is adding Python 3 support.

Not all tests are working yet, fell free to complete this PR.

Fix for #698

PR #280, opened 2019-09-09
{python3_29} fixes for Python 3: fix running problems in Python 3

cjmayo:


linkcheck/htmlutil/linkparse.py
```diff
@@ -86,13 +87,13 @@
 _quoted_pat = r"('[^']+'|\"[^\"]+\"|[^\)\s]+)"
 css_url_re = re.compile(r"url\(\s*(?P<url>%s)\s*\)" % _quoted_pat)
 swf_url_re = re.compile("(?i)%s" % urlutil.safe_url_pattern)
-c_comment_re = re.compile(r"/\*.*?\*/", re.DOTALL)
+c_comment_re = re.compile(b"/\*.*?\*/", re.DOTALL)
```
mgedmin: This is going to cause SyntaxWarnings on Python 3.8 because `\*` is not a valid escape in non-raw string literals.

(I sometimes use `[*]` instead of `\*` in regexes to avoid multilevel quoting madness.)

cjmayo: This change is actually reverted in a subsequent commit:
https://github.com/linkchecker/linkchecker/pull/40/commits/7c135790ed2f89915c6e39e5b19a4f2e083fe65b
I'll merge that in with this one. N.B. it also adds one other line.

cjmayo:
Updated to include changes from:
https://github.com/linkchecker/linkchecker/commit/7c135790ed2f89915c6e39e5b19a4f2e083fe65b

PR #281, opened 2019-09-09
{python3_30} Python3: fix decoding strings

cjmayo:


linkcheck/url.py
```diff
@@ -256,8 +256,8 @@ def url_fix_common_typos (url):
 
 def url_fix_mailto_urlsplit (urlparts):
     """Split query part of mailto url if found."""
-    if "?" in urlparts[2]:
-        urlparts[2], urlparts[3] = urlparts[2].split('?', 1)
+    if "?" in decode_for_unquote(urlparts[2]):
+        urlparts[2], urlparts[3] = decode_for_unquote(urlparts[2]).split('?', 1)
```
mgedmin: This changes the type of the strings stored in urlparts, but only sometimes!  I don't like this!  

This function is called only once, from url_norm() in the same file.  It gets passed the result of `list(urlparse.urlsplit(url)`.  `url` can be a byte string or it can be a unicode string, because of course *sigh*.

urlsplit() preserves the types of the arguments:

```
# python 3
>>> import urllib.parse as urlparse
>>> list(urlparse.urlsplit('mailto:foo@example.com'))
['mailto', '', 'foo@example.com', '', '']
>>> list(urlparse.urlsplit(b'mailto:foo@example.com'))
[b'mailto', b'', b'foo@example.com', b'', b'']
```

```
# python 2
>>> import urlparse
>>> list(urlparse.urlsplit(u'mailto:foo@example.com'))
[u'mailto', '', u'foo@example.com', '', '']
>>> list(urlparse.urlsplit('mailto:foo@example.com'))
['mailto', '', 'foo@example.com', '', '']
```

Therefore this function should do it too.

Suggestion:

```
    sep = b"?" if isinstance(urlparts[2], bytes) else u"?"
    if sep in urlparts[2]:
        urlparts[2], urlparts[3] = urlparts[2].split(sep, 1)
```

cjmayo: Changed as suggested.

PR #284, opened 2019-09-09
{python3_33} Python3: fix fileurl

cjmayo:


linkcheck/checker/fileurl.py
```diff
@@ -193,7 +193,7 @@ def check_connection (self):
             self.set_result(_("directory"))
         else:
             url = fileutil.pathencode(self.url)
-            self.url_connection = urlopen(url)
+            self.url_connection = urlopen(urlutil.decode_for_unquote(url))
```
mgedmin: I'm a bit lost to see we're carefully encoding the file name in the OS's filename encoding, and then immediately jumping back and decoding it as UTF-8.

mgedmin:
Let's go with #291 instead.

PR #286, opened 2019-09-09
{python3_35} Python3: fix unichr() in htmlparser

cjmayo:


linkcheck/HtmlParser/__init__.py
```diff
@@ -193,7 +194,7 @@ def _resolve_entity (mo):
         # unknown entity -> ignore
         return u''
     try:
-        return unichr(num)
+        return chr(num)
```
mgedmin: (but I only accept this because python 2 dies Real Soon Now and the "but chr() returns bytes on 2... oh, there's a hidden `from builtins` import elsewhere that changes everything" gotcha will die with it.)

anarcat: we'll catch this in the py2 cleanup, will we?

mgedmin: `except NameError`, so yes

PR #287, opened 2019-09-09
{python3_36} fixes for Python 3 + Travis test: fix cgi

cjmayo:


tests/test_cgi.py
```diff
@@ -58,8 +64,8 @@ def test_checklink (self):
 
     def test_application (self):
         form = dict(url="http://www.example.com/", level="0")
-        formdata = urllib.urlencode(form)
-        environ = {'wsgi.input': BytesIO(formdata)}
+        formdata = urllib_parse.urlencode(form)
+        environ = {'wsgi.input': StringIO(formdata)}
```
mgedmin: The WSGI spec ([PEP-3333](https://www.python.org/dev/peps/pep-3333/)) says wsgi.input is supposed to be a stream that returns bytestrings.  StringIO is wrong here, it should be BytesIO.

linkcheck/lc_cgi.py
```diff
@@ -54,7 +55,7 @@ def application(environ, start_response):
         request_body = environ['wsgi.input'].read(request_body_size)
     else:
         request_body = environ['wsgi.input'].read()
-    form = urlparse.parse_qs(request_body)
+    form = urlparse.parse_qs(request_body.decode())
```
mgedmin: Maybe `.decode(self.HTML_ENCODING)`?

This looks like a wsgi script that is meant to accept data from real users, so we should not crash in the presence of non-ASCII data.

tests/test_cgi.py
```diff
@@ -68,9 +71,9 @@ def start_response(status, headers):
             test_status[0] = status
             test_headers[0] = headers
         for str_data in application(environ, start_response):
-            if not isinstance(str_data, str):
+            if not isinstance(str_data, bytes):
                 err = "answer is not a byte string: %r" % limit(str_data, 30)
                 self.assertTrue(False, err)
-            test_response += str_data
+            test_response += str(str_data)
```
mgedmin: `str(bytes)` does the wrong thing on Python 3:

```
>>> str(b'hello')
"b'hello'"
```

While this test only looks for a substring and doesn't care about the extra b prefix and extra quotes, I'd prefer if we did it correctly.

(Which means `test_response = b""` and concatenating it all as bytes before calling `test_response.decode("UTF-8")`, since in theory the response may be split in the middle of a multibyte character.)

linkcheck/lc_cgi.py
cjmayo: Changed this (HTML_ENCODING it is not part of a class).

tests/test_cgi.py
cjmayo: Have implemented with `self.assertTrue(b"Generated by LinkChecker" in test_response)`. Only because decoding made the line too long. test_response isn't used subsequently.

PR #288, opened 2019-09-09
{python3_37} fixes for Python 3 + Travis test: fix cookies

cjmayo:


linkcheck/cookies.py
```diff
@@ -22,12 +22,12 @@
     from http.cookiejar import split_header_words
 except ImportError: # Python 2
     from cookielib import split_header_words
-try: # Python 3
-    from http.client import HTTPMessage
-except ImportError: # Python 2
-    from httplib import HTTPMessage
```
anarcat: what was wrong with this approach? why are we switching to the email module? this at least warrants a comment in the source.

linkcheck/cookies.py
```diff
@@ -58,17 +58,16 @@ def from_headers (strheader):
     @raises: ValueError for incomplete or invalid data
     """
     res = []
-    fp = BytesIO(strheader)
-    headers = HTTPMessage(fp, seekable=True)
-    if "Host" not in headers:
+    fp = StringIO(strheader)
+    headers = message_from_file(fp)
+    if "Host" not in dict(headers.items()):
```
mgedmin: Why not `if "Host" not in headers:`?

linkcheck/cookies.py
mgedmin: I would like to know too!

cjmayo: Commit comment added in #293.

cjmayo:
> I would like to understand what's happening here.
> 
> from_headers() is called only once, by from_file in the the same cookies.py. All bytestrings become unicode strings because from_file() opens the file in text mode.
> 
> Does HTTPMessage expect byte strings always and that is why you're switching to the email module? Why not give HTTPMessage byte strings by changing the `open(filename)` to `open(filename, 'rb')` in from_file()?

I could not make sense of this so decided to revert it and start myself from scratch and produced #249.

mgedmin:
Closing because #293 got merged.

PR #290, opened 2019-09-10
[WIP] Prepare test_parse.py for bs4 when using Python 2

cjmayo:
When testing Beautiful Soup on Python 2 multiple test_parse.py tests are failing with UnicodeEncodeErrors e.g.:
```
self = <linkcheck.HtmlParser.htmllib.HtmlPrettyPrinter object at 0x7f2a1566d310>, tag = 'a', attrs = {u'a\xe4': u'b'}, end = '>'

    def _start_element (self, tag, attrs, end):
...
        self.fd.write(u"<%s" % tag.replace("/", ""))
        for key, val in attrs.items():
            if val is None:
                self.fd.write(u" %s" % key)
            else:
>               self.fd.write(u' %s="%s"' % (key, quote_attrval(val)))
E               UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 2: ordinal not in range(128)

linkcheck/HtmlParser/htmllib.py:133: UnicodeEncodeError
```
---

This is what I was referring to in:
https://github.com/linkchecker/linkchecker/pull/278#discussion_r322589576

Maybe it is fixing the wrong thing - it is found by taking #249, removing the Python 3 only commits and testing with Python 2. There are still 4 other test failures.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -20,10 +20,7 @@ Test html parsing.
 
 import linkcheck.HtmlParser.htmlsax
 import linkcheck.HtmlParser.htmllib
-try:
-    from cStringIO import StringIO
-except ImportError:
-    from io import StringIO
+from io import StringIO
 import unittest
 
 from parameterized import parameterized


cjmayo:
Looking at the test results it is an after transition to bs4 change anyway.

tests/test_parser.py
```diff
@@ -20,10 +20,7 @@
 
 import linkcheck.HtmlParser.htmlsax
 import linkcheck.HtmlParser.htmllib
-try:
-    from cStringIO import StringIO
-except ImportError:
-    from io import StringIO
+from io import StringIO
```
mgedmin: So I don't see how this improves things?

io.StringIO will be used on Python 3 regardless.

cStringIO.StringIO is more permissive than io.StringIO on Python 2: it accepts both bytes and unicode, while io.StringIO requires unicode.

Where io.StringIO is better is when you want to have early problem detection when you accidentally write something that's not unicode into your StringIO, instead of having it explode later when you call .getvalue().

cjmayo: I'm afraid this was just the result of experimentation. All I can say is it fixed 9 test failures like the one above.

What I had seen in the Python 2 documentation for cStringIO.StringIO was:
"Unlike the StringIO module, this module is not able to accept Unicode strings that cannot be encoded as plain ASCII strings."

cjmayo:
Closing now to try and keep the list of pending PRs clear. Looks unlikely that we will be able to maintain Python 2 support. Can always revisit later.

PR #295, opened 2019-09-15
{python3_50} Python3: fix test_mail_good

cjmayo:


tests/checker/test_mail_good.py
```diff
@@ -146,7 +146,10 @@ def test_valid_mail4 (self):
     @need_network
     def test_unicode_mail (self):
         mailto = u"mailto:�lvin@users.sourceforge.net"
-        url = self.norm(mailto, encoding="iso-8859-1")
+        if isinstance(mailto, str):
+            url = self.norm(mailto, encoding="utf-8")
+        else:
+            url = self.norm(mailto, encoding="iso-8859-1")
```
anarcat: i don't get this either. this is a hardcoded string, with a hardcoded encoding, why would it vary? explain in a comment.

cjmayo: Without this change it fails as below. But maybe the real problem is in `linkcheck.url_norm()` which is also implicated in #296 and either its urlsplit needs replacing or the input always needs to be unicode.

```
$ tox -e py37 tests/checker/test_mail_good.py
...
self = <tests.checker.test_mail_good.TestMailGood testMethod=test_unicode_mail>

    @need_network
    def test_unicode_mail (self):
        mailto = u"mailto:ölvin@users.sourceforge.net"
        url = self.norm(mailto, encoding="iso-8859-1")
        resultlines = [
            u"url %s" % url,
            u"cache key %s" % mailto,
            u"real url %s" % url,
            u"valid",
        ]
>       self.direct(url, resultlines)

tests/checker/test_mail_good.py:156: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/checker/__init__.py:267: in direct
    self.fail_unicode(str_text(os.linesep).join(l))
tests/checker/__init__.py:244: in fail_unicode
    self.fail(msg)
E   AssertionError: Differences found testing mailto:%F6lvin@users.sourceforge.net
E   @@ -1,4 +1,4 @@
E    url mailto:%F6lvin@users.sourceforge.net
E   -cache key mailto:ölvin@users.sourceforge.net
E   -real url mailto:%F6lvin@users.sourceforge.net
E   +cache key mailto:�lvin@users.sourceforge.net
E   +real url mailto:%EF%BF%BDlvin@users.sourceforge.net
E    valid
----------------------------------------------------------------- Captured log call ------------------------------------------------------------------
WARNING  linkcheck.check:log.py:70 Sitemap URL 'mailto:%EF%BF%BDlvin@users.sourceforge.net' does not start with http: or https:.
```

anarcat: well in any case this needs to be clarified or fixed in the actual code, so I don't think this is ready for merging.

cjmayo:
I agree this change should not be implemented. I do have a solution I'll describe on #308.

PR #296, opened 2019-09-15
{python3_51} Python3: fix test_url by skipping iso8859-1 variant

cjmayo:


cjmayo:
```diff
+    def test_norm_quote_latin (self):
+        """
+        Check quoting URL with Latin 1 characters
+        This does not work on Python 3,
+        because urllib.parse.quote works only with UTF-8
```
Is it rather `urllib.parse.urlsplit` that is UTF-8 only?:

```
>>> urllib.parse.quote("ÄÆê".encode("latin1"))
'%C4%C6%EA'
>>> urllib.parse.urlsplit("ÄÆê".encode("latin1"))
  File "/usr/lib/python3.7/urllib/parse.py", line 418, in urlsplit
    url, scheme, _coerce_result = _coerce_args(url, scheme)
  File "/usr/lib/python3.7/urllib/parse.py", line 123, in _coerce_args
    return _decode_args(args) + (_encode_result,)
  File "/usr/lib/python3.7/urllib/parse.py", line 107, in _decode_args
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
  File "/usr/lib/python3.7/urllib/parse.py", line 107, in <genexpr>
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 0: ordinal not in range(128)
```

tests/test_url.py
```diff
@@ -131,13 +131,27 @@ def test_norm_quote (self):
         url = "http://example.com/?u=http://example2.com?b="
         nurl = "http://example.com/?u=http://example2.com?b="
         self.urlnormtest(url, nurl)
-        url = "http://localhost:8001/?quoted=�"
```
mgedmin: I hate this kind of thing in a test file where the contents of a variable depend on the encoding of the file.

The `# -*- coding: iso-8859-1 -*-` at the top implies this was "...quoted=\xfc".

tests/test_url.py
```diff
@@ -131,13 +131,27 @@ def test_norm_quote (self):
...
         url = "http://host/?a=b/c+d="
         nurl = "http://host/?a=b/c%20d%3D"
         self.urlnormtest(url, nurl)
 
+    @need_python2
+    def test_norm_quote_latin (self):
+        """
+        Check quoting URL with Latin 1 characters
+        This does not work on Python 3,
+        because urllib.parse.quote works only with UTF-8
```
mgedmin: No it doesn't:

```
>>> urllib.parse.quote(u"http://localhost/?q=\xfc", safe=":/?=", encoding="iso-8859-1")
'http://localhost/?q=%FC'
```

mgedmin:
I expect urlsplit() wants a (Unicode) string, not bytes.

cjmayo:
After changing to decoding to Unicode straight away I propose a different approach in #309.

cjmayo:
We don't need this. I do have a solution I'll describe on #308.

PR #297, opened 2019-09-15
{python3_40} Python3: fixes form checker/__init__.py

cjmayo:


linkcheck/checker/__init__.py
```diff
@@ -20,7 +20,10 @@
 
 import os
 import cgi
-import urllib
+try: # Python 3
+    from urllib import parse as urlparse
+except ImportError:
+    import urllib as urlparse
```
anarcat: i recently saw this code in another project... 

```
from six.moves import xmlrpc_client as xmlrpc
```

i wonder if it's soemthing we want to do instead of the try/except pattern? of course now it's a bit late to change this, but i figured it was better late than ever...

mgedmin: I ♥ six, but yeah, maybe let's not mix two different approaches.

In theory we'll rip out Python 2 code in a few months anyway.

cjmayo: Looking at it there is a `future` equivalent:
`from future.moves.urllib.parse import urlparse`
https://python-future.org/compatible_idioms.html#urllib-module

As said I guess we need to be consistent. Maybe their will be motivation left for one big cleanup...

anarcat: no big deal, don't worry about it. :)

PR #298, opened 2019-09-15
{python3_41} fixes for Python 3: fix markdowncheck

cjmayo:


linkcheck/plugins/markdowncheck.py
```diff
@@ -94,7 +95,7 @@ def applies_to(self, url_data, pagetype=None):
 
     def check(self, url_data):
         """Extracts urls from the file."""
-        content = url_data.get_content()
+        content = url_data.get_content().decode("utf-8")
```
mgedmin: But can we assume the data was encoded in UTF-8?

If the source of the data was a URL, then it should've been served with a Content-Type header declaring the correct encoding.  If the source of the data was a local file, then using the locale encoding would me more appropriate.

anarcat: the comment here seems to say it's "from a file" and it's the markdown plugin, so it might indeed be better to assume local encoding, although i always find that approach finnicky...

what do you think @cjmayo - did you have this in mind? how about documenting whatever decision was taken about this as a comment?

cjmayo: > the comment here seems to say it's "from a file" and it's the markdown plugin, so it might indeed be better to assume local encoding, although i always find that approach finnicky...
> 
That does sound like what an ordinary open() would do. I am thinking the cleanest solution is to put the decoding in the checker (#308) though. There is a FileUrl checker, so that could have a specialised version of the UrlBase decoding.

anarcat: decode early sounds like a better approach indeed. should we just close this in favor of #308 then?

mgedmin: :+1:

cjmayo: Yes. We still need the rest of this commit but it doesn't work on its own on master without the decoding.

cjmayo:
Last version of this PR included in #308.

PR #301, opened 2019-09-15
{python3_44} Python3: fixes for httpserver

cjmayo:


tests/checker/test_http.py
```diff
@@ -53,7 +53,7 @@ def _test_status(self, status):
         ]
         if status in (204,):
             resultlines.append(u"warning No Content")
-        if (status != 101 and status < 200) or status >= 400:
+        if (status not in [101, 102] and status < 200) or status >= 400:
```
mgedmin: Usually this sort of `in` check uses a tuple rather than a list on the right side; I don't really know why.  (Possibly tuples are more efficient, as if half a nanosecond mattered for a test function called a couple of times!)

Heh, you can see the `status in (204, )` check two lines above uses a tuple.

anarcat: more fundamentally, *what is this stuff*? why is the py3 port messing aroudn with http status codes?

cjmayo: I did a quick check and it appears to stop the test hanging.

anarcat: weird but okay

PR #302, opened 2019-09-15
{python3_45} Python3: fix loggers

cjmayo:


linkcheck/logger/csvlog.py
```diff
@@ -92,7 +92,7 @@ def log_url (self, url_data):
         if self.has_part("baseref"):
             row.append(url_data.base_ref)
         if self.has_part("result"):
-            row.append(url_data.result)
+            row.append(str(url_data.result))
```
mgedmin: These scare me.  What are they meant to handle?

unicode on Python 2?  str() could raise an exception

bytes on Python 3?  str() will add `b'` ... `'` and convert everything non-ascii to hex escapes, is that what we want in CSV logs?

ints/floats?  str() is fine (but why would url_data.result be a number?  I think this case is unlikely)

None?  do we want that to show up as "None" in the log?

mgedmin: All the items in the `row` list are filtered through `strformat.unicode_safe()`, why can't it handle whatever it was in url_data.result?

cjmayo:
My mistake. Sorry for the wasted time. I intended to drop this one:
https://github.com/linkchecker/linkchecker/pull/210#issuecomment-531591240

It doesn't appear necessary from testing or a simple search through the code.

anarcat:
no problems @cjmayo you're doing great.. :)

PR #303, opened 2019-09-15
{python3_46} Python3: fix test error message

cjmayo:


tests/checker/__init__.py
```diff
@@ -239,7 +239,8 @@ def file_test (self, filename, confargs=None):
     def fail_unicode (self, msg):
         """Print encoded fail message."""
         # XXX self.fail() only supports ascii
-        msg = msg.encode("ascii", "replace")
+        if isinstance(msg, bytes):
+            msg = msg.encode("ascii", "replace")
```
mgedmin: bytes don't have a .encode(), bytes only has a .decode().

self.fail() on Python 2 would fail with unicode data; self.fail() on Python 3 handles unicode just fine, and we don't want to convert assertion failure messages to bytes on Python 3.

I suggest something like

```python
    if not isinstance(msg, str) and isinstance(msg, str_text):  # this can be true only on Python 2
        msg = msg.encode("ascii", "backslashreplace")
```

cjmayo: Updated as suggested. Changed comment above to:

`# XXX self.fail() only supports ascii on Python 2`

PR #304, opened 2019-09-15
{python3_47} Python3: fix test_http_misc

cjmayo:


cjmayo:
If it has indeed been retrieved by HttpUrl then we should know the encoding. But there are other checker classes.
But Beautiful Soup does do content detection - I'm coming to the conclusion that this PR should be dropped and we stay in bytes and hand them to Beautiful Soup to sort out. Bit more work to do though before I am 100% on this.

cjmayo:
This one is a bad example for that though because links are just found in swf files with a regular expression and not Beautiful Soup. If we stay in bytes we have to encode the regexp - the same problem.

cjmayo:
Staying in bytes looks like a lot of effort, if it would even work. See #308 for an alternative approach.

cjmayo:
Superseded by #308.

PR #305, opened 2019-09-15
{python3_48} Python3: fix displaying help

cjmayo:


anarcat:
pooor pooooor windows

PR #306, opened 2019-09-15
{python3_49} enable and fix remaining bookmark tests

cjmayo:


tox.ini
```diff
@@ -10,6 +10,7 @@ deps =
     pytest-xdist
     pytest-cov
     miniboa
+    biplist
```
anarcat: what's this for?

cjmayo:
Added a commit message:
```
biplist module preferred for reading Safari bookmarks in
bookmarks/safari.py so install it for tox testing.
```

but arguably this should be a matrix with and without biplist to test all the code. Perhaps even a separate commit..

PR #308, opened 2019-09-20
Decode content when retrieved

cjmayo:
Not for merging yet, it needs reviewing and is based on the other two big branches. As such only the "Commits on Sep 20, 2019" should be considered.

The key change is to decode early in HttpUrl or UrlBase. HttpUrl uses requests to do the decoding, UrlBase tries UTF-8 and if that fails uses Beautiful Soup to detect the encoding.

This would replace #295, #296, #304 and simplify #298.

### DIFF
--- a/linkcheck/bookmarks/opera.py
+++ b/linkcheck/bookmarks/opera.py
@@ -63,9 +63,9 @@ def parse_bookmark_data (data):
     for line in data.splitlines():
         lineno += 1
         line = line.strip()
-        if line.startswith(b"NAME="):
+        if line.startswith("NAME="):
             name = line[5:]
-        elif line.startswith(b"URL="):
+        elif line.startswith("URL="):
             url = line[4:]
             if url and name is not None:
                 yield (url, name, lineno)
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -38,13 +38,16 @@ import time
 import errno
 import socket
 import select
-try:
-    from cStringIO import StringIO
-except ImportError:
-    # Python 3
-    from io import StringIO
+from io import BytesIO
 from builtins import str as str_text
 from future.utils import python_2_unicode_compatible
+from warnings import filterwarnings
+
+filterwarnings("ignore",
+    message="The soupsieve package is not installed. CSS selectors cannot be used.",
+    category=UserWarning, module="bs4")
+
+from bs4 import BeautifulSoup
 
 from . import absolute_url, get_url_from
 from .. import (log, LOG_CHECK,
@@ -216,6 +219,8 @@ class UrlBase (object):
         self.url_connection = None
         # data of url content,  (data == None) means no data is available
         self.data = None
+        # url content as a Unicode string
+        self.text = None
         # cache url is set by build_url() calling set_cache_url()
         self.cache_url = None
         # extern flags (is_extern, is_strict)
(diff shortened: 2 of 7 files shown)


cjmayo:
And the "Commits on Sep 21, 2019"... There is some ugliness in those that will need to be sorted but it is just enough to show that the tests can be passed.

cjmayo:
Should be easier to see the changes now. I realised that the key part will work on Python 2 with the old HTML parser. Still work to be done before it can be committed.

linkcheck/checker/httpurl.py
```diff
@@ -312,12 +316,12 @@ def read_content(self):
...
         maxbytes = self.aggregate.config["maxfilesizedownload"]
-        buf = BytesIO()
-        for data in self.url_connection.iter_content(chunk_size=self.ReadChunkBytes):
+        buf = StringIO()
+        for data in self.url_connection.iter_content(chunk_size=self.ReadChunkBytes, decode_unicode=True):
             if buf.tell() + len(data) > maxbytes:
                 raise LinkCheckerError(_("File size too large"))
             buf.write(data)
-        return buf.getvalue()
+        return buf.getvalue().encode("utf-8")
```
mgedmin: This is somewhat surprising: after we went though pains to decode into unicode, why encode back to bytes?

(I imagine because callers expect bytes, and it may take quite a few refactorings until we're using unicode internally everywhere?)

cjmayo: Yes, not nice. The get_content() from the parent class, UrlBase, is calling read_content() and expecting bytes. I'm having a go rewriting get_content() now (it wasn't finished anyway) but this is currently breaking the bookmark parsing...

cjmayo:
I think I got there. As noted in dde0fa255, the latest change to UrlBase:

functionality from get_content() has been split out into  get_raw_content() which returns "data" and download_content() which calls read_content() and sets the download related variables.
This allows for subclasses to do their own decoding and parsers to use bytes.

This has added a bit of complexity in methods calling other methods, but it is more robust than the previous attempt (evidence the tests passing now) and I think fairly logical. It has allowed dropping the double decoding of http content.

anarcat:
i guess we start using bs4 now? otherwise LGTM.

cjmayo:
Yes. This is the last step before moving to bs4. I'll get that branch tidied up and close the other PRs that are no longer relevant. Then we can see where we are and decide how to make the move.

cjmayo:
Ah, maybe the question was: does this mean we are committing to bs4 now? I can't see any other options because it provides the content detection as well. Although, the way bs4 is used in this PR to convert to Unicode might then allow not using bs4 as the html parser (but bs4 is a layer on top of one of a selection of parsers). Petr's additional character set tests https://github.com/linkchecker/linkchecker/pull/125/commits/b898ed518d99efe7535334434ae77792f16e5bba are a good test of finding the encoding - using chardet alone on them results in the wrong encoding with less than 50% confidence, whereas bs4 get's the right answer. So, short of dropping support for checking files and only supporting http I can't see another way.

cjmayo:
Let's pause this one until #295 and #296 are resolved. Looks like we need a couple of extra tests to check the correct encoding is used for quoting and probably means changes to always use bs4 for UrlBase and store the original encoding.

anarcat:
> Ah, maybe the question was: does this mean we are committing to bs4 now? 

Sure.

cjmayo:
> > Ah, maybe the question was: does this mean we are committing to bs4 now?
> 
> Sure.

I should have said as well the use of bs4 (at the end of this migration to Python 3, not just this PR) is only at the edges and not replacing the internals (which could be done). In theory it wouldn't be a lot of work to replace it with something else - I just can't see an alternative right now.

cjmayo:
My solution to the problems with quoting that #295, #296, #309 were trying to fix is to use the encoding argument that has been added to urllib.parse quote and unquote in Python 3. I have got this passing all tests, it still needs turning into a decent set of commits.

The catch there of course is "in Python 3". It does mean going Python 3 only, which also means going to bs4. I will update #249 so that the whole set can be seen and we can decide how to merge them.

I will send as many preparatory PRs as possible to reduce the size of the final set. If you want to wait to merge them until you have seen the whole thing then fair enough (any comments in the meantime on the PRs are welcome).

I've updated this one as the first of the preparatory PRs. It has got a bit simpler than before in that everything now goes through bs4 for encoding detecting rather than using requests for http. Maybe the requests encoding could be blended in as well in future. What you will see in the future PR for migrating to bs4 is storing the soup object created to avoid creating it twice.

anarcat:
> The catch there of course is "in Python 3". It does mean going Python 3 only, which also means going to bs4. I will update #249 so that the whole set can be seen and we can decide how to merge them.

Does that mean this PR drops support for Python 2?

Is there a way to keep doing Python 2 for at least one major release?

cjmayo:
> Does that mean this PR drops support for Python 2?

This PR doesn't. All tests still pass on Python 2 here.

> Is there a way to keep doing Python 2 for at least one major release?

I was trying to achieve that but with the quoting I can't see how.

PR #309, opened 2019-09-27
{python3_51_alt} Change TestUrl.test_norm_quote() to use UTF-8

cjmayo:
Latin 1 input will have now been converted to UTF-8 by the checker.

---

This merges on top of #308 and replaces #296.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -131,9 +131,9 @@ class TestUrl (unittest.TestCase):
         url = "http://example.com/?u=http://example2.com?b="
         nurl = "http://example.com/?u=http://example2.com?b="
         self.urlnormtest(url, nurl)
-        url = "http://localhost:8001/?quoted=ü"
-        nurl = "http://localhost:8001/?quoted=%FC"
-        self.urlnormtest(url, nurl, encoding="iso-8859-1")
+        url = b"http://localhost:8001/?quoted=\xfc".decode("iso-8859-1")
+        nurl = "http://localhost:8001/?quoted=%C3%BC"
+        self.urlnormtest(url, nurl)
         url = "http://host/?a=b/c+d="
         nurl = "http://host/?a=b/c%20d%3D"
         self.urlnormtest(url, nurl)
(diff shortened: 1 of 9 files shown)


tests/test_url.py
```diff
@@ -131,9 +131,9 @@ def test_norm_quote (self):
         url = "http://example.com/?u=http://example2.com?b="
         nurl = "http://example.com/?u=http://example2.com?b="
         self.urlnormtest(url, nurl)
-        url = "http://localhost:8001/?quoted=�"
-        nurl = "http://localhost:8001/?quoted=%FC"
-        self.urlnormtest(url, nurl, encoding="iso-8859-1")
+        url = b"http://localhost:8001/?quoted=\xfc".decode("iso-8859-1")
+        nurl = "http://localhost:8001/?quoted=%C3%BC"
+        self.urlnormtest(url, nurl)
```
mgedmin: Yes this will make the test pass, but it tests a different thing.

The thing I worry about: there's a web server somewhere that serves files encoded in ISO-8859-1 with non-URL-quoted non-ASCII links, and expects those links to be URL-quoted in ISO-8859-1.

I will stop worrying if anyone can demonstrate that serving an ISO-8859-1 HTML file with a non-ASCII link to a modern browser like Chrome or Firefox will cause that browser to open the link with the query parameter encoded as %3C%BC instead of %FC.

cjmayo: Does #308 then cause a bigger problem then in that we have Unicode without the context and don't know is should be ISO-8859-1?

mgedmin: I don't know.

I'm not familiar with the linkchecker codebase, which is why I tend to rely on its test suite (and view with suspicion any changes that might affect the meaning of the tests).

cjmayo: It's a good policy not to break the existing tests. I hadn't understood quite what the quoting was about - thanks for the insight.

cjmayo:
I agree this change should not be implemented. I do have a solution I'll describe on #308.

PR #311, opened 2019-10-01
Add new tests for URL quoting

cjmayo:
Check that the encoding detected in UrlBase is then used correctly to
quote URLs.

### DIFF
--- a/tests/checker/data/file_url_quote.html
+++ b/tests/checker/data/file_url_quote.html
@@ -1,0 +1,6 @@
+<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
+<a href="http://localhost:8001/?quoted=ü">l1</a>
+<a href="http://localhost:8001/?quoted=%FC">The same address as l1 therefore not reported</a>
+<a href="http://localhost:8001/?quoted=%FF">l2</a>
+<a href="mailto:ölvin@users.sourceforge.net">l3</a>
+<a href="mailto:%E4lvin@users.sourceforge.net">l4</a>
--- a/tests/checker/data/file_url_quote.html.result
+++ b/tests/checker/data/file_url_quote.html.result
@@ -1,0 +1,29 @@
+url file://%(curdir)s/%(datadir)s/file_url_quote.html
+cache key file://%(curdir)s/%(datadir)s/file_url_quote.html
+real url file://%(curdir)s/%(datadir)s/file_url_quote.html
+name %(datadir)s/file_url_quote.html
+valid
+
+url http://localhost:8001/?quoted=ü
+cache key http://localhost:8001/?quoted=%%C3%%BC
+real url http://localhost:8001/?quoted=%%C3%%BC
+name l1
+error
+
+url http://localhost:8001/?quoted=%%FF
+cache key http://localhost:8001/?quoted=%%C3%%BF
+real url http://localhost:8001/?quoted=%%C3%%BF
+name l2
+error
+
+url mailto:ölvin@users.sourceforge.net
+cache key mailto:ölvin@users.sourceforge.net
+real url mailto:%%C3%%B6lvin@users.sourceforge.net
+name l3
+valid
+
+url mailto:%%E4lvin@users.sourceforge.net
+cache key mailto:älvin@users.sourceforge.net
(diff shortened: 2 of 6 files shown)


cjmayo:
The idea of these is to check the new decode to Unicode using bs4 encoding detection. Some concerns:
* I don't understand what the cache entry is
* http URLs need to be marked as errors. I think this means an assert is firing off, perhaps because they are now Unicode, but I haven't worked out how to track it down yet.

Interestingly tests fail here because the existing code is detecting latin-1 as utf-8:

```
E    http://localhost:8001/?quoted=\xfc
E   -cache key http://localhost:8001/?quoted=%FC
E   -real url http://localhost:8001/?quoted=%FC
E   +cache key http://localhost:8001/?quoted=%C3%BC
E   +real url http://localhost:8001/?quoted=%C3%BC

E    url mailto:\xf6lvin@users.sourceforge.net
E    cache key mailto:\xf6lvin@users.sourceforge.net
E   -real url mailto:%F6lvin@users.sourceforge.net
E   +real url mailto:%C3%B6lvin@users.sourceforge.net
```

mgedmin:
Before we go too far out of our way to support preserving original encodings how about we verify that browsers actually do this?

Test plan:
- set up a web server that serves an HTML document with `<a href="ø.hmtl">test</a>` with a Content-Type: text/html; charset=ISO-8859-1 header
- set up a couple of sibling pages, one named ø.html in ISO-8859-1 with content "latin-1" the other named ø.html in UTF-8 with content "unicode"
- click link, see which page gets opened when you click link in Firefox and Chrome (and whatever other browsers you may have lying around)

I imagine this could be one with Apache, maybe nginx, or even a short custom Python web server.  Ideally using low-level WSGI code to avoid frameworks muddying up the issue by attempting to decode PATH_INFO into unicode for us.

mgedmin:
With

```python
#!/usr/bin/python3
from wsgiref.simple_server import make_server


def app(environ, start_response):
    status = '200 OK'
    headers = [('Content-Type', 'text/html; charset=ISO-8859-1')]

    start_response(status, headers)
    return [
        b'<p>PATH_INFO is %r.</p>' % environ.get('PATH_INFO'),
        b'<p><a href="/\xf8.html">test</a></p>'
    ]


with make_server('127.0.0.1', 8000, app) as server:
    print('Listening on http://127.0.0.1:8000')
    server.serve_forever()
```

I can see that clicking on a link encoded in ISO-8859-1 decodes it to UTF-8 and sends the UTF-8 path to the HTTP server in both Firefox and Chromium.

mgedmin:
gnome-web and links behave the same way.

I don't have lynx or Opera installed, nor a handy Windows system to try Edge.

w3m opens /%F8.html, but it is not a mainstream browser.

cjmayo:
A few extra results.

**On Linux:**
vivaldi is the same, displays `http://127.0.0.1:8000/ø.html` in the address bar and does
`"GET /%C3%B8.html HTTP/1.1"`

links: `"GET /ø.html HTTP/1.1"`

**On Windows:**
Edge is the same: `"GET /%C3%B8.html HTTP/1.1"`

---

So, it looks like it's OK to always encode in utf-8?

I'm also looking at adding variants where the links are already encoded as latin-1 to the test files. Existing linkchecker 9.4.0 just passes those through as they are, currently I am seeing the new code decode them again.

cjmayo:
Not quite correct - 9.4.0 passes on the latin-1 encoded http link but silently skips the mailto link.

cjmayo:
Hopefully with 9.4.0 that is the cache at work and it is just not reporting on a link that is a duplicate of another when decoded.

I've updated the tests to include one that is the same when decoded and four different, two encoded and two not.

Changing to utf-8 only output sorts out the quoting but there is still quite a number of unquotes and they need the encoding:

```
>>> from urllib import parse
>>> parse.unquote("%FF")
'�'
>>> parse.unquote("%FF", "latin1")
'ÿ'
```

mgedmin:
Oh!  The decoding is to handle links that are already-URL-encoded in the HTML source!  Now I get it!

(Also, I want to play again and see how browsers handle that.)

cjmayo:
Included in master with #337 here: 74d5c6809417ccb82.

PR #314, opened 2019-10-05
Replace memoized with functools.lru_cache and deprecations

cjmayo:
To merge only once Python 3 only and based on #249.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,8 +31,8 @@ Change it very carefully.
 from __future__ import print_function
 import sys
 if not (hasattr(sys, 'version_info') or
-        sys.version_info < (2, 7, 0, 'final', 0)):
-    raise SystemExit("This program requires Python 2.7 or later.")
+        sys.version_info < (3, 5, 0, 'final', 0)):
+    raise SystemExit("This program requires Python 3.5 or later.")
 import os
 import re
 import codecs
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -23,5 +23,4 @@ include doc/Makefile
 include doc/examples/*.sh doc/examples/*.bat doc/examples/*.py
 include doc/examples/linkcheckerrc_loginurl
 include scripts/*.sh scripts/*.py
-recursive-include third_party *.py *.txt Makefile example*
 recursive-include tests *.py *.result *.html *.ico *.txt *.zip *.asc *.css *.xhtml *.sqlite *.adr *.swf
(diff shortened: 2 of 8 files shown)


tests/test_parser.py
```diff
@@ -280,10 +282,9 @@ def test_flush (self, _in, _out):
 
     def test_entities (self):
         # Test entity resolving.
-        resolve = linkcheck.HtmlParser.resolve_entities
         for c in "abcdefghijklmnopqrstuvwxyz":
-            self.assertEqual(resolve("&#%d;" % ord(c)), c)
-        self.assertEqual(resolve("&#1114112;"), u"")
+            self.assertEqual(html.unescape("&#%d;" % ord(c)), c)
+        self.assertEqual(html.unescape("&#1114112;"), b"\xef\xbf\xbd".decode())
```
mgedmin: I don't think it makes sense for linkchecker to have unit tests for non-linkchecker code.  We could just remove this test, since the code it was testing was removed.

(html is part of the standard library in Python 3.x, and I assume it comes from the future package on Python 2.)

mgedmin:
I've added WIP to the PR title because otherwise I keep forgetting that this is the "wait until Python 2 support is no longer needed" PR and keep reviewing it as something that was aiming to get merged soon.

tests/test_parser.py
cjmayo: test_entities() removed.

cjmayo:
Updated on current master and ready to go now (Green tick reported back from Travis!).

The commit "Replace HtmlParser.resolve_entities() with html.unescape()" that caused the review comments above has been removed, superseded by #349 which removed the code being updated.

Three new commits:
Remove third_party directory from MANIFEST.in
Remove unused function linkcheck.fileutil.write_file()
Raise minimum Python version to 3.5 in setup.py

some commentary in the first two.

anarcat:
so wait, we're going 3.5+ only now? i guess i missed that switch... i thought the BS4 and other py3 merges kept 2.7 compat...

cjmayo:
Back here,  e46fb7fe9cb18f02. The urllib URL quote (or was it unquote) functions only support an encoding in Python 3.

setup.py
```diff
@@ -31,8 +31,8 @@
 from __future__ import print_function
 import sys
 if not (hasattr(sys, 'version_info') or
-        sys.version_info < (2, 7, 0, 'final', 0)):
-    raise SystemExit("This program requires Python 2.7 or later.")
+        sys.version_info < (3, 5, 0, 'final', 0)):
+    raise SystemExit("This program requires Python 3.5 or later.")
```
mgedmin, line 35: TBH we should also add a `python_requires='>= 3.5'` to the `setup()` call.

mgedmin: → https://github.com/linkchecker/linkchecker/pull/352

PR #315, opened 2019-10-18
Remove unused code from network subpackage

cjmayo:
Includes the C extension. Haven't looked to see where it all stopped being used, hopefully all tests are going to pass...

PR #316, opened 2019-10-19
Don't add linkcheck_dns directory to sys.path

cjmayo:
This code was added in:
efbbb656 ("Remove python-dns conflict by moving the dns module into a custom subdirectory.", 2012-12-07)

Installation of linkcheck_dns stopped with:
0a13fae3 ("remove third party packages and use them as dependency", 2018-01-06)
.


cjmayo:
Looks like this needs to be after #315 to get the tests to pass (failing obfuscate_test is removed).

anarcat:
needs rebase on top of #315

cjmayo:
Rebased on master and passes now.

PR #318, opened 2019-10-21
Add instructions to install current release tag from git via pip

tkfu:
Signed-off-by: Jon Oster <jon.oster@here.com>


tkfu:
Just force-pushed a minor update; I had mistakenly put `9.4.0` when the real tag is `v9.4.0`.

PR #321, opened 2019-10-21
Wait for threads to exit after stopping them

### LINKED ISSUE #320: linkchecker exits with status code 0 after internal error
## Summary

While debugging #317 I noticed that linkchecker exits with status code 0 even when it encounters an internal error.

## Steps to reproduce

 1. Get a slightly older version of linkchecker (or introduce an error on purpose)
 2. Run it on a single HTML file
 3. Observe the exit code

## Actual result

Exit status 0

## Expected result

Exit status 2

## Environment

Ubuntu 19.10.
Python 2.7.16+

## Other notes

With the cunning use of print statements I noticed that the main thread checks logger.stats.internal_errors and exits before the "internal error" log message gets printed and the logger.stats.internal_errors counter incremented.  I'm not sure how that happens.

mgedmin:
This fixes a race condition where the main thread would check if any
internal errors happened and get back a 0 while a worker thread was
still busy printing the internal error message before incrementing the
counter.

Fixes #320.

My experiments show that this adds no perceptible delay to the script
runtime (on Linux).  More specifically, there already is an annoying
perceptible delay of about 1 second, but it's not caused by this change.

I thought about adding a regression test for this, but couldn't come up with a way to do it.

### DIFF
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -199,6 +199,8 @@ class Aggregate (object):
             self.cancel()
         for t in self.threads:
             t.stop()
+        for t in self.threads:
+            t.join(timeout=1.0)
 
     @synchronized(_threads_lock)
     def is_finished (self):


mgedmin:
This sort of bugfix probably should be mentioned in the changelog.  Do we have a changelog? (LOL I call myself a "maintainer" and I don't know the answer to that question.)

mgedmin:
I found a changelog, but it needs some TLC before we start adding new features/fixes to it: #322.

PR #328, opened 2019-10-22
Enable ClamAV integration tests on Travis CI

mgedmin:
Let's see if they pass!

I've seen [other people use clamav-daemon in Travis CI jobs](https://github.com/viper-framework/viper/blob/master/.travis.yml), so it might just work!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ python:
   - "3.6"
   - "3.7"
 # command to install dependencies
+addons:
+  apt:
+    packages:
+      - clamav-daemon
 install:
   - pip install tox-travis coveralls
 # command to run tests
@@ -16,7 +20,7 @@ notifications:
 after_script:
   - coveralls
 matrix:
-   allow_failures:
-      - python: "3.5"
-      - python: "3.6"
-      - python: "3.7"
+  allow_failures:
+    - python: "3.5"
+    - python: "3.6"
+    - python: "3.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ install:
   - pip install tox-travis coveralls
 before_script:
   # we need to download the clamav signature database before the daemon will run
-  - service clamav-freshclam stop
-  - freshclam
-  - service clamav-daemon restart
-  - service clamav-daemon status
+  - sudo service clamav-freshclam stop
+  - sudo freshclam
+  - sudo service clamav-daemon restart
+  - sudo service clamav-daemon status
   - while ! test -S /var/run/clamav/clamd.ctl; do printf "."; sleep 1; done
 # command to run tests
 script:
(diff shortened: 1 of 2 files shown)


mgedmin:
So, this works!

The downside is that the test runs now take 4 minutes instead of 1.  I think that's acceptable.

(The extra time is 0.5m for apt update && apt install; 1m for clamav db download; 1m for clamav-daemon to start up.  The rest is rounding error?  The numbers displayed in the travis log don't quite add up for me!)

.travis.yml
```diff
@@ -5,8 +5,19 @@ python:
...
+      - clamav-daemon
 install:
   - pip install tox-travis coveralls
+before_script:
+  # we need to download the clamav signature database before the daemon will run
+  - sudo service clamav-freshclam stop
+  - sudo freshclam
+  - sudo service clamav-daemon restart
+  - sudo service clamav-daemon status
+  - while ! test -S /var/run/clamav/clamd.ctl; do printf "."; sleep 1; done
```
anarcat: maybe if we wrap this all up in a single shell command it would be faster? (why) do we need to restart the daemon to fetch new entries?

mgedmin: I don't understand your comment about wrapping this up in a single shell command for speed.

We need to (re)start the daemon _after_ fetching the virus signature database because otherwise it won't be runnning!  The systemd service definition for clamav-daemon has

```
# Check for database existence
ConditionPathExistsGlob=/var/lib/clamav/main.{c[vl]d,inc}
ConditionPathExistsGlob=/var/lib/clamav/daily.{c[vl]d,inc}
```

which tells it to not start up if there virus DB is not present.

(I thought I commented about this somewhere on GitHub, but I can't find my comment.  Maybe it got lost; I had some disappearing public wifi problems during lunch today.)

mgedmin:
> in general, no objections for me, but i don't quite understand what this change brings in...

The two tests in test_clam.py used to be skipped, now they're passing.

> we still allow failure on this anyways, so it's unclear why we're doing this.

We don't allow _failure_ per se.  I could revert the virus signature change in the test file and we would see the test fail, because clamav changed the name for the test virus file.

What is true is that if the `@needs_clamav` check were to be changed in some way that makes it return False, then the clamav tests would start being skipped again, and we might not notice.  I think the risk is low, and I don't want to spend any more time overengineering this.

.travis.yml
anarcat: > I don't understand your comment about wrapping this up in a single shell command for speed.

i doubt it's a real factor, but i figured that each line in that script might be a different call. that could be my `Dockerfile` paradigm leaking though. ;) i don't know how travis works, it's just a silly assertion.

> We need to (re)start the daemon after fetching the virus signature database because otherwise it won't be runnning! The systemd service definition for clamav-daemon has
> [...]

makes sense. i guess i was hoping to shave of some seconds here by avoiding a needless restart, but if it's not running at all, i guess we need to bite the bullet anyways.

thanks for your work!

Issue #323, opened 2019-10-21
Review get_content() calls for possible type issues


mgedmin:
In the comments of bug #317 I've asked for a review of all `url_data.get_content()` calls to see if there are any other bugs introduced by the change in 5fc01455b7e2707509accaa58a586f7d1dc41443 that made it return unicode instead of bytes.

This is a tracking bug for that review:

- [x] linkcheck/checker/fileurl.py:            self.content_type = mimeutil.guess_mimetype(self.url, read=self.get_content)
- [x] linkcheck/checker/ftpurl.py:        self.content_type = mimeutil.guess_mimetype(self.url, read=self.get_content)
- [x] linkcheck/checker/httpurl.py:            parser.feed(self.get_content()) → https://github.com/linkchecker/linkchecker/pull/325
- [x] linkcheck/checker/httpurl.py:            rtype = mimeutil.guess_mimetype_read(self.get_content)
- [x] linkcheck/parser/\_\_init__.py:    for url, name, lineno in parse_bookmark_data(url_data.get_content()):
- [x] linkcheck/parser/\_\_init__.py:    for url, name in parse_bookmark_data(url_data.get_content()):
- [x] linkcheck/parser/\_\_init__.py:    for line in url_data.get_content().splitlines():
- [x] linkcheck/parser/\_\_init__.py:    for line in strip_comments(url_data.get_content()).splitlines():
- [x] linkcheck/parser/\_\_init__.py:    for mo in linkfinder(url_data.get_content()):
- [x] linkcheck/parser/sitemap.py:        data = url_data.get_content() → https://github.com/linkchecker/linkchecker/pull/330
- [x] linkcheck/plugins/markdowncheck.py:        content = url_data.get_content()
- [x] linkcheck/plugins/parsepdf.py:        data = url_data.get_content() → https://github.com/linkchecker/linkchecker/pull/324
- [x] linkcheck/plugins/parseword.py:        content = url_data.get_content() → https://github.com/linkchecker/linkchecker/pull/326
- [x] linkcheck/plugins/regexcheck.py:        content = url_data.get_content()
- [x] linkcheck/plugins/viruscheck.py:        data = url_data.get_content() → https://github.com/linkchecker/linkchecker/pull/329

mgedmin:
[...]

PR #329, opened 2019-10-22
WIP: Fix viruscheck plugin

mgedmin:
The clamav interface needs bytes, not unicode: see #323.

It would be nice if we had tests for this code.  That's why I'm putting WIP in the title.

I don't know if I'll find enough steam to actually try and add the tests in question.  If I do, they'll probably depend on #328 being merged.)


cjmayo:
Merged with test in #390.

PR #335, opened 2019-10-26
Fix XmlTagUrlParser and make Python 3 compatible

cjmayo:
URLs within a sitemap file were not being captured.

### DIFF
--- a/linkcheck/parser/sitemap.py
+++ b/linkcheck/parser/sitemap.py
@@ -29,7 +29,10 @@ class XmlTagUrlParser(object):
         self.tag = tag
         self.parser = ParserCreate()
         self.parser.buffer_text = True
-        self.parser.returns_unicode = True
+        try:
+            self.parser.returns_unicode = True
+        except AttributeError:
+            pass  # Python 3
         self.parser.StartElementHandler = self.start_element
         self.parser.EndElementHandler = self.end_element
         self.parser.CharacterDataHandler = self.char_data
@@ -37,7 +40,7 @@ class XmlTagUrlParser(object):
     def parse(self, url_data):
         """Parse XML URL data."""
         self.url_data = url_data
-        self.loc = False
+        self.in_tag = False
         self.url = u""
         data = url_data.get_raw_content()
         isfinal = True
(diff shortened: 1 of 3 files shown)


cjmayo:
It looks like sitemap checking never worked, this is an attempt to fix that as well as Python 3 support. Having said that I haven't done a successful check of a real world sitemap file.

I have gone back to `url_data.get_content()`, i.e. passing a string to `xmlparser.Parse()` and it is succeeding here on Python 2. @mgedmin I don't now what test data you used that failed, maybe this works because the test file states an encoding? Else we could pass the encoding from url_data to xml.parsers.expat.ParserCreate() and force the issue.

linkcheck/parser/sitemap.py
```diff
@@ -29,22 +29,26 @@ def __init__(self, tag):
...
         """Parse XML URL data."""
         self.url_data = url_data
-        self.loc = False
+        self.in_tag = False
         self.url = u""
-        data = url_data.get_raw_content()
+        data = url_data.get_content()
         isfinal = True
         try:
             self.parser.Parse(data, isfinal)
```
mgedmin: This explodes on Python 2 as soon as the sitemap.xml contains any non-ASCII characters.  Try this:

```diff
diff --git a/tests/checker/data/sitemap.xml b/tests/checker/data/sitemap.xml
index 30e23048..b0ae9a20 100644
--- a/tests/checker/data/sitemap.xml
+++ b/tests/checker/data/sitemap.xml
@@ -6,4 +6,10 @@
       <changefreq>monthly</changefreq>
       <priority>0.8</priority>
    </url>
+   <url>
+      <loc>http://www.example.com/?ascii=nø</loc>
+      <lastmod>2005-01-01</lastmod>
+      <changefreq>monthly</changefreq>
+      <priority>0.8</priority>
+   </url>
 </urlset> 
```

anarcat: sounds like we should add that to the commit and fix :)

cjmayo: I should have said I thought get_raw_content() was failing on Python 3 and that was why I changed it. However, it seems to work fine so I have restored it.

Thanks for explaining the problem with non-ASCII characters. I've added the suggested test. Passing here for Python 2 and also on #249 which I have identically updated.

PR #337, opened 2019-10-30
Change HtmlParser to use Beautiful Soup

cjmayo:
This PR replaces the HTML parsing code included in LinkChecker with Beautiful Soup 4 which enables us to support Python 3. It includes the outstanding commits from Petr Dlouhy in #40 and #119 and #125.

It is a continuation of #249, created because the branch has moved to this repository.

Currently it does not support Python 2 because urllib.unquote() does not accept an encoding argument. https://github.com/linkchecker/linkchecker/pull/249#issuecomment-541023273


linkcheck/checker/httpurl.py
```diff
@@ -84,8 +84,6 @@ def content_allows_robots (self):
         handler = linkparse.MetaRobotsFinder()
         parser = htmlsax.parser(handler)
         handler.parser = parser
-        if self.charset:
-            parser.encoding = self.charset
```
mgedmin: I have doubts about this entire commit.

If we have charset information from the Content-Type header, there should be a way to force the HTML parser to use it.  (The spec requires that if the document itself specifies charset X in a `<meta>` tag, but the server specifies charset Y in the Content-Type header, then the parser must use charset Y.)

linkcheck/httputil.py
```diff
@@ -44,6 +44,7 @@ def x509_to_dict(x509):
     }
     notAfter = x509.get_notAfter()
     if notAfter is not None:
+        notAfter = notAfter.decode()
```
mgedmin: There's no regression test for this fix.  :(

linkcheck/parser/__init__.py
```diff
@@ -128,9 +128,9 @@ def find_links (url_data, callback, tags):
     handler.parser = parser
     # parse
     try:
-        content = url_data.get_raw_content()
+        soup = url_data.get_soup()
         with parse_mutex:
```
mgedmin: Hey, we can drop the parse_mutex now!

mgedmin:
So, remember the browser tests I did in some other GH issue/PR and gleefully reported that modern browsers decode everything and treat all links as UTF-8?  Well, about that.

- xdg-open tests/checker/data/file_url_quote.html
- click the l3 link
- observe http://localhost:8001/?quoted=%FF in the address bar!

The tests added in this branch make linkchecker treat that link as http://localhost:8001/?quoted=%C3%BF!

This makes me nervous.  Linkchecker should behave the same way browsers do.

And it's not a difference between file:// vs http://, I can reproduce this even if I do

- python3 -m http.server -d tests/checker/data
- click file_url_quote.html
- click the l3 link
- observe http://localhost:8001/?quoted=%FF in the address bar!

I should dig up my earlier experiment with a custom wsgi script and figure out what makes Chromium behave differently.

linkcheck/checker/httpurl.py
cjmayo: It's straightforward to implement (add a get_content() method to HttpUrl that takes the encoding from requests and tells Beautiful Soup to use it). Harder to test all the permutations.

linkcheck/httputil.py
cjmayo: Indeed the limited https testing is completely disabled. There's a lot of code around certificates using sockets which probably needs ripping out and replacing with getting the certificate from requests. Best left until after the bs4 changes though.
This change is just a result of the pyOpenSSL API - which itself recommends using cryptography, and that would give us a nice datetime!

cjmayo:
> So, remember the browser tests I did in some other GH issue/PR and gleefully reported that modern browsers decode everything and treat all links as UTF-8? Well, about that.
> 
>     * xdg-open tests/checker/data/file_url_quote.html
> 
>     * click the l3 link
> 
>     * observe http://localhost:8001/?quoted=%FF in the address bar!
> 
> 
> The tests added in this branch make linkchecker treat that link as http://localhost:8001/?quoted=%C3%BF!
> 
I think this is the key reason for the WIP on this PR. (#311 demonstrates and is included here).

cjmayo:
How do we want to manage changes to this? PRs against the branch?
I guess avoiding reviewing the same thing twice is a key objective, but also having an understandable set of commits for the future. Have PRs with squash commits, and then rebase before merging into master?

linkcheck/httputil.py
cjmayo: Looks like there is no simple way to get the certificate from requests. Session.cert is the default client certificate (must update my local copy of the requests docs...).

cjmayo: I've included this commit with a test in #338.

cjmayo:
OK, rebased on master, just the tox.ini conflict as pointed out.

Two new commits on top as per the review above:
- Get rid of threading in `linkcheck/parser/__init__.py`
- Use the HTTP header encoding.

But now if fails testing... Beautiful Soup 4.8.2 changed how it handles DOCTYPEs:
https://bugs.launchpad.net/beautifulsoup/+bug/1848401

A bit of investigation required to decide what results we should expect.

mgedmin:
Those tests seem reasonable to me.

I've played a bit and the fix for the test problem seems very simple, so I pushed it to the branch.

mgedmin:
I was too hasty: my fix only works on newer Python version (3.6, 3.7, 3.8), because tox-travis runs the "py35-oldbs4" testenv on 3.5.

How about we drop support for old BeautifulSoup releases?

mgedmin:
> How about we drop support for old BeautifulSoup releases?

An alternative would be to drop the doctype tests when we detect an old BeautifulSoup release.

I don't think linkchecker itself cares about doctype anywhere, so it shouldn't result in any loss of function.

anarcat:
while it would be nice to keep on supporting 3.5, i don't consider that a requirement at this point. newer platforms enforcing python 3 are more likely to enforce python 3.6, 3.7 or 3.8 than 3.5.

dropping the test for older BS releases is fine with me too.

anarcat:
i guess we should remove the WIP here then?

cjmayo:
For me personally, and Gentoo Linux, no problem requiring beautifulsoup >= 4.8.2. If we did go that way I think we should remove the allow_failures from Travis. (and remove the alternative test from tests/checker/test_all_parts.py)

But seems extreme just to get the tests to pass when it shouldn't impact the user - the beautifulsoup parsing has been done by this point. Skipping the tests for beautifulsoup < 4.8.2 seems the best idea.

Maybe in future all the doctype code can just be ripped out but that would take time to check which probably isn't a priority now.

mgedmin:
> For me personally, and Gentoo Linux, no problem requiring beautifulsoup >= 4.8.2. If we did go that way I think we should remove the allow_failures from Travis. (and remove the alternative test from tests/checker/test_all_parts.py)

Not just allow_failures, the entire oldbs4 job from the travis matrix (and from tox.ini).

> But seems extreme just to get the tests to pass when it shouldn't impact the user - the beautifulsoup parsing has been done by this point. Skipping the tests for beautifulsoup < 4.8.2 seems the best idea.

+1

Does BeautifulSoup define its version in the bs4 package somewhere?  Or do we use pkg_resources to get its version?

mgedmin:
> Does BeautifulSoup define its version in the bs4 package somewhere? Or do we use pkg_resources to get its version?

Or maybe we can do a feature-based test somehow?

```python
bs_can_handle_doctypes_correctly = BeautifulSoup("<!doctype html>", "html.parser").something == something?
```

cjmayo:
> Does BeautifulSoup define its version in the bs4 package somewhere? Or do we use pkg_resources to get its version?

Haven't checked the docs but a quick grep got me:
```py
>>> bs4.__version__
'4.8.2'
```

Thinking about this a bit more - if it turns out to be too fiddly to isolate the DOCTYPE test strings, making the code compatible with all versions would only take something like:
```diff
--- a/linkcheck/HtmlParser/htmllib.py
+++ b/linkcheck/HtmlParser/htmllib.py
@@ -151,7 +151,7 @@ class HtmlPrettyPrinter (object):
         @type data: string
         @return: None
         """
-        self.fd.write("<!DOCTYPE%s>" % data)
+        self.fd.write("<!DOCTYPE %s>" % data)
 
     def pi (self, data):
         """
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -93,7 +93,8 @@ class Parser(object):
                             self.handler.comment(comment)
             elif isinstance(content, Doctype):
                 if hasattr(self.handler, 'doctype'):
-                    self.handler.doctype(content[7:])
+                    self.handler.doctype(content[8:] if
+                        content.upper().startswith('DOCTYPE ') else content)
             elif isinstance(content, Comment):
                 if hasattr(self.handler, 'comment'):
                     self.handler.comment(content.strip())
```

TBC, but I think it may never actually be called by running the `linkchecker` command. Part of the future work might even involve not just dropping the doctype handler but moving the HtmlPrettyPrinter class into the test suite.

cjmayo:
I missed the thumbs up - they don't generate an email notification.

@mgedmin do you want to do something with  the three "Split the oldbs4 into a separate Travis job" commits? By all means squash them or otherwise; the "allow_failures" needs to go?
And the "Fix doctype tests" commit, feel free to replace with the change in the comment above (or just remove it, I have the coment as a commit - I'm just not sure what to do with the Travis changes),

I've done some testing of moving and removing more parsing related code (including the doctype handling), nearly ready, but I think that would better wait for another PR.

mgedmin:
> I missed the thumbs up - they don't generate an email notification.

Sorry about that!  I meant to apply your suggestion and push it, but ran out of motivation.

> @mgedmin do you want to do something with the three "Split the oldbs4 into a separate Travis job" commits?

I didn't want to force-push to a shared branch.  We could squash them if you want to, but having them as separate commits doesn't bother me.

> By all means squash them or otherwise; the "allow_failures" needs to go?

Yes: allow_failures needs to go once the code is fixed.  I would commit the removal as a separate commit, but I won't mind if you want to play with some history rewriting instead ;)

> And the "Fix doctype tests" commit, feel free to replace with the change in the comment above (or just remove it, I have the coment as a commit - I'm just not sure what to do with the Travis changes),

I would prefer to have the oldbs4 job run as a separate Travis job, for two reasons:

- so its failures don't get mistaken with Python 3.5 compatibility problems
- so we don't accidentally drop it when we drop Python 3.5 support

tox-travis has been disappointing to me: I hoped to make the Python 3.5 job run just the py35 tox job without an explicit TOXENV override in the environment, but couldn't make it happen by configuring tox-travis in tox.ini.  https://tox-travis.readthedocs.io/en/stable/envlist.html#advanced-configuration did nothing.

Now I have a new idea: if the tox environment were called `oldbs4` without the `py35-` prefix, maybe tox-travis wouldn't run it by default on Python 3.5?  And then we can have a cleaner .travis.yml.

I'll experiment, stay tuned!

mgedmin:
Changes pushed.

I have no motivation to play with history rewriting (and I think it's interesting to see all the missteps made along the way).

Maybe let's merge this?

mgedmin:
The builds are green on Travis CI; I don't know why GitHub doesn't see that yet:
- https://travis-ci.org/github/linkchecker/linkchecker/builds/668711103
- https://travis-ci.org/github/linkchecker/linkchecker/builds/668711086

anarcat:
humm... mayeb we can rerun travis to fix that? (just close and reopen the PR)

my approval still stands, in any case, but i will let you folks do the honors. :)

mgedmin:
I reran a single job in one of the builds, and I don't see any changes here.

We can temporarily disable the branch protection, if there's consensus that this can be merged.

@cjmayo, what's your opinion?

anarcat:
i tried to kick CI, let's see if GH picks it up now.

should we consider switching to GH Actions? :)

mgedmin:
> should we consider switching to GH Actions? :)

Maybe?  I don't like them.

cjmayo:
One more idea for Travis. I had a simple commit available, to remove doc/python3.txt, I've pushed that (hopefully it doesn't make things worse...).

Anyway, all the tests I can run locally pass, I'm happy for it to be merged - removed the WIP!

cjmayo:
There are some problems reported on previous days at https://www.traviscistatus.com/
related to "repositories using the legacy Services integration"

> You can migrate your repositories to GitHub Apps on https://travis-ci.com/account/repositories.

mgedmin:
Uhh, why is the Travis CI service inactive?
Ekrano nuotrauka iš 2020-03-30 20-01-06

I've been postponing service → app migration in my personal projects in hopes GitHub or Travis will automate the process.  Travis is now telling me

> We are only able to migrate accounts that have 50 or fewer repositories using the Legacy Services Integration. Please refer to our documentation on how to migrate your account.

so I don't know if that'll ever happen.

The other link says

> Note: While this process is ongoing, we have been migrating from GitHub services to webhooks on the https://travis-ci.org enabled repositories. The closure of GitHub services will not affect your repositories even if you are currently on .org.

and yeah, we have a working webhook:

Ekrano nuotrauka iš 2020-03-30 20-08-02

This explains why the service was disabled.

anarcat:
> Maybe? I don't like them.

maybe we can move this in a different issue, but could you expand? :)

that said, travis still seems to not be running:

> Expected — Waiting for status to be reported

i can't help but wonder if travis is suffering *because* of GH actions.. they're bound to have more integration troubles than GH itself... :/ unfair, really...

mgedmin:
> maybe we can move this in a different issue, but could you expand? :)

No objective reason, just the UX feels bad.  (Like the checks tab in pull requests suddenly expands to take 100% of the screen width, while all the other tabs are narrow columns with wide margins on both sides.  Also the YAML I've seen feels redundant, repetitive, and the opposite of beautiful.)

> that said, travis still seems to not be running:

It is running, just not reporting back.  Like here's the PR build for cjmayo's latest commit: https://travis-ci.org/github/linkchecker/linkchecker/builds/668847616

mgedmin:
I've disabled the branch protection in Travis temporarily to merge this, then reenabled it for future PRs.

PR #338, opened 2019-11-05
Enable https checking using a test server

cjmayo:
Previously disabled because it was using Amazon. Verification is disabled because we generate our own certificate.

Also use the generated certificate to test httputil.x509_to_dict() and the fix required for Python 3.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,6 +3,7 @@ miniboa
 parameterized
 pdfminer
 pyftpdlib
+pyopenssl
 pytest
 pytest-xdist
 # for building on osx:
--- a/tests/checker/httpserver.py
+++ b/tests/checker/httpserver.py
@@ -20,7 +20,8 @@ Define http test support classes for LinkChecker tests.
 
 from html import escape as html_escape
 from http.server import SimpleHTTPRequestHandler, HTTPServer
-from http.client import HTTPConnection
+from http.client import HTTPConnection, HTTPSConnection
+import ssl
 import time
 import threading
 try:
@@ -29,6 +30,7 @@ except ImportError:
     import urllib as urllib_parse
 from io import BytesIO
 from . import LinkCheckTest
+from .. import get_file
 
 
 class StoppableHttpRequestHandler (SimpleHTTPRequestHandler, object):
@@ -175,19 +177,45 @@ class HttpServerTest (LinkCheckTest):
         return u"http://localhost:%d/tests/checker/data/%s" % (self.port, filename)
 
 
+class HttpsServerTest(HttpServerTest):
+    """
+    Start/stop an HTTPS server that can be used for testing.
+    """
 
-def start_server (handler):
+    def setUp(self):
+        """Start a new HTTPS server in a new thread."""
+        self.port = start_server(self.handler, https=True)
+        assert self.port is not None
+
+    def tearDown(self):
+        """Send QUIT request to http server."""
+        stop_server(self.port, https=True)
+
+    def get_url(self, filename):
+        """Get HTTP URL for filename."""
+        return u"https://localhost:%d/tests/checker/data/%s" % (self.port, filename)
+
+
+def start_server (handler, https=False):
     """Start an HTTP server thread and return its port number."""
     server_address = ('localhost', 0)
     handler.protocol_version = "HTTP/1.0"
     httpd = StoppableHttpServer(server_address, handler)
+    if https:
+        httpd.socket = ssl.wrap_socket(httpd.socket,
+            keyfile=get_file("https_key.pem"),
+            certfile=get_file("https_cert.pem"), server_side=True)
     port = httpd.server_port
     t = threading.Thread(None, httpd.serve_forever)
     t.start()
     # wait for server to start up
     while True:
         try:
-            conn = HTTPConnection("localhost:%d" % port)
+            if https:
+                conn = HTTPSConnection("localhost:%d" % port,
+                        context=ssl._create_unverified_context())
+            else:
+                conn = HTTPConnection("localhost:%d" % port)
             conn.request("GET", "/")
             conn.getresponse()
             break
@@ -196,9 +224,13 @@ def start_server (handler):
     return port
 
 
-def stop_server (port):
+def stop_server (port, https=False):
     """Stop an HTTP server thread."""
-    conn = HTTPConnection("localhost:%d" % port)
+    if https:
+        conn = HTTPSConnection("localhost:%d" % port,
+                               context=ssl._create_unverified_context())
+    else:
+        conn = HTTPConnection("localhost:%d" % port)
     conn.request("QUIT", "/")
     conn.getresponse()
 
--- a/tests/checker/test_https.py
+++ b/tests/checker/test_https.py
@@ -15,34 +15,51 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """
-Test news checking.
+Test https.
 """
-import pytest
+from OpenSSL import crypto
 
 from tests import need_network
-from . import LinkCheckTest
+from .httpserver import HttpsServerTest, CookieRedirectHttpRequestHandler
+from .. import get_file
 
 
-class TestHttps (LinkCheckTest):
+class TestHttps(HttpsServerTest):
     """
     Test https: link checking.
     """
 
-    @pytest.mark.xfail
+    def __init__(self, methodName='runTest'):
+        super(TestHttps, self).__init__(methodName=methodName)
+        self.handler = CookieRedirectHttpRequestHandler
+
+    @classmethod
+    def setUpClass(cls):
+        key = crypto.PKey()
+        key.generate_key(crypto.TYPE_RSA, 2048)
+        cert = crypto.X509()
+        cert.get_subject().CN = "localhost"
+        cert.set_serial_number(1000)
+        cert.gmtime_adj_notBefore(0)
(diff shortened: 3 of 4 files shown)


tests/checker/test_https.py
```diff
@@ -15,34 +15,50 @@
...
+        super(TestHttps, self).__init__(methodName=methodName)
+        self.handler = CookieRedirectHttpRequestHandler
+        temp_key_file = get_file("key.pem")
+        subprocess.check_output(["openssl", "req", "-x509", "-newkey",
+            "rsa:2048", "-passout", "pass:linkchecker", "-subj", "/DC=org",
+            "-keyout", temp_key_file, "-out", get_file("https_cert.pem")])
+        # Strip pass phrase from the key, else it is requested at server start
+        subprocess.check_output(["openssl", "pkey",
+            "-passin", "pass:linkchecker",
+            "-in", temp_key_file, "-out", get_file("https_key.pem")])
```
mgedmin: 1. I think this belongs in `setUp`, not `__init__`
2. Surely there's a way of generating a self-signed cert without a password
3. Generating a new SSL key/cert for every test seems expensive, can we not do that when the key already exists?  (And maybe set expiration date to something long -- a decade, a thousand years -- to prevent problems from cert reuse.)
4. I wonder if pyopenssl (or cryptogtaphy) have APIs for generating SSL keys/self-signed certs, ti avoid shelling out.

tests/checker/httpserver.py
```diff
@@ -175,19 +177,45 @@ def get_url(self, filename):
...
     port = httpd.server_port
     t = threading.Thread(None, httpd.serve_forever)
     t.start()
     # wait for server to start up
     while True:
         try:
-            conn = HTTPConnection("localhost:%d" % port)
+            if https:
+                conn = HTTPSConnection("localhost:%d" % port,
+                        context=ssl._create_unverified_context())
```
mgedmin: Is that a documented API method?  I worry because it starts with an underscore.

mgedmin: _create_unverified_context() is not mentioned in https://docs.python.org/3/library/ssl.html.  Let's not use it.

We can inline the implementation from the stdlib and eliminate all the constants from default arguments, which will give us

```python
context = ssl.SSLContext()
context.check_hostname = False
context.verify_mode = ssl.CERT_NONE
```

and the docs say that check_hostname is False by default except in one specific circumstance that doesn't apply to us, so this assignment can be omitted.  The default value of verify_mode is not documented.  I'm pretty sure it is CERT_NONE in all current versions of Python, so we might simplify this to

```python
                conn = HTTPSConnection(..., context=ssl.SSLContext())
```

tests/checker/httpserver.py
```diff
@@ -196,9 +224,13 @@ def start_server (handler):
     return port
 
 
-def stop_server (port):
+def stop_server (port, https=False):
     """Stop an HTTP server thread."""
-    conn = HTTPConnection("localhost:%d" % port)
+    if https:
+        conn = HTTPSConnection("localhost:%d" % port,
+                               context=ssl._create_unverified_context())
```
mgedmin: (the above applies here also)

tests/checker/test_https.py
mgedmin: 2. I believe the `-nodes` argument to `openssl req` suppresses the password prompt (it means "don't encrypt the key" because openssl's command-line UX is terrible).

dev-requirements.txt
```diff
@@ -3,6 +3,7 @@ miniboa
 parameterized
 pdfminer
 pyftpdlib
+pyopenssl
```
mgedmin: Huh, I see we're already using pyopenssl (the vendored copy from requests):

```
linkcheck/httputil.py:    from requests.packages.urllib3.contrib.pyopenssl import get_subj_alt_name
```

I wonder if we should change that to `from pyopenssl import get_subj_alt_name`, or if we should continue using requests.packages.urllib3.contrib.pyopenssl for new pyopenssl api calls.

Incidentally, where are they?  I don't see them in this diff.

tests/checker/test_https.py
mgedmin: Re 4: https://skippylovesmalorie.wordpress.com/2010/02/12/how-to-generate-a-self-signed-certificate-using-pyopenssl/

dev-requirements.txt
anarcat: > Basically ignore me, I've just shown how unqualified I'm to review pull requests :)

I think that's amazingly inaccurate and probably the only relevant inaccuracy in your reviews. :p You rock, don't say stuff like that. ;)

tests/checker/test_https.py
anarcat: > 3. Generating a new SSL key/cert for every test seems expensive, can we not do that when the key already exists? (And maybe set expiration date to something long -- a decade, a thousand years -- to prevent problems from cert reuse.)

There should be a way to generate SSL keys with garbage entropy or `/dev/urandom` so that it's fast. There's such a flag in GnuPG for sure, so maybe there's something similar in openssl?

Seems like a reasonable request anyways.

cjmayo: >     1. I think this belongs in `setUp`, not `__init__`

Is `setUp` called for each method, the docs aren't really clear? I did think about `setUpClass`, saw New in 3.2 and didn't use it - however it is in 2.7 and seems to work.

cjmayo: > Re 4: https://skippylovesmalorie.wordpress.com/2010/02/12/how-to-generate-a-self-signed-certificate-using-pyopenssl/

I will have a go. That link suggests you can set the notAfter time which could make for a better test than just True.

tests/checker/httpserver.py
cjmayo: > _create_unverified_context() is not mentioned in https://docs.python.org/3/library/ssl.html. Let's not use it.

It came from the http.client.HTTPSConnection documentation:
https://docs.python.org/3/library/http.client.html?highlight=_create_unverified_context#http.client.HTTPSConnection

but I don't mind replacing it.

tests/checker/test_https.py
```diff
@@ -15,34 +15,50 @@
...
     """
     Test https: link checking.
     """
 
-    @pytest.mark.xfail
+    def __init__(self, methodName='runTest'):
+        super(TestHttps, self).__init__(methodName=methodName)
+        self.handler = CookieRedirectHttpRequestHandler
+        temp_key_file = get_file("key.pem")
+        subprocess.check_output(["openssl", "req", "-x509", "-newkey",
```
cjmayo: That should have been `subprocess.check_call` rather than `subprocess.check_output`.

tests/checker/test_https.py
mgedmin: Yes, setUp is called before each test method.

mgedmin: Note that `__init__` is called once for every test method too, while constructing the test suite!

The unittest API requires that each instance of a test case has a runTest() method that runs the test.  When you define a test class, the default test loader invokes unittest.MakeSuite(YourTestClass) which creates an empty TestSuite(), enumerates all the test methods of the class, and then instantiates the class by passing the test method name to `__init__` to get a test case instance that will run that particular method from its runTest().

So if you wanted to avoid creating new SSL keys for every test method, `__init__ `is the wrong way to go about it.

(Caveat: I don't know for sure that pytest's unittest compatibility mode also works the same way, but I think it's very likely.)

tests/checker/httpserver.py
mgedmin: That makes me feel much better, but also makes me want to SCREAM at Python developers who did this.

cjmayo:
OK, now with key and certificate generation by pyOpenSSL which enables a better check for x509_to_dict:

```py
self.assertEqual(httputil.x509_to_dict(cert)["notAfter"],
                 "Jan 02 03:04:05 2119 GMT")
```

mgedmin:
(a reminder that force-pushes force me to review the entire thing again, since I cannot see just the changes since the last time)

tests/checker/test_https.py
```diff
@@ -15,34 +15,59 @@
...
+
+class TestHttps(HttpsServerTest):
     """
     Test https: link checking.
     """
 
-    @pytest.mark.xfail
+    def __init__(self, methodName='runTest'):
+        super(TestHttps, self).__init__(methodName=methodName)
+        self.handler = CookieRedirectHttpRequestHandler
```
mgedmin: Please don't overwrite `__init__`.

You can assign `handler = CookieRediectHttpRequestHandler` as a class attribute.

tests/checker/test_https.py
```diff
@@ -15,34 +15,59 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        key = crypto.PKey()
+        key.generate_key(crypto.TYPE_RSA, 2048)
+        cert = crypto.X509()
+        cert.get_subject().CN = "localhost"
+        cert.set_serial_number(1000)
+        cert.gmtime_adj_notBefore(0)
+        cert.set_notAfter(b"21190102030405Z")
```
mgedmin: One hundred years, haha!

tests/checker/test_https.py
cjmayo: I just copied tests/checker/test_http.py. test_http_redirect.py and test_https_redirect.py also overwrite `__init__`. 
The parent class, HttpsServerTest, does use `self.handler`, with the `__init__` being in its parent, HttpServerTest. So, there is a fair amount to rewrite here that is possibly best done in a new PR. Would you have a class attribute in the parent and then override it in the children?

mgedmin: Ah, ok then. Consistency with existing code trumps general guidelines.

PR #344, opened 2020-02-06
Require beautifulsoup4 instead of bs4

hroncok:
bs4 is a dummy package managed by the developer of Beautiful Soup to prevent
name squatting. The official name of PyPI’s Beautiful Soup Python package is
beautifulsoup4. The bs4 package ensures that if you type pip install bs4 by
mistake you will end up with Beautiful Soup.

However, for requirements, it's cleaner to use the proper name.
For downstream packaging in Fedora, this avoids the need of packaging
the dummy package.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # required:
-bs4
+beautifulsoup4
 requests >= 2.4
 pyxdg
 dnspython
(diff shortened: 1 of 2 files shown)


mgedmin:
Thank you!

PR #351, opened 2020-03-31
Remove support for non-Tag elements from Parser

cjmayo:
This change is made because the linkchecker handlers only process Tags.

The test HtmlPrettyPrinter handler is updated to output element text because its support for non-Tag elements has been removed. This results in a number of the existing tests still passing.

---

We can go further than just removing DOCTYPE support.
I've just deleted parsetests strings that were causing failures (for the most part).
We don't have our own HTML parser to test any more, on the other hand it seems worth having a dedicated parser test (we could just leave it to be inferred from the specialist tests). What do you think?

I'll have a look at what is left of the parser and the interface to it. I think there might be some more tidying possible there.

I couldn't resist a Python 2 tidy up as well, hopefully it is obviously separate from the main changes here. Plenty more of that to do.

### DIFF
--- a/linkcheck/HtmlParser/__init__.py
+++ b/linkcheck/HtmlParser/__init__.py
@@ -26,11 +26,6 @@ in the ListDict (ie. "<a href>" with lead to a {href: None} dict entry).
 
 Used callbacks of a handler are:
 
-- Comments: <!--data-->
-  def comment (data)
-  @param data:
-  @type data: Unicode string
-
 - Start tag: <tag {attr1:value1, attr2:value2, ..}>
   def start_element (tag, attrs)
   @param tag: tag name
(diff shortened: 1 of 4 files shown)


anarcat:
why does coveralls complain about all those PRs?

mgedmin:
Apparently we remove more tested code than untested code, which makes the remaining untested code a slightly larger fraction of the entire remaining code.

It would be nice if coveralls warned about increases in the total number of missed lines, rather than about fractions.  Unfortunately it can't.

Eventually maybe we'll get to 100% coverage and then I'll start paying attention to coveralls complains.

mgedmin:
I've tweaked the coveralls settings so it'll fail a build if the decrease is 1% or more, and, to avoid the slippery slope of inching towards 0, I also asked it to fail the build if the coverage falls below 73%.

PR #354, opened 2020-04-01
Remove use of Python 2 unicode() and related u prefixes

cjmayo:
Several instances for MS Windows left unchanged.

---

The instances in the bookmarks modules can't be part of the existing tests (they haven't been failing under Python 3). They seem simple enough to update, for Linux, not sure about Windows so have left those.

### DIFF
--- a/linkcheck/htmlutil/formsearch.py
+++ b/linkcheck/htmlutil/formsearch.py
@@ -33,16 +33,8 @@ class Form(object):
         self.data[key] = value
 
     def __repr__(self):
-        """Return unicode representation displaying URL and form data."""
-        return unicode(self)
-
-    def __unicode__(self):
-        """Return unicode string displaying URL and form data."""
-        return u"<url=%s data=%s>" % (self.url, self.data)
-
-    def __str__(self):
         """Return string displaying URL and form data."""
-        return unicode(self).encode('utf-8')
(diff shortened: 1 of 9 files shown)


linkcheck/htmlutil/formsearch.py
```diff
@@ -33,16 +33,8 @@ def add_value(self, key, value):
         self.data[key] = value
 
     def __repr__(self):
-        """Return unicode representation displaying URL and form data."""
-        return unicode(self)
-
-    def __unicode__(self):
-        """Return unicode string displaying URL and form data."""
-        return u"<url=%s data=%s>" % (self.url, self.data)
```
anarcat, line 41: seems like we lost the __str__ representation here.

cjmayo: Yes, I was relying on:
`
If a class defines __repr__() but not __str__(), then __repr__() is also used when an “informal” string representation of instances of that class is required.
`

[https://docs.python.org/3/reference/datamodel.html#object.__repr__](https://docs.python.org/3/reference/datamodel.html#object.__repr__)

But I am not clear if there is test using this.

anarcat: oh wait, both representations are identical, nevermind. :)

PR #355, opened 2020-04-02
Remove unecessary parser related code

cjmayo:
This is the first phase of reworking the parser (that probably makes #351 phase zero) and removes code - there are a few equivalent replacements resulting from swapping the use of ListDict with bs4 attrs dict. The next PR will improve the code.


cjmayo:
OK. Forgot that, it will have to be everything at once. I believe these changes are correct though if you want to get an early look.

PR #356, opened 2020-04-02
Remove unecessary parser related code

cjmayo:
This is the first phase of reworking the parser (that probably makes #351 phase zero) and removes code - there are a few equivalent replacements resulting from swapping the use of ListDict with bs4 attrs dict. The next PR will improve the code.

One less commit than last time.


cjmayo:
> One less commit than last time.

Not that simple either. I won't close this one else I can't update it but please don't merge.

mgedmin:
Once we drop Python 3.5 (EOL on September 13, 2020) dict ordering will become deterministic.

We could drop 3.5 sooner if that makes our lifes easier.

PR #357, opened 2020-04-04
Simplify the Parser class

cjmayo:
This is the remaining work on the parser. The main idea is to replace the Parser class and its recursive method with a function that used the find_all method of a BeautifulSoup object.

I've tried to break it down into a set of commits that explains how we get there. Do suggest any improvements.

### DIFF
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -36,13 +36,10 @@ class Parser(object):
 
     def __init__(self, handler):
         self.handler = handler
-        self.reset()
 
     def feed_soup(self, soup):
-        self.soup = soup
-
-    def reset(self):
-        self.soup = None
+        self.parse_contents(soup.contents)
+        self.encoding = soup.original_encoding
 
     def parse_contents(self, contents):
         for content in contents:
@@ -63,11 +60,6 @@ class Parser(object):
                         self.parse_contents(content.contents)
                     if hasattr(self.handler, 'end_element'):
                         self.handler.end_element(content.name)
-
-    def flush(self):
-        if hasattr(self.soup, 'contents'):
-            self.parse_contents(self.soup.contents)
-        self.encoding = self.soup.original_encoding
 
 
 def parser(handler=None):
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -32,14 +32,12 @@ def make_soup(markup, from_encoding=None):
 
 class Parser(object):
     handler = None
-    encoding = None
 
     def __init__(self, handler):
         self.handler = handler
 
     def feed_soup(self, soup):
         self.parse_contents(soup.contents)
-        self.encoding = soup.original_encoding
 
     def parse_contents(self, contents):
         for content in contents:
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -49,22 +49,21 @@ class Parser(object):
     def reset(self):
         self.soup = None
         self.html_doc = None
-        self.tag_lineno = None
-        self.tag_column = None
 
     def parse_contents(self, contents):
         for content in contents:
             if isinstance(content, Tag):
-                self.tag_lineno = content.sourceline
-                self.tag_column = None if content.sourcepos is None \
+                tag_column = None if content.sourcepos is None \
                     else content.sourcepos + 1
                 if content.is_empty_element:
                     self.handler.start_end_element(
                         content.name, content.attrs, content.text.strip(),
+                        content.sourceline, tag_column
                     )
                 else:
                     self.handler.start_element(
                         content.name, content.attrs, content.text.strip(),
+                        content.sourceline, tag_column
                     )
                     if hasattr(content, 'contents'):  # recursion
                         self.parse_contents(content.contents)
@@ -78,12 +77,6 @@ class Parser(object):
         if hasattr(self.soup, 'contents'):
             self.parse_contents(self.soup.contents)
         self.encoding = self.soup.original_encoding
-
-    def lineno(self):
-        return self.tag_lineno
-
-    def column(self):
-        return self.tag_column
 
 
 def parser(handler=None):
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -27,6 +27,10 @@ filterwarnings("ignore",
 from bs4 import BeautifulSoup, Tag
 
 
+def make_soup(markup, from_encoding=None):
+    return BeautifulSoup(markup, "html.parser", from_encoding=from_encoding,
+                         multi_valued_attributes=None)
+
 class Parser(object):
     handler = None
     encoding = None
--- a/linkcheck/HtmlParser/htmlsax.py
+++ b/linkcheck/HtmlParser/htmlsax.py
@@ -17,7 +17,6 @@
 HTML parser implemented using Beautiful Soup and html.parser.
 """
 
-from io import BytesIO, StringIO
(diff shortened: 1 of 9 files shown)


cjmayo:
Handler class variables in bdddb57918 weren't being properly addressed. Maybe still a question about defining tag_filter as a class variable in TagFinder and it's use as an instance variable in MetaRobotsFinder and FormFinder and htmlsax.process_soup()?

cjmayo:
I think tag_filter should always be an instance variable. I've removed the last two commits from this PR. I'll rework them, and now that we only have two functions in the HtmlParser subpackage I'm thinking of moving htmlsax.py to htmlutil and maybe even renaming it - it's only SAXish now?

cjmayo:
Updated commits added back plus a commit renaming and moving the module.

cjmayo:
Tests are passing, but maybe some more thought needed. start/end callbacks don't mean quite the same thing now - especially end. FormFinder is the only handler that has an end callback, needs a bit of investigation to check that, perhaps add a further test.

linkcheck/htmlutil/formsearch.py
```diff
@@ -37,26 +38,25 @@ def __repr__(self):
...
 
     def __init__(self):
         """Initialize local variables."""
         super(FormFinder, self).__init__()
-        # parser object will be initialized when it is used as
-        # a handler object
-        self.parser = None
         self.forms = []
         self.form = None
+        self.tag_filter = FormFinder.TAG_FORM, FormFinder.self.TAG_INPUT
```
mgedmin, line 51: Minor: this would be more readable if you enclosed the tuple literal in parentheses.
```suggestion
        self.tag_filter = (FormFinder.TAG_FORM, FormFinder.self.TAG_INPUT)
```

setup.py
```diff
@@ -382,7 +382,6 @@ def get_file_list (self):
         'linkcheck.configuration',
         'linkcheck.director',
         'linkcheck.htmlutil',
-        'linkcheck.HtmlParser',
```
mgedmin, line 385: (At some point we ought to switch to `setuptools.find_packages()` instead of manually enumerating all the subpackages that exist.)

tests/test_parser.py
```diff
@@ -127,7 +127,7 @@
     ("""<track  src=bla>""", """<track src="bla"/>"""),
     ("""<video  src=bla>""", """<video src="bla"></video>"""),
     # Test inserted tag s
-    ("""<b><a></a></b>""", """<b><a></a></b>"""),
+    ("""<b><a></a></b>""", """<b></b><a></a>"""),
```
mgedmin, line 130: all these malformed html correction changes look like improvements/bugfixes to me!

linkcheck/htmlutil/linkparse.py
```diff
@@ -99,42 +99,44 @@ class StopParse(Exception):
...
     """Class for finding robots.txt meta values in HTML."""
+    TAG_BODY = "body"
+    TAG_META = "meta"
 
     def __init__ (self):
         """Initialize follow and index flags."""
         super(MetaRobotsFinder, self).__init__()
         log.debug(LOG_CHECK, "meta robots finder")
         self.follow = self.index = True
+        self.tag_filter = MetaRobotsFinder.TAG_BODY, MetaRobotsFinder.TAG_META
```
mgedmin, line 130:
```suggestion
        self.tag_filter = (MetaRobotsFinder.TAG_BODY, MetaRobotsFinder.TAG_META)
```

linkcheck/htmlutil/htmlparse.py
```diff
@@ -0,0 +1,96 @@
...
+                element.name, element.attrs, element.text.strip(),
+                element.sourceline, tag_column
+            )
+        else:
+            handler.start_element(
+                element.name, element.attrs, element.text.strip(),
+                element.sourceline, tag_column
+            )
+            if hasattr(handler, 'end_element'):
+                handler.end_element(element.name)
```
mgedmin, line 96: Yeah, this is the questionable part all right.

It feels to me that end_element() ought to be called after all of the element's children have been processed.

Do the FormFinder tests not complain about formless inputs where the inputs aren't actually formless?  I suppose warnings emitted during test runs don't actually make the test fail.

cjmayo:
> I'm withholding approval for now, because of the uncertainty about end_element.
> 

Yes, so the plan is:

1) Add a test for the login form #358.
2) Reset this PR back to the last of the old Parser class improvements and rebase on master when the new test is merged.
3) Rework the use of find_all.
4) Remove Python 2 compatibility code and fix Python 3.8 deprecation errors - I think they overlap because of urllib.

mgedmin:
> Remove Python 2 compatibility code and fix Python 3.8 deprecation errors - I think they overlap because of urllib.

But this will be done in a separate PR, yes, please?

cjmayo:
> But this will be done in a separate PR, yes, please?

Yes at least one PR for each of the steps. I got a bit eager with this one, seeing the end in sight but thinking we were closer than we were. This PR will only be for step 2.

linkcheck/htmlutil/formsearch.py
cjmayo: This is now not included in this PR and will be superseded in the next.

setup.py
cjmayo: Noted.

tests/test_parser.py
cjmayo: So far, so good. The next PR will show a difference.

linkcheck/htmlutil/linkparse.py
cjmayo: The change is now not included in this PR, a fixed version will be in the next.

linkcheck/htmlutil/htmlparse.py
cjmayo: Yes we don't see the log messages. Will be superseded in the next PR.

cjmayo:
This PR has now been reset to only include the changes to the Parser class. It is rebased on master to include the new login form test.

linkcheck/htmlutil/linkparse.py
```diff
@@ -104,18 +104,15 @@ class TagFinder (object):
     def __init__ (self):
         """Initialize local variables."""
         super(TagFinder, self).__init__()
-        # parser object will be initialized when it is used as
-        # a handler object
-        self.parser = None
 
-    def start_element (self, tag, attrs):
```
anarcat, line 111: i don't know how we make note of that, but we should definitely keep in mind that the next release should change major version as per Semver. I know there's probably a billion other changes like this that warrant that, but it's just something i figured i would bring up.

linkchecker 10! :)

linkcheck/HtmlParser/htmlsax.py
```diff
@@ -27,6 +27,10 @@
 from bs4 import BeautifulSoup, Tag
 
 
+def make_soup(markup, from_encoding=None):
```
anarcat, line 29: is it okay to have multiple such objects in parallel? AKA should this be a singleton?

linkcheck/htmlutil/linkparse.py
cjmayo: I was thinking the same. Maybe even some 9.9's (tags only? because `pip install` looks easy  - haven't tried), some early testers would be great.

linkcheck/HtmlParser/htmlsax.py
cjmayo: I think we want multiple objects - they are not just interfaces to a parser but representations of that piece of markup with methods you can use to interrogate.

linkcheck/htmlutil/linkparse.py
anarcat: i would do 10.0rc1 maybe? at least if we're going to do a test release... let's move this to another ticket: #359 :)

PR #358, opened 2020-04-06
Add a test for search_form

cjmayo:
Just as well, this has been broken since 51a06d8a. Also commits to fix that and a problem message character.

### DIFF
--- a/tests/test_formsearch.py
+++ b/tests/test_formsearch.py
@@ -1,0 +1,55 @@
+# Copyright (C) 2020 Chris Mayo
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+Test login form functions.
+"""
+
+import unittest
+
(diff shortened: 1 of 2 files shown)


tests/test_formsearch.py
```diff
@@ -0,0 +1,49 @@
...
+
+login_form = """
+<html>
+<body>
+<h1>Not a Login</h1>
+<form action="https://not.a.login">
+<input name="some_field">
+</form>
+<h1>Login</h1>
+<form action="https://log.me.in">
```
mgedmin, line 32: I would feel slightly better if we used [IANA-reserved](https://www.iana.org/domains/reserved) domain names like https://notlogin.example.com and https://login.example.com for example URLs in tests.

cjmayo: OK, good point. Actually these could just be relative locations - shall we hold off merging until updated?

mgedmin: Either way is fine for me.

(BTW good idea on the relative locations bit!)

cjmayo:
Changed to relative locations, added another input field to the login form and another test for the case when no login fields are found:

```diff
--- a/tests/test_formsearch.py
+++ b/tests/test_formsearch.py
@@ -25,13 +25,14 @@ login_form = """
 <html>
 <body>
 <h1>Not a Login</h1>
-<form action="https://not.a.login">
+<form action="/not_a_login">
 <input name="some_field">
 </form>
 <h1>Login</h1>
-<form action="https://log.me.in">
+<form action="/log_me_in">
 <input name="user_field">
 <input name="password_field">
+<input name="extra_field">
 </form>
 </body>
 </html>
@@ -44,6 +45,11 @@ class TestFormSearch(unittest.TestCase):
         form = formsearch.search_form(login_form,
                                       "user_field", "password_field")
         self.assertIsNotNone(form)
-        self.assertEqual(form.url, "https://log.me.in")
+        self.assertEqual(form.url, "/log_me_in")
         self.assertIn("user_field", form.data)
         self.assertIn("password_field", form.data)
+
+    def test_search_form_none(self):
+        form = formsearch.search_form(login_form,
+                                      "nouser", "nopassword")
+        self.assertIsNone(form)
```

Issue #343, opened 2020-01-21
Unable to check links behind password-protected URL 


this-guy-commits:
## Summary

I am new to linkchecker and may be missing something but I didn't see any other issues related to mine so here goes. I want to check the links behind a password-protected site that has a single password prompt at render. Once I enter my password, the page I want to test loads. 

My understanding of linkchecker is that if I want to check links behind a password-protected site, I am to modify the linkcheckerrc file found at `config/linkcheckerrc/` and include the CGI name for the password field and then provide the actual string I want to pass into the said field. 

## Steps to reproduce

 1. Modify `linkcheckerrc` `loginuserfield` and `loginpasswordfield` fields with my expected values (e.g loginuserfield="password" and loginpasswordfield="super_secret_pass")
2. Run linkchecker via `linkchecker --config ./config/linkcheckerrc <URL to check>`
3. Linkchecker will auth for me and parse password protected sites link's

## Actual result
Linkchecker does not appear to authenticate with the site I expected it to (I confirm this by seeing the presence of links only found as part of the password prompt page) 

## Expected result
Linkchecker would authenticate with the site and check all links behind the password prompt

## Environment
LinkChecker 9.4.0 released xx.xx.xxxx

 * Operating system: iOS Mojave
 * Linkchecker version: 9.4.0
 * Python version: 3.7.4
 * Install method: From source TarBall
 * Site URL: n/a

## Configuration file

[linkcheckerrc.txt](https://github.com/linkchecker/linkchecker/files/4093496/linkcheckerrc.txt)

## Notes

I know the config file is being loaded correctly as I set verbose output to on via `verbose=1` and I no longer have to specify `-v` via the CLI to observe verbose output.

cjmayo:
To try to be clear:
```
loginuserfield="password"
loginpasswordfield="super_secret_pass"
```
should work for a login form like:
```html
<form action="/log_me_in">
<input name="password">
<input name="super_secret_pass">
</form>
```
Is that what you have?

StanKatsyuk:
@cjmayo my page doesn't actually have a username field. The initial landing page has a password field (it's a Dev server). 

Would your example still work?

cjmayo:
Linkchecker version: 9.4.0
Python version: 3.7.4
Install method: From source TarBall

Is that actually Python 2.7? On 21 Jan even master would not have worked with Python 3.

In theory a password only form would use ("password" is the default so doesn't need setting in linkcheckerrc):

```
[...]

PR #360, opened 2020-04-09
Replace Parser class using BeautifulSoup.find_all()

cjmayo:
This goes after #357, as we discovered there to be able to use BeautifulSoup.find_all() first you have to remove the use of the end_element() callback. This is achieved by replacing the FormFinder handler using BeautifulSoup.find_all(). The start_end_element() callback also turns out to be redundant.


tests/htmllib.py
```diff
@@ -49,42 +49,16 @@ def start_element (self, tag, attrs, element_text, lineno, column):
...
             if val is None:
                 self.fd.write(" %s" % key)
             else:
                 self.fd.write(' %s="%s"' % (key, quote_attrval(val)))
-        self.fd.write(end)
         if element_text:
-            self.fd.write(element_text)
+            self.fd.write(">%s</%s>" % (element_text, tag))
+        else:
+            self.fd.write("/>")
```
anarcat, line 61: this seems strange - why do we need to change this behaviour here? not that i object, it makes sense to do <foo/> instead of <foo></foo>, but it seems gratuitous.

linkcheck/htmlutil/formsearch.py
```diff
@@ -37,60 +37,22 @@ def __repr__(self):
...
-    # parse
-    parser.feed_soup(htmlsax.make_soup(content))
-    log.debug(LOG_CHECK, "Found forms %s", handler.forms)
-    cginames = (cgiuser.lower(), cgipassword.lower())
-    for form in handler.forms:
-        for key, value in form.data.items():
-            if key.lower() in cginames:
-                return form
+    soup = htmlsax.make_soup(content)
+    cginames = set([cgiuser, cgipassword])
```
mgedmin, line 45: You could use a set literal

```suggestion
    cginames = {cgiuser, cgipassword}
```

linkcheck/htmlutil/formsearch.py
```diff
@@ -37,60 +37,22 @@ def __repr__(self):
...
+    for form_element in soup.find_all("form", action=True):
+        form = Form(form_element["action"])
+        for input_element in form_element.find_all("input",
+                                                   attrs={"name": True}):
+            form.add_value(
+                input_element["name"], input_element.attrs.get("value"))
+        if cginames <= set(form.data.keys()):
```
mgedmin, line 52: You're not lowercasing form.data.keys(), and I'm pretty sure beautifulsoup won't do that either.

tests/htmllib.py
mgedmin: Makes sense to me: the internal handler protocol changed -- the parser no longer calls end_element or start_end_element, there's only start_element, so now it needs to do the job of all three.

cjmayo: See review comment below:
https://github.com/linkchecker/linkchecker/pull/360#discussion_r406675103

linkcheck/HtmlParser/__init__.py
```diff
@@ -19,56 +19,31 @@
...
 
-First make a HTML SAX handler object. Missing callback functions are
-ignored. The object returned from callbacks is also ignored.
-Note that a missing attribute value is stored as the value None
-in the ListDict (ie. "<a href>" with lead to a {href: None} dict entry).
 
-Used callbacks of a handler are:
+Two functions are provided, one to make a BeautifulSoup object from markup and
+another to call a handler's callback for each element in a BeautifulSoup
+object it can process.
```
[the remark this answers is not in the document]
cjmayo: Yes on all points.

linkcheck/director/aggregator.py
```diff
@@ -88,8 +88,9 @@ def visit_loginurl(self):
         # XXX user-agent header
         # XXX timeout
         response = session.get(url)
-        cgiuser = self.config["loginuserfield"]
-        cgipassword = self.config["loginpasswordfield"]
+        # All attribute names are converted to lower case by the parser
```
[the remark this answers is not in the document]
cjmayo: Removed.

linkcheck/htmlutil/formsearch.py
cjmayo: Changed to a literal.

linkcheck/htmlutil/formsearch.py
cjmayo: Indeed.

cjmayo:
I think I got confused because the Form.data.keys were the name attribute values. Glad you spotted that. Updated the last commit:
```diff
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -88,9 +88,8 @@ class Aggregate (object):
         # XXX user-agent header
         # XXX timeout
         response = session.get(url)
-        # All attribute names are converted to lower case by the parser
-        cgiuser = self.config["loginuserfield"].lower()
-        cgipassword = self.config["loginpasswordfield"].lower()
+        cgiuser = self.config["loginuserfield"]
+        cgipassword = self.config["loginpasswordfield"]
         form = formsearch.search_form(response.text, cgiuser, cgipassword)
         form.data[cgiuser] = user
         form.data[cgipassword] = password
diff --git a/linkcheck/htmlutil/formsearch.py b/linkcheck/htmlutil/formsearch.py
index c7528e70..14c9868f 100644
--- a/linkcheck/htmlutil/formsearch.py
+++ b/linkcheck/htmlutil/formsearch.py
@@ -42,7 +42,7 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsax.make_soup(content)
-    cginames = set([cgiuser, cgipassword])
+    cginames = {cgiuser, cgipassword}
     for form_element in soup.find_all("form", action=True):
         form = Form(form_element["action"])
         for input_element in form_element.find_all("input",
diff --git a/tests/test_formsearch.py b/tests/test_formsearch.py
index 9f61654d..3363570c 100644
--- a/tests/test_formsearch.py
+++ b/tests/test_formsearch.py
@@ -30,8 +30,8 @@ login_form = """
 </form>
 <h1>Login</h1>
 <form action="/log_me_in">
-<input name="user_field">
-<input name="password_field">
+<input name="User_Field">
+<input name="Password_Field">
 <input name="extra_field">
 </form>
 </body>
@@ -43,13 +43,13 @@ class TestFormSearch(unittest.TestCase):
 
     def test_search_form(self):
         form = formsearch.search_form(login_form,
-                                      "user_field", "password_field")
+                                      "User_Field", "Password_Field")
         self.assertIsNotNone(form)
         self.assertEqual(form.url, "/log_me_in")
-        self.assertIn("user_field", form.data)
-        self.assertIn("password_field", form.data)
+        self.assertIn("User_Field", form.data)
+        self.assertIn("Password_Field", form.data)
 
     def test_search_form_none(self):
         form = formsearch.search_form(login_form,
-                                      "nouser", "nopassword")
+                                      "user_field", "password_field")
         self.assertIsNone(form)
```

cjmayo:
Oops, that should have been the first commit not the last one. Please hold.

cjmayo:
That should be OK now.

cjmayo:
Isn't it even more complex because the case-preserved keys are posted to the form's action:
https://github.com/linkchecker/linkchecker/blob/d80a075372dadd5cff2fe772c220464748704f1b/linkcheck/director/aggregator.py#L93-L99

So, if we match case-insensitively but the server with the action is case-sensitive login can fail.
Doesn't that make this PR better than the existing?

I'm not sure what the need for case-insensitivity would be? It is a one-off configuration for a specific page.

cjmayo:
OK, I think name should be insensitive:
https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT

OK to post what was configured rather than what was matched (previous behaviour)?

cjmayo:
Commit updated again, easier for me to understand by testing. The names that are posted are the ones found in the form. We are just matching insensitively.

```diff
--- a/linkcheck/htmlutil/formsearch.py
+++ b/linkcheck/htmlutil/formsearch.py
@@ -42,14 +42,16 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsax.make_soup(content)
-    cginames = {cgiuser, cgipassword}
+    # The value of the name attribute is case-insensitive
+    # https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT
+    cginames = {cgiuser.lower(), cgipassword.lower()}
     for form_element in soup.find_all("form", action=True):
         form = Form(form_element["action"])
         for input_element in form_element.find_all("input",
                                                    attrs={"name": True}):
             form.add_value(
                 input_element["name"], input_element.attrs.get("value"))
-        if cginames <= set(form.data.keys()):
+        if cginames <= set([x.lower() for x in form.data.keys()]):
             log.debug(LOG_CHECK, "Found form %s", form)
             return form
 
diff --git a/tests/test_formsearch.py b/tests/test_formsearch.py
index 3363570c..c4427337 100644
--- a/tests/test_formsearch.py
+++ b/tests/test_formsearch.py
@@ -43,7 +43,7 @@ class TestFormSearch(unittest.TestCase):
 
     def test_search_form(self):
         form = formsearch.search_form(login_form,
-                                      "User_Field", "Password_Field")
+                                      "USER_FIELD", "password_field")
         self.assertIsNotNone(form)
         self.assertEqual(form.url, "/log_me_in")
         self.assertIn("User_Field", form.data)
@@ -51,5 +51,5 @@ class TestFormSearch(unittest.TestCase):
 
     def test_search_form_none(self):
         form = formsearch.search_form(login_form,
-                                      "user_field", "password_field")
+                                      "nouser", "nopassword")
         self.assertIsNone(form)
```

cjmayo:
Ah, but the Form.data dict will be case-sensitive...

linkcheck/htmlutil/formsearch.py
```diff
@@ -37,60 +37,24 @@ def __repr__(self):
...
+    # The value of the name attribute is case-insensitive
+    # https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT
+    cginames = {cgiuser.lower(), cgipassword.lower()}
+    for form_element in soup.find_all("form", action=True):
+        form = Form(form_element["action"])
+        for input_element in form_element.find_all("input",
+                                                   attrs={"name": True}):
+            form.add_value(
+                input_element["name"], input_element.attrs.get("value"))
+        if cginames <= set([x.lower() for x in form.data.keys()]):
```
mgedmin, line 54: This could also be expressed as a set comprehension:
```suggestion
        if cginames <= {x.lower() for x in form.data.keys()}:
```

also, if `form.data` is a dict, the `.keys()` is redundant:

```suggestion
        if cginames <= {x.lower() for x in form.data}:
```

mgedmin:
Welp, you're right, the login code in aggregator.py doesn't support case-insensitive matching.  This means I don't understand why the old search_form() bothered to .lower() anything...

cjmayo:
I haven't added this to the PR, but we could always POST lower case:

```diff
diff --git a/linkcheck/director/aggregator.py b/linkcheck/director/aggregator.py
index 2ba8ef12..a321cd72 100644
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -88,8 +88,10 @@ class Aggregate (object):
         # XXX user-agent header
         # XXX timeout
         response = session.get(url)
-        cgiuser = self.config["loginuserfield"]
-        cgipassword = self.config["loginpasswordfield"]
+        # The value of the name attribute is case-insensitive
+        # https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT
+        cgiuser = self.config["loginuserfield"].lower()
+        cgipassword = self.config["loginpasswordfield"].lower()
         form = formsearch.search_form(response.text, cgiuser, cgipassword)
         form.data[cgiuser] = user
         form.data[cgipassword] = password
diff --git a/linkcheck/htmlutil/formsearch.py b/linkcheck/htmlutil/formsearch.py
index 45ad1a25..dfdffec3 100644
--- a/linkcheck/htmlutil/formsearch.py
+++ b/linkcheck/htmlutil/formsearch.py
@@ -42,16 +42,15 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsax.make_soup(content)
-    # The value of the name attribute is case-insensitive
-    # https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT
-    cginames = {cgiuser.lower(), cgipassword.lower()}
+    cginames = {cgiuser, cgipassword}
     for form_element in soup.find_all("form", action=True):
         form = Form(form_element["action"])
         for input_element in form_element.find_all("input",
                                                    attrs={"name": True}):
             form.add_value(
-                input_element["name"], input_element.attrs.get("value"))
-        if cginames <= set([x.lower() for x in form.data.keys()]):
+                input_element["name"].lower(),
+                input_element.attrs.get("value"))
+        if cginames <= {x.lower() for x in form.data}:
             log.debug(LOG_CHECK, "Found form %s", form)
             return form
 
diff --git a/tests/test_formsearch.py b/tests/test_formsearch.py
index c4427337..7c6e84b2 100644
--- a/tests/test_formsearch.py
+++ b/tests/test_formsearch.py
@@ -43,11 +43,11 @@ class TestFormSearch(unittest.TestCase):
 
     def test_search_form(self):
         form = formsearch.search_form(login_form,
-                                      "USER_FIELD", "password_field")
+                                      "user_field", "password_field")
         self.assertIsNotNone(form)
         self.assertEqual(form.url, "/log_me_in")
-        self.assertIn("User_Field", form.data)
-        self.assertIn("Password_Field", form.data)
+        self.assertIn("user_field", form.data)
+        self.assertIn("password_field", form.data)
 
     def test_search_form_none(self):
         form = formsearch.search_form(login_form,
```

mgedmin:
I don't trust random webservers to handle form field names case-insenitively, no matter what the HTML spec says.

cjmayo:
Better to change the behaviour and look case-sensitively for what the user asked and then only if that matches send the server back exactly what it asked for. Then both parties at least have a chance of debugging, by just looking at the user configuration rather than magic being done in the code.

mgedmin:
> Better to change the behaviour and look case-sensitively for what the user asked and then only if that matches send the server back exactly what it asked for. Then both parties at least have a chance of debugging, by just looking at the user configuration rather than magic being done in the code.

Yes, but in a separate PR please.  This PR refactors existing code while preserving existing behavior of `search_form()`.

cjmayo:
OK, I've only applied the set comprehension improvement:
```diff
--- a/linkcheck/htmlutil/formsearch.py
+++ b/linkcheck/htmlutil/formsearch.py
@@ -51,7 +51,7 @@ def search_form(content, cgiuser, cgipassword):
                                                    attrs={"name": True}):
             form.add_value(
                 input_element["name"], input_element.attrs.get("value"))
-        if cginames <= set([x.lower() for x in form.data.keys()]):
+        if cginames <= {x.lower() for x in form.data}:
             log.debug(LOG_CHECK, "Found form %s", form)
             return form
```
So we have case-insensitive matching while preserving the name values.
There are still a few PRs to come on the parser...

linkcheck/htmlutil/formsearch.py
```diff
@@ -37,60 +37,24 @@ def __repr__(self):
...
+    cginames = {cgiuser.lower(), cgipassword.lower()}
+    for form_element in soup.find_all("form", action=True):
+        form = Form(form_element["action"])
+        for input_element in form_element.find_all("input",
+                                                   attrs={"name": True}):
+            form.add_value(
+                input_element["name"], input_element.attrs.get("value"))
+        if cginames <= {x.lower() for x in form.data}:
+            log.debug(LOG_CHECK, "Found form %s", form)
+            return form
```
mgedmin, line 56: #343 reminded me about one subtlety in this change: the old code would find and return a form that had just one of the two fields (username + password) that we're looking for.  The new code always requires both.

There are some pages out there that are password-protected without a username field (Mailman admin/moderation web UI comes to mind), so maybe we should think about supporting that.

(I gather that while the old linkchecker code might have supported that here, in search_form, it didn't actually support it fully elsewhere, in aggregator.py, where it wanted to submit the form with both fields.  Sort of like the case insensitivity that was half-supported in one place, but not supported in the other place.)

cjmayo: > I gather that while the old linkchecker code might have supported that here, in search_form, it didn't actually support it fully elsewhere

I think that is the key point for now - to make what is there more robust. I've posted #364 with some more updates to the login form handling on the end, including the case sensitivity  and an additional check in aggregator.py.
To support user name and/or password (which would be a new feature) seems to need some careful thought, how do you know what the user intended so that you can raise the appropriate errors? My first guess would be to use whether user name and password have been configured to choose which fields to pass to search_form, but that needs some investigation.

cjmayo: I've added #366 for how this might work (at least in Aggregate and search_form).

cjmayo:
Is it OK to merge this now?

mgedmin:
> Is it OK to merge this now?

Yes!

PR #361, opened 2020-04-10
Rename htmlsax.py to htmlsoup.py and add test_content_allows_robots

cjmayo:
This goes after #360.

The suggested rename of htmlsax.py plus Coveralls has been reporting uncovered lines catching StopParse exceptions (I do like the Coveralls web interface). These lines are unreachable because htmlsoup.process_soup() does not raise this exception - so remove those.

Searching the code the only place that StopParse is used is the MetaRobotsFinder handler, this can be replaced by one line with BeautifulSoup.find() but before that it turns out we don't have any coverage for the code checking for `<meta name="robots" content="nofollow">`, so first add tests for that.


cjmayo:
Doesn't look it is testing MetaRobotsFinder:
https://coveralls.io/builds/29997325/source?filename=linkcheck/htmlutil/linkparse.py#L124

mgedmin:
Please ping me if I forget to review this after #360 gets merged and the diff becomes readable.

anarcat:
#360 is ready for merge, and i'm a bit confused by the diff here, consider this my blanket +1 for renaming files where it makes sense.

cjmayo:
Ready for review.

I did rebase on master, I think the merge commit from the previous PR in the sequence left no choice. Any better ideas do let me know.

cjmayo:
@anarcat could you have a look - the diff is simpler now, happy to answer any questions.

Issue #53, opened 2017-03-28
Windows binary availability?


battlewagon:
Is there a location for the latest version of the Windows binaries?  The latest ones I can find are here: https://wummel.github.io/linkchecker/.  Thanks.

anarcat:
i don't think we've made a new release yet, even on PyPI (#4) so we certainly haven't built binaries for windows either. This is part of the todo now!

bolk:
I'm a novice of python distutils and Windows+Python.
I'm trying to create Windows binaries but I'm stuck with the `python setup.py py2exe` which exits with 
> error: invalid command 'py2exe'
I'm compiling with Python 2.7 with py2exe 0.6.9 installed.
Looking at `setup.py`, I cannot see any use of `py2exe`: shouldn't be imported, used and exposed as a possible command in there?

localhorst:
Any news on this? I'd like to offer my help and build a Windows release but I have never worked with Python before. Can someone give me a starting point?

anarcat:
i don't know how the windows binaries are built. my hope is we rip out any non-python code from linkchecker so we don't *have* to support windows explicitely. it would then be a normal platform where the code can be installed with `pip` like everywhere else.

would that work?

Windows users: have you ever installed Python stuff through pip on windows? Does that generally work?

anarcat:
note that this is blocking at least two other bugs, which I think are due to older OpenSSL releases linkchecker links against (#66 #106).

jonathanwinter123:
I know it's a bit late but yes, I have installed python stuff through pip on Windows. Yes, it generally works. I had one or two projects which didn't work on Windows but I'm not sure if they used non-python code which caused the problem.

s0ftimage:
Can someone please help me a version that runs on WIndows?

mgedmin:
You may want to try git master, which no longer requires a C compiler during pip install.

cjmayo:
My recipe for installing master on Windows 10:

Install Python 3 from the Windows Store
https://docs.microsoft.com/en-us/windows/python/beginners

As mentioned in that page, install Git (accepting the defaults) from:
https://git-scm.com/download/win

Now do: `pip3 install git+https://github.com/linkchecker/linkchecker.git`

Find the folder into which it has been installed with `pip show LinkChecker`

Create a batch file `linkchecker.bat` containing that path name, prefixed with `python ` and `site-packages` replaced with `Scripts\linkchecker %*`

[...]

PR #362, opened 2020-04-11
Update README.rst, install.txt and upgrading.txt for Python 3

cjmayo:
Some suggestions for essential documentation updates. The 10.0.0rc1 tag does not exist at the time of writing.

Will also need updating for PyPI after #4.


README.rst
```diff
@@ -32,9 +32,9 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
 
 .. _doc/install.txt: doc/install.txt
 
-Python 2.7.2 or later is needed. It doesn't work with Python 3 yet, see `#40 <https://github.com/linkchecker/linkchecker/pull/40>`_ for details.
+Python 3.5.0 or later is needed.
 
-The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git@v9.4.0``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
+The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git@v10.0.0rc1``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
```
mgedmin, line 37: Do we have a v10.0.0rc1 tag yet on github?  I don't see one on https://github.com/linkchecker/linkchecker/tags.

Maybe we should use `@master` for now.

README.rst
```diff
@@ -32,9 +32,9 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
...
 
 Windows builds are seriously lagging behind the Linux releases, see `#53 <https://github.com/linkchecker/linkchecker/issues/53>`_ for details. For now, the only two options are to install from source or use `Docker for Windows <https://www.docker.com/docker-windows>`_.
```
mgedmin, line 39: In theory this should not be a problem any more, given we got rid of all C extension modules!

doc/upgrading.txt
```diff
@@ -1,8 +1,18 @@
...
 =========
+Migrating from 9.x to 10.0
+-------------------------
+Python 3.5 or newer is required.
+
+The Python Beautiful Soup package is now required. A C compiler is not needed
+for building.
+
+The loginuserfield and loginpasswordfield entries in the authentication section
+of linkcheckerrc are now case-sensitive.
```
mgedmin, line 11: AFAIU they were always case-sensitive.  They were used in a couple of places in the code, and while one of them handled them case-insensitively, the other one was always case-sensitive, making the overall behavior also case-sensitive.

What might have changed is the exact error you get if you get the case wrong.

cjmayo:
I missed the review comments here until now, I will get back to them but documentation seems harder then code. And they aren't holding up any of the code PRs.

Also note comment from 366:
"linkchecker now supports login forms with only a password or only a username field"

mgedmin:
This branch now has merge conflicts :/

anarcat:
> This branch now has merge conflicts :/

fixed

README.rst
mgedmin: What about this?  Do we want non-working instructions on the front page of the GitHub repo?

anarcat: nope. i guess after merging this we'd need to immediately tag rc1. this is a downside of the github workflow, i guess ;)

mgedmin: If we merge this at the same time we release 10.0.0rc1, there's little point in instructing people how to pip install from git, they can pip install 10.0.0rc1 from pypi just as well, if they use `pip install --pre linkchecker`.

anarcat: how about we instruct people to install from git HEAD there instead? something like:

```suggestion
The version in the pip repository is old. Instead, you can use pip to install the latest from git: ``pip install git+https://github.com/linkchecker/linkchecker.git``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
```

that way we don't have to repeatedly update that darn file...

mgedmin: That's where I was going with my "Maybe we should use @master for now." suggestion earlier, but you're right, `@master` is redundant and not necessary.

README.rst
cjmayo: Let's delete this paragraph then. It implies there will be something different for Windows.

README.rst
```diff
@@ -34,7 +34,7 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
 
 Python 3 or later is needed.
 
-The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git@v9.4.0``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
+The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git@v10.0.0rc1``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
```
cjmayo, line 37: Let's make that `pip3` to avoid confusion (looks like the only correct option on Ubuntu 20.04 and works on Gentoo)

doc/upgrading.txt
cjmayo: OK, so they used to match case-insensitively:
https://github.com/linkchecker/linkchecker/blob/71be9b941b2f2e101d5cbe9113913cca32a4fe27/linkcheck/htmlutil/formsearch.py#L107

But then posted case-sensitively:
https://github.com/linkchecker/linkchecker/blob/71be9b941b2f2e101d5cbe9113913cca32a4fe27/linkcheck/director/aggregator.py#L91-L92

Update the text to:

The loginuserfield and loginpasswordfield entries in the authentication section
of linkcheckerrc now match a login form case-sensitively.

README.rst
cjmayo: Let's get rid of the "pip repository is old" sentence. It is the current release now.

doc/install.txt
```diff
@@ -15,12 +15,10 @@ the local system:
 Note that this needs the Python development and a C compiler installed.
```
cjmayo, line 15: Remove this paragraph about the C compiler.

doc/install.txt
```diff
@@ -15,12 +15,10 @@ the local system:
...
 For example on Debian Linux systems use ``sudo apt-get install python-dev``.
 
-
 Setup for Windows
 -----------------
 Execute ``LinkChecker-x.y.exe`` and follow the instructions.
```
cjmayo, line 20: Replace with:

Nobody is currently offering a "exe" download for Windows
See https://github.com/linkchecker/linkchecker/issues/53

cjmayo:
Triple backticks in README.rst aren't valid and don't work on PyPI. Replaced.

cjmayo:
I couldn't see how to work with the branch as modified through the PR interface so I rebased my copy and resolved the conflict. 
But I haven't changed any of the commits, I've put the updates into a new commit.

anarcat:
i am a bit lost as to what remains to be done here, could someone with a clue resolve the resolved conversations? :)

README.rst
mgedmin: Or it will be if we merge this PR right before releasing 10.0.0.

README.rst
```diff
@@ -32,11 +32,9 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
 
 .. _doc/install.txt: doc/install.txt
 
-Python 3 or later is needed.
+Python 3.5.0 or later is needed.
 
-The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git@v9.4.0``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
-
-Windows builds are seriously lagging behind the Linux releases, see `#53 <https://github.com/linkchecker/linkchecker/issues/53>`_ for details. For now, the only two options are to install from source or use `Docker for Windows <https://www.docker.com/docker-windows>`_.
+You can use pip to install the latest from git: ``pip3 install git+https://github.com/linkchecker/linkchecker.git``. See `#4 <https://github.com/linkchecker/linkchecker/pull/4>`_.
```
mgedmin, line 37: Drop the `See #4`, it's about releasing 9.4.0, which is already out there.  We were offering git installs to overcome the problem in 9.3 where it would fail on startup with any recent requests version due to the busted version check.  Now we're offering git installs to overcome the problem in 9.4 where it will not work Python 3.

README.rst
```diff
@@ -48,12 +46,10 @@ Docker usage
 
 If you do not want to install any additional libraries/dependencies you can use the Docker image.
 
-Example for external web site check:
-```
-docker run --rm -it -u $(id -u):$(id -g) linkchecker/linkchecker --verbose https://google.com
-```
+Example for external web site check::
+
+  docker run --rm -it -u $(id -u):$(id -g) linkchecker/linkchecker --verbose https://google.com
```
mgedmin, line 51: Unrelated to this PR, but I strongly feel that we should be using IANA-approved example URLs.
```suggestion
  docker run --rm -it -u $(id -u):$(id -g) linkchecker/linkchecker --verbose https://example.com
```

doc/install.txt
```diff
@@ -10,16 +10,12 @@ the local system:
...
-For example on Debian Linux systems use ``sudo apt-get install python-dev``.
-
+``sudo pip3 install LinkChecker``
 
 Setup for Windows
 -----------------
-Execute ``LinkChecker-x.y.exe`` and follow the instructions.
-
+Nobody is currently offering an "exe" download for Windows
+See https://github.com/linkchecker/linkchecker/issues/53
```
mgedmin, line 18: We might suggest `pip install` too, now people will think LinkChecker doesn't work on Windows and won't work until that issue is fixed.

Although... maybe I should boot into Windows and see if linkchecker git actually works or not.

doc/install.txt
```diff
@@ -32,30 +28,19 @@ Setup for GNU/Linux
...
-     Depending on your distribution, several development packages
-     might be needed to provide a fully functional C development
-     environment.
-   
-   Note for developers: if you want to regenerate the po/linkchecker.pot
-   template from the source files, you will need xgettext with Python
-   support. This is available in gettext >= 0.12.
-
-2. Python >= 2.7.2 from http://www.python.org/
+1. Python >= 3.5.0 from http://www.python.org/
```
mgedmin, line 35:
```suggestion
1. Python >= 3.5.0 from https://www.python.org/
```

doc/install.txt
```diff
@@ -32,30 +28,19 @@ Setup for GNU/Linux
...
    
    Be sure to also have installed the included distutils module.
    On most distributions, the distutils module is included in
    an extra ``python-dev`` package.
 
-3. Python requests module from https://pypi.python.org/pypi/requests
+2. Python requests package from https://pypi.python.org/pypi/requests
```
mgedmin, line 41: PyPI is now using https://pypyi.org/project/... URLs, but I wonder if we want to encourage manual installation of dependencies?

doc/upgrading.txt
```diff
@@ -1,8 +1,18 @@
...
+The loginuserfield and loginpasswordfield entries in the authentication section
+of linkcheckerrc are now matched to a login form case-sensitively.
+
 Migrating from 8.x to 9.0
 -------------------------
-The Python requests module is now required.
+The Python requests package is now required.
 
 Several checks have been moved to plugins (see below).
 Plugins have to be enabled in the configuration file.
```
mgedmin, line 18: A couple of lines down from this I see this text which is news to me:

> The following commandline and configuration options have been deprecated
> and do not have any effect:

@cjmayo, I remember you had a PR removing those options?  I objected because I thought t might cause pain to users if we remove working functionality, and that declaring it deprecated was not a good enough reason to do that.  I didn't realize it wasn't actually **working** functionality.  I retract my objection.

cjmayo: New PR with just the options commit:  #413.

README.rst
cjmayo: We've already got `pip3 install LinkChecker` in install.txt...

That starts to make me think we don't want "install from git" the only instruction in the README. And that this should move as an alternative to install.txt.

README.rst
cjmayo: Done.

README.rst
cjmayo: Done and upgraded the example.com above to https.

doc/install.txt
cjmayo: Done.

doc/install.txt
cjmayo: I read it that "Setup with pip" was universal and then there were specific suggestions if that didn't work for you. Perhaps that can be emphasised e.g. with headings.

doc/install.txt
cjmayo: Updated the URL. Manual is the third option. Where else do we document the optionals? Indeed pip isn't going to help with those

doc/install.txt
mgedmin: People skim.  If I were a Windows user, and I saw a section about installing on Windows, I wouldn't even look at the other text.

doc/install.txt
mgedmin: I can't help noticing "the URL", singular.  There were multiple pypi.python.org URLs in the text; I didn't highlight them all.

cjmayo: I have updated them, not posted yet, and converted GeoIP and mod_wsgi to PyPI. And more htttps.

doc/install.txt
cjmayo: Perhaps the reality for the foreseeable future is that we are a source only distribution that strongly recommends pip for installation and should just get rid of the mentions of binaries in the install.txt and just have some pointer for how to use pip in different environments?

README.rst
cjmayo: > That starts to make me think we don't want "install from git" the only instruction in the README. And that this should move as an alternative to install.txt.

Ignore me...

doc/install.txt
cjmayo: Added a commit with pip instructions for Windows and Mac -  have tried both Windows methods, don't have a Mac.

cjmayo:
OK, I've refreshed everything. Now with pip instructions, "sudo pip3" everywhere, and a new commit for the comments in setup.py - should be PyPI project links all over now. Updated README.rst to say Python 3.5 or later (rather than just Python 3 - needed the rebase on master to resolve that update).

README.rst
```diff
@@ -32,28 +32,26 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
 
 .. _doc/install.txt: doc/install.txt
 
-Python 3 or later is needed.
+Python 3.5 or later is needed.
 
-The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``pip install git+https://github.com/linkchecker/linkchecker.git``. See `#359 <https://github.com/linkchecker/linkchecker/issues/359>`_.
+The version in the pip repository is old. Instead, you can use pip to install the latest release from git: ``sudo pip3 install git+https://github.com/linkchecker/linkchecker.git``. See `#359 <https://github.com/linkchecker/linkchecker/issues/359>`_.
```
mgedmin, line 37: I cannot in good conscience recommend `sudo pip3`.

Quoting [The Official Tutorial](https://packaging.python.org/tutorials/installing-packages/):

> Do _not_ run any of the commands in this tutorial with `sudo`: if you get a permissions error, come back to the section on creating virtual environments, set one up, and then continue with the tutorial as written.

README.rst
```diff
@@ -32,28 +32,26 @@ See `doc/install.txt`_ in the source code archive for general information. Excep
...
 
 Windows builds are seriously lagging behind the Linux releases, see `#53 <https://github.com/linkchecker/linkchecker/issues/53>`_ for details. For now, the only two options are to install from source or use `Docker for Windows <https://www.docker.com/docker-windows>`_.
 
 Usage
 ------
-Execute ``linkchecker http://www.example.com``.
+Execute ``linkchecker https://www.example.com``.
```
mgedmin, line 43: :smile: :+1:

doc/install.txt
```diff
@@ -8,71 +8,61 @@ Setup with pip
...
 the local system:
 
-.. _pip: https://pypi.python.org/pypi/pip
+.. _pip: https://pypi.org/project/pip/
 
-``sudo pip install LinkChecker``
-
-Note that this needs the Python development and a C compiler installed.
-For example on Debian Linux systems use ``sudo apt-get install python-dev``.
+``sudo pip3 install LinkChecker``
```
mgedmin, line 13: (and yet I see we used to suggest sudo pip, ouch)

README.rst
cjmayo: We could say ``pip3 install --user`` and link to:
https://packaging.python.org/tutorials/installing-packages/#installing-to-the-user-site

cjmayo: I'll take sudo out for now and merge this, to not hold up this PR any longer for one point. We can always add more later.

anarcat: i would assume `--user` is the default if not `sudo`. also, `pip3` might be a Debianism, so we might want to be careful about that as well - some platforms point `pip` to `python3`, iirc. but that's another discussion i guess...

cjmayo: Gentoo has `pip3`, as does Python from the Microsoft Store. I think we are OK.

PR #364, opened 2020-04-14
Stop using HTML handlers and improve login form error handling

cjmayo:
This goes after #361.

This is the last of the HTML parser review, replacing MetaRobotsFinder with BeautifulSoup.find() means code can be simplified further with the removal of HTML handlers.

There is more work on the login form searching to try and make it more robust with better logging.

### DIFF
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -26,11 +26,12 @@ import warnings
 warnings.simplefilter('ignore', requests.packages.urllib3.exceptions.InsecureRequestWarning)
 
 from io import BytesIO
+import re
 
 from .. import (log, LOG_CHECK, strformat, mimeutil,
     url as urlutil, LinkCheckerError, httputil)
 from . import (internpaturl, proxysupport)
-from ..htmlutil import htmlsoup, linkparse
+from ..htmlutil import htmlsoup
 # import warnings
 from .const import WARN_HTTP_EMPTY_CONTENT
 from requests.sessions import REDIRECT_STATI
@@ -41,6 +42,9 @@ HTTP_SCHEMAS = ('http://', 'https://')
 
 # helper alias
 unicode_safe = strformat.unicode_safe
+
+# match for robots meta element content attribute
+nofollow_re = re.compile(r"\bnofollow\b", re.IGNORECASE)
 
 class HttpUrl (internpaturl.InternPatternUrl, proxysupport.ProxySupport):
     """
@@ -78,15 +82,9 @@ class HttpUrl (internpaturl.InternPatternUrl, proxysupport.ProxySupport):
         """
         if not self.is_html():
             return True
-        # construct handler object
-        handler = linkparse.MetaRobotsFinder()
-        # parse
-        try:
-            htmlsoup.process_soup(handler, self.get_soup())
-        except linkparse.StopParse as msg:
-            log.debug(LOG_CHECK, "Stopped parsing: %s", msg)
-            pass
-        return handler.follow
+
+        soup = self.get_soup()
+        return not soup.find("meta", attrs={"name": "robots", "content": nofollow_re})
 
     def add_size_info (self):
         """Get size of URL content from HTTP header."""
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -98,11 +98,6 @@ def strip_c_comments (text):
     return c_comment_re.sub('', text)
 
 
-class StopParse(Exception):
-    """Raised when parsing should stop."""
-    pass
-
-
 class TagFinder (object):
     """Base class handling HTML start elements.
     TagFinder instances are used as HTML parser handlers."""
@@ -114,26 +109,6 @@ class TagFinder (object):
     def start_element (self, tag, attrs, element_text, lineno, column):
         """Does nothing, override in a subclass."""
         pass
-
-
-class MetaRobotsFinder (TagFinder):
-    """Class for finding robots.txt meta values in HTML."""
-
-    def __init__ (self):
-        """Initialize follow and index flags."""
-        super(MetaRobotsFinder, self).__init__()
-        log.debug(LOG_CHECK, "meta robots finder")
-        self.follow = self.index = True
-
-    def start_element (self, tag, attrs, element_text, lineno, column):
-        """Search for meta robots.txt "nofollow" and "noindex" flags."""
-        if tag == 'meta' and attrs.get('name') == 'robots':
-            val = attrs.get('content', u'').lower().split(u',')
-            self.follow = u'nofollow' not in val
-            self.index = u'noindex' not in val
-            raise StopParse("found <meta name=robots> tag")
-        elif tag == 'body':
-            raise StopParse("found <body> tag")
 
 
 def is_meta_url (attr, attrs):
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -30,17 +30,17 @@ LinkTags = {
     'a':        [u'href'],
     'applet':   [u'archive', u'src'],
     'area':     [u'href'],
-    'audio':    [u'src'], # HTML5
+    'audio':    [u'src'],  # HTML5
     'bgsound':  [u'src'],
     'blockquote': [u'cite'],
     'body':     [u'background'],
-    'button':   [u'formaction'], # HTML5
+    'button':   [u'formaction'],  # HTML5
     'del':      [u'cite'],
     'embed':    [u'pluginspage', u'src'],
     'form':     [u'action'],
     'frame':    [u'src', u'longdesc'],
     'head':     [u'profile'],
-    'html':     [u'manifest'], # HTML5
+    'html':     [u'manifest'],  # HTML5
     'iframe':   [u'src', u'longdesc'],
     'ilayer':   [u'background'],
     'img':      [u'src', u'lowsrc', u'longdesc', u'usemap', u'srcset'],
@@ -53,13 +53,13 @@ LinkTags = {
     'object':   [u'classid', u'data', u'archive', u'usemap', u'codebase'],
     'q':        [u'cite'],
     'script':   [u'src'],
-    'source':   [u'src'], # HTML5
+    'source':   [u'src'],  # HTML5
     'table':    [u'background'],
     'td':       [u'background'],
     'th':       [u'background'],
     'tr':       [u'background'],
-    'track':    [u'src'], # HTML5
-    'video':    [u'src'], # HTML5
+    'track':    [u'src'],  # HTML5
+    'video':    [u'src'],  # HTML5
     'xmp':      [u'href'],
     None:       [u'style', u'itemtype'],
 }
@@ -130,7 +130,7 @@ class LinkFinder:
         # set universal tag attributes using tagname None
         self.universal_attrs = set(tags.get(None, []))
         self.tags = dict()
-        for  tag, attrs in tags.items():
+        for tag, attrs in tags.items():
             self.tags[tag] = set(attrs)
             # add universal tag attributes
             self.tags[tag].update(self.universal_attrs)
@@ -153,7 +153,7 @@ class LinkFinder:
             name = self.get_link_name(tag, attrs, attr, element_text)
             # possible codebase
             base = u''
-            if tag  == 'applet':
+            if tag == 'applet':
                 base = attrs.get('codebase', u'')
             if not base:
                 base = self.base_ref
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -98,19 +98,6 @@ def strip_c_comments (text):
     return c_comment_re.sub('', text)
 
 
-class TagFinder (object):
-    """Base class handling HTML start elements.
-    TagFinder instances are used as HTML parser handlers."""
-
-    def __init__ (self):
-        """Initialize local variables."""
-        super(TagFinder, self).__init__()
-
-    def start_element (self, tag, attrs, element_text, lineno, column):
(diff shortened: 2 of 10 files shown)


tests/test_formsearch.py
```diff
@@ -30,8 +30,8 @@
 </form>
 <h1>Login</h1>
 <form action="/log_me_in">
-<input name="user_field">
-<input name="password_field">
+<input name="User_Field">
+<input name="Password_Field">
```
anarcat, line 34: this worries me. why do we have to change the test corpus to accomodate this change?

linkcheck/checker/httpurl.py
```diff
@@ -78,15 +78,11 @@ def content_allows_robots (self):
...
-        # parse
-        try:
-            htmlsoup.process_soup(handler, self.get_soup())
-        except linkparse.StopParse as msg:
-            log.debug(LOG_CHECK, "Stopped parsing: %s", msg)
-            pass
-        return handler.follow
+
+        if self.get_soup().find("meta", attrs={"name": "robots",
+                                               "content": "nofollow"}):
```
anarcat, line 83: interesting. i hadn't noticed that code when i submitted a patch to skip "robots.txt" (9d899d1d) - maybe that should respect the `--no-robots` flag as well? but that's not new to this change so i guess it's just a note to self...

linkcheck/htmlutil/loginformsearch.py
```diff
@@ -42,16 +42,14 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsoup.make_soup(content)
-    # The value of the name attribute is case-insensitive
-    # https://www.w3.org/TR/html401/interact/forms.html#adef-name-INPUT
-    cginames = {cgiuser.lower(), cgipassword.lower()}
```
anarcat: > The keys of the form.data dictionary are case-sensitive

I'm confused. The (removed) comment above explicitely says that "the value of the name attribute is case-insensitive". Has this changed in later HTML standards? If not, why are we changing this behavior here?

I understand we might have had a KeyError inside our code, but it seems like a bad idea to remove those `lower()` call to fix it - it feels backwards?

If there's a KeyError in some cases, maybe we should have a test case to cover for that and confirm it's fixed?

In any case, color me confused by this commit (662de1818f48a49f89589f438342ca51b40f624a).

linkcheck/htmlutil/loginformsearch.py
```diff
@@ -54,5 +54,5 @@ def search_form(content, cgiuser, cgipassword):
             return form
 
     # not found
-    log.debug(LOG_CHECK, "Form with fields %s not found", ",".join(cginames))
+    log.warn(LOG_CHECK, "Form with fields %s not found", ",".join(cginames))
```
anarcat, line 57: isn't `warn` deprecated?

linkcheck/director/aggregator.py
```diff
@@ -87,16 +87,22 @@ def visit_loginurl(self):
...
+        if not form:
+            raise LinkCheckerError("Login form not found at %s" % url)
         form.data[cgiuser] = user
         form.data[cgipassword] = password
         for key, value in self.config["loginextrafields"].items():
             form.data[key] = value
         formurl = parse.urljoin(url, form.url)
+        log.debug(LOG_CHECK, "Posting login data to %s", formurl)
         response = session.post(formurl, data=form.data)
+        response.raise_for_status()
```
anarcat, line 105: does someone catch those exceptions above?

cjmayo:
> phew! that's a big set of changes! :) 

That was brave! I don't mind splitting them further, just say. I wouldn't really have submitted this one early but some of the commits were supporting a discussion elsewhere (more of that in some of the answers below).

tests/test_formsearch.py
cjmayo: There was quite a long discussion from https://github.com/linkchecker/linkchecker/pull/360#pullrequestreview-391371004 to nearly the end of that PR (this test is new and only introduced by #358), which concluded that being case sensitive was the way to go.

mgedmin:
Please ping me if I forget to review this after #361 gets merged and the diff becomes readable.

linkcheck/checker/httpurl.py
cjmayo: And even where the check should go, perhaps the existing one should be in HttpUrl.check_connection() and this one would be in UrlBase.allows_recursion()

linkcheck/htmlutil/loginformsearch.py
cjmayo: As per the discussion linked in the comment above we were confused several times too. I think we got to the right place, minimising possible user confusion.

linkcheck/htmlutil/loginformsearch.py
cjmayo: It's our own:

https://github.com/linkchecker/linkchecker/blob/0f18c9b8f0784dbb0acf2d0867081df52d48bb36/linkcheck/log.py#L98-L105

logging.WARN doesn't generate deprecation warnings, which is just as well because we would have quite a bit to change, not just a simple search and replace

linkcheck/director/aggregator.py
cjmayo: Yes (and reraises):

https://github.com/linkchecker/linkchecker/blob/0f18c9b8f0784dbb0acf2d0867081df52d48bb36/linkcheck/director/__init__.py#L36-L41

It is an attempt reduce user confusion again, to give the best information about why the login failed.

cjmayo:
Rebased on master with no changes - just to reduce the diff. Hopefully that is manageable know, please complete the review.

linkcheck/checker/httpurl.py
anarcat: should we open a separate issue for this?

cjmayo: Good idea. Created https://github.com/linkchecker/linkchecker/issues/374.

linkcheck/parser/__init__.py
```diff
@@ -116,8 +116,7 @@ def find_links (url_data, callback, tags):
     """Parse into content and search for URLs to check.
     Found URLs are added to the URL queue.
     """
-    handler = linkparse.LinkFinder(callback, tags)
-    htmlsoup.process_soup(handler, url_data.get_soup())
+    linkparse.LinkFinder(callback, tags, url_data.get_soup())
```
mgedmin, line 119: As a rule I don't like classes that do work in `__init__`.

This here looks like an instantiation of a class that then gets immediately discarded without being used.  You have to know that `__init__` has externally-visible side effects.

Splitting `__init__`  into methods so this would look like 

```suggestion
    handler = linkparse.LinkFinder(callback, tags)
    handler.process_soup(url_data.get_soup())
```

might look better?

(This is a style question, which is very subjective and a matter of taste.  Feel free to disagree!)

tests/test_parser.py
```diff
@@ -142,8 +142,7 @@ class TestParser (unittest.TestCase):
     def test_parse (self, _in, _out):
         # Parse all test patterns in one go.
         out = StringIO()
-        handler = HtmlPrettyPrinter(out)
-        htmlsoup.process_soup(handler, htmlsoup.make_soup(_in))
+        HtmlPrettyPrinter(out, htmlsoup.make_soup(_in))
```
mgedmin, line 145: Here too: we have a class that seemingly does nothing.  It would be fine if it were a function `pretty_print_htm(out, htmlsoup.make_soup(_in))`, but make the name upper-case and it looks wrong.

linkcheck/checker/httpurl.py
```diff
@@ -78,15 +78,11 @@ def content_allows_robots (self):
...
+            return False
+        return True
```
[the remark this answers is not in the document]
mgedmin: Our test suite has a single file with

```
<meta name="robots" content="nofollow">
```

so I think we need to add more tests that have variations (`content="noindex, nofollow, nosnippet"` etc), including bad ones (`<meta name="robots">` without `content=` etc).

Ideally I wouldn't want each of those tests to require an entire test file -- talk about overhead!  How about simple unit tests that overwrite some attributes of a local url_data variable store this or that bit of HTML instead of having it go and fetch it from the network?

linkcheck/htmlutil/linkparse.py
```diff
@@ -92,44 +92,6 @@ def strip_c_comments (text):
...
-        """Initialize follow and index flags."""
-        super(MetaRobotsFinder, self).__init__()
-        log.debug(LOG_CHECK, "meta robots finder")
-        self.follow = self.index = True
-
-    def start_element (self, tag, attrs, element_text, lineno, column):
-        """Search for meta robots.txt "nofollow" and "noindex" flags."""
-        if tag == 'meta' and attrs.get('name') == 'robots':
-            val = attrs.get('content', u'').lower().split(u',')
-            self.follow = u'nofollow' not in val
```
mgedmin, line 126: I want to remark that the old code did handle substring matches a la `<meta name="robots" content="...,nofollow,...">`, but it could be tricked by adding spaces after commas!

According to https://moz.com/learn/seo/robots-meta-directives, spaces after commas are allowed.

The `.lower()` is surprising to me, I would't have realized this directive is case-insensitive. More unit tests needed to cover all the cases!

cjmayo:
> This could've been two PRs, one for meta robots, the other for login forms.
> 
Yes, it got posted early when the login forms commits wouldn't merge into master, They will now so I've created  #375 for them - I think there will be a fair bit of change on this one which will then be easier to follow.

linkcheck/checker/httpurl.py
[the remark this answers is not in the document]
cjmayo:
```py
        soup = self.get_soup()
        return not soup.find(..
```
That's better, it did look ugly.

```py
return 'nofollow' not in meta_robots.get('content', '')
```
.find() takes a regexp so I have gone with:
```py
re.compile(r"\bnofollow\b", re.IGNORECASE)
```

> a local url_data variable store this or that bit of HTML instead of having it go and fetch it from the network?

I think the complication is that it is a method of a HttpUrl object (grandchild of UrlBase) which takes a bit of initialising. I've made the example in the file more complex for now,  could add another commit later.

linkcheck/parser/__init__.py
cjmayo: I do agree really, equally I don't like creating an object and only making one method call.
But it was worth a second look - I think the solution is to move find_links() out of `parser/__init__.py` and create a new version of it in `htmlutil/linkparse.py`, which is a kind of combination of find_links() and process_soup(). This feels right in that now there are only parse_* functions in `parser/__init__.py` and it saves an import in `plugins/anchorcheck.py`.

tests/test_parser.py
cjmayo: Yes, it only needs to be a function now and not a class - happy to take naming suggestions like that!

linkcheck/htmlutil/linkparse.py
cjmayo: As above.

cjmayo:
This is a real v2, all apart from the straightforward "Fix space style..." and "Remove encoding and default fd from HtmlPrettyPrinter" have changed. I've added a separate commit for HtmlPrettyPrinter replacement. Please go again, hopefully the comments above help as a pointer to the changes.

linkcheck/checker/httpurl.py
```diff
@@ -78,15 +82,9 @@ def content_allows_robots (self):
...
-        # parse
-        try:
-            htmlsoup.process_soup(handler, self.get_soup())
-        except linkparse.StopParse as msg:
-            log.debug(LOG_CHECK, "Stopped parsing: %s", msg)
-            pass
-        return handler.follow
+
+        soup = self.get_soup()
+        return not soup.find("meta", attrs={"name": "robots"}, content=nofollow_re)
```
mgedmin, line 87: I'm a bit thrown by the asymmetry between matching the `name` attribute and matching the `content` attribute.

Would either of these two alternatives work?

```suggestion
        return not soup.find("meta", attrs={"name": "robots", "content": nofollow_re})
```

```suggestion
        return not soup.find("meta", name="robots", content=nofollow_re)
```

?

linkcheck/htmlutil/linkparse.py
```diff
@@ -251,3 +212,14 @@ def found_url(self, url, name, base, lineno, column):
         """Add newly found URL to queue."""
         assert isinstance(url, str_text) or url is None, repr(url)
         self.callback(url, line=lineno, column=column, name=name, base=base)
+
+def find_links(soup, callback, tags):
```
mgedmin, line 216: PEP-8 asks for two blank lines between top level classes and functions.

```suggestion

def find_links(soup, callback, tags):
```

linkcheck/checker/httpurl.py
cjmayo: name is reserved by Beautiful Soup for its own use. Bit of making a point there...

linkcheck/htmlutil/linkparse.py
cjmayo: Oops. Concentrated on `parser/__init__.py` and not that one. I will fix.

linkcheck/checker/httpurl.py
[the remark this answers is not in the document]
cjmayo: Can copy the object creation at least from e.g. test_internpat.py.

linkcheck/checker/httpurl.py
mgedmin: what about the other one?

cjmayo: Yes, that works too. I've changed it to use it instead.

cjmayo:
Added additional nofollow tests in a new commit, fixed the line spacing in `htmlutil/linkparse.py` (and `tests/checker/test_content_allows_robots.py`), changed the soup.find() attrs as above.

PR #366, opened 2020-04-15
Support login forms with user and/or password

cjmayo:
This is based on #364 only the last commit is new.

An idea of how login forms with usernames or passwords only could be supported. But - I think more understanding of how the related code works and tests required.

### DIFF
--- a/linkcheck/htmlutil/loginformsearch.py
+++ b/linkcheck/htmlutil/loginformsearch.py
@@ -41,7 +41,7 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsoup.make_soup(content)
-    cginames = {cgiuser, cgipassword}
+    cginames = {cgiuser, cgipassword} - {None}
     for form_element in soup.find_all("form", action=True):
         form = Form(form_element["action"])
         for input_element in form_element.find_all("input",
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -83,19 +83,24 @@ class Aggregate:
         if not url:
             return
         user, password = self.config.get_user_password(url)
+        if not user and not password:
+            raise LinkCheckerError(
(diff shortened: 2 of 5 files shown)


cjmayo:
Now with tests, looks good. To be considered after the parser updates are complete.

cjmayo:
OK, all supporting PRs are merged, rebased on master and ready for review.

linkcheck/htmlutil/loginformsearch.py
```diff
@@ -41,7 +41,7 @@ def search_form(content, cgiuser, cgipassword):
     CGI fields. If no form is found return None.
     """
     soup = htmlsoup.make_soup(content)
-    cginames = {cgiuser, cgipassword}
+    cginames = {x for x in (cgiuser, cgipassword) if x is not None}
```
mgedmin, line 44: It's a set, so we can have fun with set operations!
```suggestion
    cginames = {cgiuser, cgipassword} - {None}
```

I'm not sure that reads better, but it's fun!

mgedmin: When `cgiuser is cgipassword is None`, this function will return the first found form (because the empty set is a subset of any other set).  That is not _wrong_ per se, but is it useful?

mgedmin: Ah, the caller currently ensures either `cgiuser` or `cgipassword` will not be None.

cjmayo: > I'm not sure that reads better, but it's fun!
It is a bit of a mind-bender but I do like it, done.

cjmayo: > Ah, the caller currently ensures either `cgiuser` or `cgipassword` will not be None.

I did think about putting more checks in here, but as search_form() was only used in one place and there are those checks in visit_loginurl() and more in configuration/__init__.py I thought it was just adding adding overhead that would never be useful.

PR #367, opened 2020-04-19
Add test for loginurl

cjmayo:
A new cgi-bin directory is created to identify the scripts to be run by
http.server.CGIHTTPRequestHandler.

---

Adds coverage for director.aggregator.Aggregate.visit_loginurl(). Can be extended to support user or password only testing.

### DIFF
--- a/tests/checker/test_loginurl.py
+++ b/tests/checker/test_loginurl.py
@@ -1,0 +1,46 @@
+# Copyright (C) 2020 Chris Mayo
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+Test director.aggregator.Aggregate.visit_loginurl().
+This includes the retrieval and search of a login form and posting credentials.
+"""
+import re
+
+from .httpserver import HttpServerTest, CGIHandler
+from . import get_test_aggregate
+
+class TestLoginUrl(HttpServerTest):
+    """Test loginurl retrieval, search and posting credentials."""
+
+    def __init__(self, methodName="runTest"):
+        super().__init__(methodName=methodName)
+        self.handler = CGIHandler
+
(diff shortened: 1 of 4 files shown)


tests/checker/test_loginurl.py
```diff
@@ -0,0 +1,46 @@
...
+
+from .httpserver import HttpServerTest, CGIHandler
+from . import get_test_aggregate
+
+class TestLoginUrl(HttpServerTest):
+    """Test loginurl retrieval, search and posting credentials."""
+
+    def __init__(self, methodName="runTest"):
+        super().__init__(methodName=methodName)
+        self.handler = CGIHandler
```
mgedmin, line 30: Now I know this is existing practice in linkchecker's test suite, but I'm mildly uncomfortable overriding `__init__` in test classes.  This could be a class attribute, if the parent class also used class attributes instead of assigning to `self.port` and `self.handler` in an overridden `__init__`.

anarcat: could we just do the **kwargs dance here to carry those around instead?

mgedmin: I'm not sure what you mean, @anarcat

tests/checker/test_loginurl.py
```diff
@@ -0,0 +1,46 @@
...
+import re
+
+from .httpserver import HttpServerTest, CGIHandler
+from . import get_test_aggregate
+
+class TestLoginUrl(HttpServerTest):
+    """Test loginurl retrieval, search and posting credentials."""
+
+    def __init__(self, methodName="runTest"):
+        super().__init__(methodName=methodName)
```
anarcat, line 29: this, i think:

```suggestion
    def __init__(self, *args, methodName="runTest", **kwargs):
        super().__init__(*args, methodName=methodName, **kwargs)
```

tests/checker/test_loginurl.py
anarcat: suggested a change above... maybe i misunderstood your comment?

mgedmin: ah, the "Conversation" tab shows these comments in a different order, and there's nothing above this one.  I see a suggested change below :)

tests/checker/test_loginurl.py
mgedmin: Why?  The parent class (HttpServerTest) doesn't accept anything other than `methodName`.  Neither does unittest.TestCase.

anarcat: this was trying to answer:

> Now I know this is existing practice in linkchecker's test suite, but I'm mildly uncomfortable overriding __init__ in test classes. This could be a class attribute, if the parent class also used class attributes instead of assigning to self.port and self.handler in an overridden __init__.

mgedmin: Yes, but I don't see how adding *args, **kwargs helps?

anarcat: okay, maybe i just misunderstood - i thought you were refering to methodName here...

PR #370, opened 2020-04-21
Skips links marked as known broken

nickpiggott:
Skips anchor links that have a class value of "broken_link".

Should add this as a config option to allow any class name to be excluded from checking.

### DIFF
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -175,6 +175,10 @@ class LinkFinder (TagFinder):
         log.debug(LOG_CHECK, "line %d col %d", lineno, column)
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
+        if tag =="a" and attrs.get_true('class', u''):
+           if ("broken_link" in attrs.get('class')):
+               log.debug(LOG_CHECK," Found a broken_link to %s, not considering further", attrs.get('href'))
+               return
         tagattrs = self.tags.get(tag, self.universal_attrs)
         # parse URLs in tag (possibly multiple URLs in CSS styles)
         for attr in sorted(tagattrs.intersection(attrs)):


linkcheck/htmlutil/linkparse.py
```diff
@@ -175,6 +175,10 @@ def start_element (self, tag, attrs, element_text, lineno, column):
         log.debug(LOG_CHECK, "line %d col %d", lineno, column)
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
+        if tag =="a" and attrs.get_true('class', u''):
```
mgedmin, line 178: This is breaking our unit tests because

> E   AttributeError: 'dict' object has no attribute 'get_true'

(also there's a missing space after the `==`)

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,6 +175,10 @@ def start_element (self, tag, attrs, element_text, lineno, column):
...
+           if ("broken_link" in attrs.get('class')):
```
mgedmin, line 179: I wonder if this should be configurable?  Some sites might use a different class name.

I think we should `.split()` the class attribute so substring matching wouldn't cause accidental matches for unrelated classes like `<a class="unbroken_linkage">` or something.

nickpiggott:
Hi

I'm afraid I'm definitely not a Python pro, so thanks for the guidance.

* Yes, it should be configurable on the command line / config file. I'll add an option
* Yes, it should split(). That was a crude hammer to see if it was working as expected and broadly the right approach

I'll make some changes, add the config option, and document. I'll need to get some help on designing the unit test.

nickpiggott:
I've made some changes, including adding switches / config options for a list of classes to ignore.

I'm not sure the best way to pull that definition of class names into linkparse.py, so that's still statically coded as a variable definition in linkparse.py. Suggestions welcome on how best to pull config["ignoreclass"] in to linkparse.py.

linkcheck/configuration/confparse.py
```diff
@@ -227,6 +227,8 @@ def read_filtering_config (self):
             for line in read_multiline(self.get(section, "nofollow")):
                 pat = get_link_pat(line, strict=0)
                 self.config["externlinks"].append(pat)
+		if self.has_option(section, "ignoreclass"):
+                self.config["ignoreclass"] = [f.strip() for f in self.get(section, 'ignoreclass').split(',')]
```
mgedmin, line 231: The indentation looks very wrong here.

Are you using literal tab characters instead of spaces for indentation, or something?  Please don't.

linkcheck/configuration/__init__.py
```diff
@@ -191,6 +191,7 @@ def __init__ (self):
         self["ignorewarnings"] = []
         self["internlinks"] = []
         self["checkextern"] = False
+		self["ignoreclass"] = []
```
mgedmin: Here too.

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,10 +175,10 @@ def start_element (self, tag, attrs, element_text, lineno, column):
         log.debug(LOG_CHECK, "line %d col %d", lineno, column)
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
-        if tag =="a" and attrs.get_true('class', u''):
-           if ("broken_link" in attrs.get('class')):
-               log.debug(LOG_CHECK," Found a broken_link to %s, not considering further", attrs.get('href'))
-               return
+         if tag == "a" and attrs.get('class'):
```
mgedmin: The indentation looks wrong on this line (it's not matching the previous `if` line).  I would expect a syntax error from the Python parser.

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,10 +175,10 @@ def start_element (self, tag, attrs, element_text, lineno, column):
...
+            if ("broken_link" in attrs.get('class').split()):
```
mgedmin: This indentation change looks correct -- I didn't notice the short 3-space indent in my earlier review.  This change corrects it to the usual 4-space indent recommended by most Python style guides.

linkcheck/configuration/confparse.py
mgedmin: Okay, the new GitHub notifications UI is messed up, and apparently showed me an outdated diff.

(Good job GitHub!  *slow clap*)

Anyway, the indentation is not entirely correct here either: the code inside the `if` should be indented with 4 spaces, not 8 spaces.

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,10 +175,11 @@ def start_element (self, tag, attrs, element_text, lineno, column):
...
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
-        if tag =="a" and attrs.get_true('class', u''):
-           if ("broken_link" in attrs.get('class')):
-               log.debug(LOG_CHECK," Found a broken_link to %s, not considering further", attrs.get('href'))
-               return
+        ignoreclasses = ['broken_link','borked_link']
+        if tag == "a" and attrs.get('class') and ignoreclasses:
+            if any(item in ignoreclasses for item in attrs.get('class').split()):
+                log.debug(LOG_CHECK,"Found link classed \"%s\" to %s, not considering further", attrs.get('class'), attrs.get('href'))
```
mgedmin, line 181: Style tip: space after a comma.

Style tip: instead of backslash-escaping double quotes inside double quotes, use single quotes for the string literal:

```suggestion
                log.debug(LOG_CHECK, 'Found link classed "%s" to %s, not considering further', attrs.get('class'), attrs.get('href'))
```

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,10 +175,11 @@ def start_element (self, tag, attrs, element_text, lineno, column):
         log.debug(LOG_CHECK, "line %d col %d", lineno, column)
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
-        if tag =="a" and attrs.get_true('class', u''):
-           if ("broken_link" in attrs.get('class')):
-               log.debug(LOG_CHECK," Found a broken_link to %s, not considering further", attrs.get('href'))
-               return
+        ignoreclasses = ['broken_link','borked_link']
```
mgedmin, line 178: I suggest passing ignoreclasses (maybe better spelled `ignore_classes` for readability?) to the `__init__` method and stashing them as an attribute on self.

There's only one place (outside test code) where a LinkFinder is created: find_links() in `linkcheck/parser/__init__.py`.  It can access the config via `url_data`, which is an instance of UrlBase.  There's precedent in linkcheck/plugins/sslcertcheck.py:

```python
    config = url_data.aggregate.config
```

so you can then do

```python
    handler = linkparse.LinkFinder(callback, tags, ignore_class=config['ignoreclass'])
```

linkcheck/htmlutil/linkparse.py
```diff
@@ -175,10 +175,10 @@ def start_element (self, tag, attrs, element_text, lineno, column):
...
         if tag == "base" and not self.base_ref:
             self.base_ref = attrs.get("href", u'')
-        ignoreclasses = ['broken_link','borked_link']
-        if tag == "a" and attrs.get('class') and ignoreclasses:
-            if any(item in ignoreclasses for item in attrs.get('class').split()):
-                log.debug(LOG_CHECK,"Found link classed \"%s\" to %s, not considering further", attrs.get('class'), attrs.get('href'))
+        # ignore_classes = ['broken_link', 'borked_link']
+        if tag == "a" and attrs.get('class') and ignore_classes:
+            if any(item in ignore_classes for item in attrs.get('class').split()):
+                log.debug(LOG_CHECK,'Found link classed "%s" to %s, not considering further', attrs.get('class'), attrs.get('href'))
```
mgedmin, line 181: Spaces after commas, please

```suggestion
                log.debug(LOG_CHECK, 'Found link classed "%s" to %s, not considering further', attrs.get('class'), attrs.get('href'))
```

linkchecker
```diff
@@ -331,7 +331,7 @@ group.add_argument("--ignore-url", action="append", metavar="REGEX",
 """Only check syntax of URLs matching the given regular expression.
  This option can be given multiple times."""))
 group.add_argument("--ignore-class", action="append",
-                 dest="ignoreclass", help=_(
+                 dest="ignoreclasses", help=_(
 """Ignore any anchor links styled with this class name"""))
```
mgedmin, line 335: The help text should mention that multiple class names can be specified, separated by commas

linkchecker
```diff
@@ -331,7 +331,7 @@ group.add_argument("--ignore-url", action="append", metavar="REGEX",
 """Only check syntax of URLs matching the given regular expression.
  This option can be given multiple times."""))
 group.add_argument("--ignore-class", action="append",
```
mgedmin, line 333: Maybe 
```suggestion
group.add_argument("--ignore-classes", action="append",
```
?

argparse allows the user to shorten long option names, so --ignore-class will continue to be allowed.

linkcheck/configuration/confparse.py
```diff
@@ -227,8 +227,8 @@ def read_filtering_config (self):
             for line in read_multiline(self.get(section, "nofollow")):
                 pat = get_link_pat(line, strict=0)
                 self.config["externlinks"].append(pat)
-        if self.has_option(section, "ignoreclass"):
-                self.config["ignoreclass"] = [f.strip() for f in self.get(section, 'ignoreclass').split(',')]
+        if self.has_option(section, "ignoreclasses"):
+                self.config["ignoreclasses"] = [f.strip() for f in self.get(section, 'ignoreclasses').split(',')]
```
mgedmin, line 231: Indentation fix:

```suggestion
        if self.has_option(section, "ignoreclasses"):
            self.config["ignoreclasses"] = [f.strip() for f in self.get(section, 'ignoreclasses').split(',')]
```

nickpiggott:
I think this latest set of changes fixes up everything. the --ignore-classes switch takes a comma separated list of classes, which are excluded if any of them are found on the anchor link.

I'll work out how to create and run some tests. I'm using this on a production box each night, so can keep an eye on it too.

linkcheck/htmlutil/linkparse.py
```diff
@@ -156,10 +156,11 @@ class LinkFinder (TagFinder):
     """Find HTML links, and apply them to the callback function with the
     format (url, lineno, column, name, codebase)."""
 
-    def __init__ (self, callback, tags):
+    def __init__ (self, callback, tags, ignore_classes):
```
mgedmin, line 159: This is a new required positional parameter.  You've updated one call site, but two more remain:

```
tests/test_linkparser.py:        h = linkparse.LinkFinder(self._test_one_url(url), linkparse.LinkTags)
tests/test_linkparser.py:        h = linkparse.LinkFinder(callback, linkparse.LinkTags)
```

One possibility: make this a keyword argument with the default value of `None`, then you won't have to touch unrelated unit tests.

OTOH by looking at the existing tests for LinkFinder you may discover how to add a new test for this feature.  Or maybe an integration test could be better.

mgedmin:
> I'll work out how to create and run some tests. I'm using this on a production box each night, so can keep an eye on it too.

It would be nice to have one integration test. AFAIU these are mostly pairs of files in tests/checker/data/: an HTML file and a .result file.  I would suggest copying an existing test and modifying the HTML to add a class attribute and modifying the .results file to exclude the link that should be omitted.  Next, the test itself that uses the files -- check out tests/checker/test_http.py.  There's a test class, and a test method, and it calls self.file_test() and passes the filename.  So if you e.g. copy http.html to http_ignoreclass.html, you can then add an invocation to self.file_test, and then I think you can pass the ignoreclasses=... config option in the `confargs`.

I'm not sure what stage of config parsing this exists in, and whether you should store a list or a comma-separated string in `confargs`.  I would guess a list.

Running the tests should be simple: `tox` (if you've got it installed) should take care of installing all the dependencies in its local virtualenvs etc.  For rapidly iterating on one particular test I like `tox -e py37 -- tests/checker/test_http.py::TestClass::test_name`.

(I hope I'm not misleading you with thes suggestions: I'm not one of the original developers of linkchecker and my familiarity with its internal workings and the test suite extends only as much as I've needed to figure out a couple of bug fixes in the past.  I'm only doing code reviews because somebody has to, and the original devs are MIA.)

mgedmin:
> It would be nice to have one integration test.

I said one, because zero is bad (maybe all units work correctly but aren't hooked up together right), and more than one is also bad (integration tests are slow, and one bug tends to break many integration tests).

I haven't said how many unit tests I want.  The right number is somewhere between zero and however much is needed to achieve 100% coverage for all the changed lines of code.  (https://pypi.org/project/diff-cover/ is a nice tool for this, we should maybe hook it up.)  It depends on how much time you have and how much you enjoy writing unit tests ;)

linkcheck/configuration/confparse.py
nickpiggott: Think fixed in later commit.

linkcheck/htmlutil/linkparse.py
anarcat: or at least linting... which our newer CI should at least flag...

anarcat:
the CI build fails here now and conflicts need to be resolved, sorry... also pay attention to flake8 warnings in the CI logs: they are not enforced yet, but they are worth fixing.

linkcheck/htmlutil/linkparse.py
nickpiggott: I think this fixed in later commits. (My editor keeps defaulting back to tab =tab, not tab = 4 spaces)

nickpiggott:
Update: I'm working on some tests now, but is turning out to be more work than I anticipated, and needs more code refactoring.

linkcheck/htmlutil/linkparse.py
anarcat: ah sorry for the noise

PR #371, opened 2020-04-24
Switch to mandoc for generating html man pages

cjmayo:
Removes the need for diff files and is a currently maintained project.

Cross references are only supported for mdoc macros but because we only have two pages this can be achieved with sed.

A clean target is added to the Makefile to make development easier.

---

mandoc is the best choice I have found for html man pages, FWIW debiman would seem to agree:
https://github.com/Debian/debiman/#prerequisites.

I'm no expert in `groff an.tmac` so "Update man pages to optimise for both html and man" is probably the commit most to look it. I have tried to find a balance for what works in both in the browser and console.


cjmayo:
The web site pages:

https://linkchecker.github.io/linkchecker/man1/linkchecker.1.html
https://linkchecker.github.io/linkchecker/man5/linkcheckerrc.5.html

Are out of date even for 9.4.0.

anarcat:
> I've always used rst2man from docutils for my projects that have manual pages. (No, I'm not volunteering to rewrite linkchecker's man pages to ReStructuredText, and I'm not demanding that anyone else do so. Unless they want to, of course.)

That would be great. I wonder if we could take the HTML output of mandoc and turn that back into RST using pandoc or something. That could work.

cjmayo:
I am tempted to do the ReST rewrite - but of course it is all or nothing, this PR could do just enough. We'll see.

Looks like the web site uses the now defunct wok static site generator. There's another issue. I was wondering about going Sphinx for everything, which can produce an integrated web site and standalone man pages. e.g. flake8

https://raw.githubusercontent.com/PyCQA/flake8/master/docs/source/manpage.rst

cjmayo:
Sphinx is using docutils to generate man pages, so could be a two stage process.

anarcat:
> I was wondering about going Sphinx for everything, which can produce an integrated web site and standalone man pages.

That would be great: it's what I usually do for larger projects.

PR #377, opened 2020-04-30
Remove u string prefixes

cjmayo:
Thought this one deserved its own PR...


cjmayo:
> Wow! I'm not going to look at the diff, I trust you made no mistakes ;)

Thanks. Changing the diff mode to split makes it not too bad... What it does make it easier to see is that converting to Python 3 meant there are some duplicated tests in test_strformat.py. I'll remove those in another PR.

PR #376, opened 2020-04-30
Remove # -*- coding: lines and inheriting object

cjmayo:
Back to the Python 2 tidying. A few tests didn't get their coding lines removed, recorded in the commit message. One super() can go in ansicolor.py along with the object inheritance.

PR #378, opened 2020-04-30
Make _Logger an abstract base class

mgedmin:
The `__metaclass__` syntax is a Python-2-ism.  It was replaced with

    class _Logger (object, metaclass=abc.ABCMeta):

in Python 3.  And then Python 3.4 introduced abc.ABC which is an empty class that has ABCMeta as the metaclass, making it simpler to define abstract base classes.

### DIFF
--- a/linkcheck/logger/__init__.py
+++ b/linkcheck/logger/__init__.py
@@ -130,7 +130,7 @@ class LogStatistics (object):
         self.internal_errors += 1
 
 
-class _Logger (object):
+class _Logger (abc.ABC):
     """
     Base class for logging of checked urls. It defines the public API
     (see below) and offers basic functionality for all loggers.
@@ -158,7 +158,6 @@ class _Logger (object):
     * log_url(url_data)
         Log a checked URL. Called by log_filter_url if do_print is True.
     """
-    __metaclass__ = abc.ABCMeta
 
     # A lowercase name for this logger, usable for option values
     LoggerName = None


mgedmin:
This will likely conflict with #376.  I'll rebase when that one is merged.

linkcheck/logger/__init__.py
```diff
@@ -130,7 +130,7 @@ def log_internal_error (self):
         self.internal_errors += 1
 
 
-class _Logger (object):
+class _Logger (abc.ABC):
```
cjmayo, line 133: Where do you stand on the space between the class name and the bracket? We have both. I have got another commit removing all of them - I mistakenly thought they caused flake8 E211 errors. As they don't I could junk that commit.

anarcat: > Where do you stand on the space between the class name and the bracket? We have both. I have got another commit removing all of them - I mistakenly thought they caused flake8 E211 errors. As they don't I could junk that commit.

Ah! Whitespace! *FINALLY* something I can have a strong opinion and definite skill on! ;)

Joking aside, this is one thing PEP8 is (ironically) silent on. it doesn't explicitely mandate that, strangely enough. flake8 also does not warn on this, so if we pick a policy on this, we might have trouble applying it in the future.

That said, pep8 does say that we should avoid spaces after *functions*:

https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

... so I would argue a class declaration should follow that model.

It's also what [Black](https://github.com/psf/black) does, so I would remove the space, which looks odd to me.

This, obviously, is a rather trivial matter that we shouldn't bother with too much. ;) The first rule of PEP8 is, of course:

> A Foolish Consistency is the Hobgoblin of Little Minds

mgedmin: I hate these spaces.  I've always thought that PEP-8 forbids them as well.  Sometimes I'm tempted to run [`black`](https://pypi.org/project/black/) on the source tree, but I don't because that might make life harder for any currently outstanding PRs.

As usual, consistency with existing code trumps PEP-8 (until we collectively decide to reformat all the existing code), and so I haven't been changing the style of existing code.

PR #380, opened 2020-05-13
linkchecker and setup.py tidy up

cjmayo:
Back to the list of tidy ups, mainly removing Python 2 code.

There are quite a few so if we did them individually we would be here quite a while. I'm trying to group them into themes of a small number of commits that don't overlap. Let me know if it doesn't work and there are too many (or too few...)

### DIFF
--- a/linkchecker
+++ b/linkchecker
@@ -1,4 +1,4 @@
-#!/usr/bin/python -Ru
+#!/usr/bin/python -u
 # Copyright (C) 2000-2014 Bastian Kleineidam
 #
 # This program is free software; you can redistribute it and/or modify
--- a/linkchecker
+++ b/linkchecker
@@ -578,13 +578,12 @@ else:
 if do_profile:
     if has_profile:
         if os.path.exists(_profile):
-            question = _("""Overwrite profiling file %(file)r?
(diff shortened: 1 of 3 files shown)


linkchecker
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python -Ru
+#!/usr/bin/python -u
```
mgedmin, line 1: Better make this
```suggestion
#!/usr/bin/python3 -u
```

setup.py
```diff
@@ -373,19 +373,7 @@ def get_file_list (self):
...
-        'linkcheck.checker',
-        'linkcheck.configuration',
-        'linkcheck.director',
-        'linkcheck.htmlutil',
-        'linkcheck.logger',
-        'linkcheck.network',
-        'linkcheck.parser',
-        'linkcheck.plugins',
-    ],
+    packages = find_packages(include=["linkcheck", "linkcheck.*"]),
```
mgedmin, line 376: :100:

linkchecker
cjmayo: Thanks, I do forget not everyone has been using Python 3 as python.

setup.py
cjmayo: Someone's bright idea.

PR #382, opened 2020-05-14
Make urllib imports and html.escape Python 3 only

cjmayo:
Only two commits, and the second one overlaps the first, but the overlaps are only html_escape -> html.escape - I think that change is worth it to make it clear we are using the standard function and not our own.

I do recommend the split diff view for these.

### DIFF
--- a/linkcheck/strformat.py
+++ b/linkcheck/strformat.py
@@ -32,11 +32,7 @@ import codecs
 import os
 import math
 import time
-try:
-    import urlparse
-except ImportError:
-    # Python 3
-    from urllib import parse as urlparse
+import urllib.parse
 import locale
 import pydoc
 from . import i18n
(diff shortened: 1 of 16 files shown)


cjmayo:
Replacing parse_host_port() in linkcheck/checker/proxysupport.py slipped in there. I'll just split that out.

linkcheck/strformat.py
```diff
@@ -95,8 +91,8 @@ def is_encoding (text):
 
 
 def url_unicode_split (url):
-    """Like urlparse.urlsplit(), but always returning unicode parts."""
-    return [unicode_safe(s) for s in urlparse.urlsplit(url)]
+    """Like urllib.parse.urlsplit(), but always returning unicode parts."""
+    return [unicode_safe(s) for s in urllib.parse.urlsplit(url)]
```
mgedmin, line 95: Out of scope for this PR, but I wonder if this function is still necessary.  I don't think urllib.parse ever deals with byte strings on Python 3.

cjmayo: #383 created to track this.

PR #384, opened 2020-05-15
Replace deprecated urllib.parse methods

cjmayo:
Finally we can get rid of those deprecation warnings. These are local tidy-ups, there may be the potential for a rewrite of these modules using the new methods but that is a bigger job.

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -664,7 +664,7 @@ class UrlBase:
         """
         if self.userinfo:
             # URL itself has authentication info
-            o = urllib.parse.urlparse(self.userinfo)
+            o = urllib.parse.urlparse("//{}@".format(self.userinfo))
             return (o.username, o.password)
         return self.aggregate.config.get_user_password(self.url)
 
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -383,15 +383,16 @@ class UrlBase:
         Also checks for obfuscated IP addresses.
         """
         # check userinfo@host:port syntax
-        self.userinfo, host = urllib.parse.splituser(self.urlparts[1])
-        port = urlutil.default_ports.get(self.scheme, 0)
-        host, port = urlutil.splitport(host, port=port)
+        o = urllib.parse.urlparse("//%s" % self.urlparts[1])
+        self.userinfo = ":".join([x for x in (o.username, o.password) if x is not None])
+        port = o.port
+        if not port:
+            port = urlutil.default_ports.get(self.scheme, 0)
         if port is None:
             raise LinkCheckerError(_("URL host %(host)r has invalid port") %
-                    {"host": host})
+                    {"host": o.hostname})
         self.port = port
-        # set host lowercase
-        self.host = host.lower()
+        self.host = o.hostname
         if self.scheme in scheme_requires_host:
             if not self.host:
                 raise LinkCheckerError(_("URL has empty hostname"))
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -663,7 +663,8 @@ class UrlBase:
         """
         if self.userinfo:
             # URL itself has authentication info
-            return urllib.parse.splitpasswd(self.userinfo)
+            o = urllib.parse.urlparse(self.userinfo)
+            return (o.username, o.password)
         return self.aggregate.config.get_user_password(self.url)
 
     def add_url (self, url, line=0, column=0, page=0, name="", base=None):
--- a/linkcheck/url.py
+++ b/linkcheck/url.py
@@ -395,10 +395,10 @@ def url_quote (url, encoding):
 
 def document_quote (document):
     """Quote given document."""
-    doc, query = urllib.parse.splitquery(document)
-    doc = urllib.parse.quote(doc, safe='/=,')
-    if query:
-        return "%s?%s" % (doc, query)
+    o = urllib.parse.urlparse(document)
+    doc = urllib.parse.quote(o.path, safe='/=,')
+    if o.query:
+        return "%s?%s" % (doc, o.query)
     return doc
 
 
--- a/linkcheck/url.py
+++ b/linkcheck/url.py
@@ -188,7 +188,9 @@ def url_fix_host (urlparts, encoding):
     if not urlparts[1]:
         urlparts[2] = urllib.parse.unquote(urlparts[2], encoding=encoding)
         return False
-    userpass, netloc = urllib.parse.splituser(urlparts[1])
+    o = urllib.parse.urlparse("//%s" % urlparts[1])
+    userpass = ":".join([x for x in (o.username, o.password) if x is not None])
+    netloc = ":".join([str(x) for x in (o.hostname, o.port) if x is not None])
     if userpass:
         userpass = urllib.parse.unquote(userpass, encoding=encoding)
     netloc, is_idn = idna_encode(urllib.parse.unquote(netloc, encoding=encoding).lower())
--- a/linkcheck/url.py
+++ b/linkcheck/url.py
@@ -446,13 +446,11 @@ def url_split (url):
     hostname is always lowercased.
     Precondition: url is syntactically correct URI (eg has no whitespace)
     """
-    scheme, netloc = urllib.parse.splittype(url)
-    host, document = urllib.parse.splithost(netloc)
-    port = default_ports.get(scheme, 0)
-    if host:
-        host = host.lower()
-        host, port = splitport(host, port=port)
-    return scheme, host, port, document
+    o = urllib.parse.urlparse(url)
+    port = o.port
+    if port is None:
+        port = default_ports.get(o.scheme, 0)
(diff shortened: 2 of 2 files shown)


linkcheck/checker/urlbase.py
```diff
@@ -383,15 +383,16 @@ def build_url_parts (self):
         Also checks for obfuscated IP addresses.
         """
         # check userinfo@host:port syntax
-        self.userinfo, host = urllib.parse.splituser(self.urlparts[1])
-        port = urlutil.default_ports.get(self.scheme, 0)
-        host, port = urlutil.splitport(host, port=port)
+        o = urllib.parse.urlparse("//%s" % self.urlparts[1])
+        self.userinfo = ":".join([x for x in (o.username, o.password) if x is not None])
```
mgedmin, line 387: This seems weird to me, so if username is None and password is not, we produce self.userinfo = password.

And if password is None and username is not, we produce self.userinfo = username.

I had to go read the stdlib sources to figure out that if password is not None, then username is guaranteed to be not None as well.

(I also checked if urlparse does any %-decoding, and no, it doesn't.)

linkcheck/checker/urlbase.py
```diff
@@ -383,15 +383,16 @@ def build_url_parts (self):
...
+        port = o.port
+        if not port:
+            port = urlutil.default_ports.get(self.scheme, 0)
         if port is None:
```
mgedmin, line 391: Can it ever be `None`, given the above check?

TBH I don't see how it could be None given the old code as well.

linkcheck/checker/urlbase.py
```diff
@@ -383,15 +383,16 @@ def build_url_parts (self):
...
             raise LinkCheckerError(_("URL host %(host)r has invalid port") %
-                    {"host": host})
+                    {"host": o.hostname})
         self.port = port
-        # set host lowercase
-        self.host = host.lower()
+        self.host = o.hostname
```
mgedmin, line 395: I had to go read stdlib sources again: urlparse lowercases the hostname.

linkcheck/checker/urlbase.py
```diff
@@ -663,7 +664,8 @@ def get_user_password (self):
         """
         if self.userinfo:
             # URL itself has authentication info
-            return urllib.parse.splitpasswd(self.userinfo)
+            o = urllib.parse.urlparse(self.userinfo)
+            return (o.username, o.password)
```
mgedmin, line 668: This fells wrong.

```
>>> urllib.parse.urlparse('user:passwd')
ParseResult(scheme='user', netloc='', path='passwd', params='', query='', fragment='')
>>> urllib.parse.urlparse('user:passwd').username
>>> urllib.parse.urlparse('user:passwd').password
```
```suggestion
            o = urllib.parse.urlparse('//{}@'.format(self.userinfo))
            return (o.username, o.password)
```

linkcheck/url.py
```diff
@@ -395,10 +397,10 @@ def url_quote (url, encoding):
...
 def document_quote (document):
     """Quote given document."""
-    doc, query = urllib.parse.splitquery(document)
-    doc = urllib.parse.quote(doc, safe='/=,')
-    if query:
-        return "%s?%s" % (doc, query)
+    o = urllib.parse.urlparse(document)
+    doc = urllib.parse.quote(o.path, safe='/=,')
+    if o.query:
+        return "%s?%s" % (doc, o.query)
```
mgedmin, line 403: This seems fine, as long as no part of the `document` contains characters that could be misinterpreted, such as : or @ or //.

I'd have to find all callers of document_quote and see where the `document` arg comes from, to be thorough.  I don't want to...

linkcheck/url.py
```diff
@@ -446,13 +448,11 @@ def url_split (url):
...
-    port = default_ports.get(scheme, 0)
-    if host:
-        host = host.lower()
-        host, port = splitport(host, port=port)
-    return scheme, host, port, document
+    o = urllib.parse.urlparse(url)
+    port = o.port
+    if port is None:
+        port = default_ports.get(o.scheme, 0)
+    return o.scheme, o.hostname, port, o.path
```
mgedmin, line 455: This seems pretty straightforward.

It might raise ValueError where the old splitport code didn't, so that might have consequences?

linkcheck/checker/urlbase.py
cjmayo: Glad you caught that. I've added the change.

linkcheck/url.py
cjmayo: From the get()? I thought that was pretty robust. What would you have to pass to it to cause an exception?

linkcheck/checker/urlbase.py
cjmayo: I can't see how.

linkcheck/checker/urlbase.py
mgedmin: What worries me is that no unit test caught this.

Let's add one?

linkcheck/url.py
mgedmin: No, it's the urllib.parse.urlparse()'s return object's `port` property that calls `int()` without cathing ValueError:

```
>>> urllib.parse.urlparse('//host:port')
ParseResult(scheme='', netloc='host:port', path='', params='', query='', fragment='')

>>> urllib.parse.urlparse('//host:port').port
  File "/usr/lib/python3.8/urllib/parse.py", line 174, in port
    raise ValueError(message) from None
ValueError: Port could not be cast to integer value as 'port'
```

The old splitport() helper used the is_numeric_port() helper which checked if the port was all-numeric before passing it to `int()`, so it would've returned the default value.

I like that we're not silently ignoring errors, but I'm not sure the ValueError will be caught and reported as a problem with the URL, rather than cause a traceback and "oops you found a bug" message.

linkcheck/checker/urlbase.py
anarcat: i agree this could be more readable. i can barely parse it now, and so i will definitely have trouble parsing it later. could we maybe expan that loop or at least explain what we're doing here or why?

what was wrong with splituser anyways?

in any case, this was more readable before and now seems a little weird, so i would like to keep the readability, if possible.

linkcheck/url.py
```diff
@@ -188,7 +188,9 @@ def url_fix_host (urlparts, encoding):
     if not urlparts[1]:
         urlparts[2] = urllib.parse.unquote(urlparts[2], encoding=encoding)
         return False
-    userpass, netloc = urllib.parse.splituser(urlparts[1])
+    o = urllib.parse.urlparse("//%s" % urlparts[1])
+    userpass = ":".join([x for x in (o.username, o.password) if x is not None])
+    netloc = ":".join([str(x) for x in (o.hostname, o.port) if x is not None])
```
anarcat, line 193: and here's that pattern again! would it be worth using a wrapper function to clarify this?

linkcheck/checker/urlbase.py
mgedmin: > what was wrong with splituser anyways?

it emits deprecation warnings now:

https://github.com/python/cpython/blob/3.8/Lib/urllib/parse.py#L1038-L1042

linkcheck/checker/urlbase.py
mgedmin: I should probably insist, but I'm so tired I can't bring myself to care.

linkcheck/checker/urlbase.py
anarcat: hum. well it's still not pretty. :)

linkcheck/checker/urlbase.py
anarcat: don't worry about it then. :)

linkcheck/checker/urlbase.py
mgedmin: yeah, it feels as if we're doing something wrong.

self.urlparts is already the result of urlparse, isn't it?  in which case maybe we have a self.urlparts.username and self.urlparts.password already?

mgedmin: > self.urlparts is already the result of urlparse, isn't it? in which case maybe we have a self.urlparts.username and self.urlparts.password already?

No, but only because we use strformat.url_unicode_split(), which converts the ParseResult object into a plain list.

url_unicode_split is not necessary in a Python 3 only world and we should remove it.

I think I'm going to retract my approval for this PR.  Better things are possible!

cjmayo: Yes. The review comments are making me think it is pretty fragile and the effort would be better put towards reviewing the entire file. Something post 10? The deprecations only came in with Python 3.8 so hopefully will be still supported for a while.

mgedmin: Yep, there's no rush!

mgedmin:
The most important thread in this discussion (IMHO) is a bit hidden in the middle of review comments, so I'll post a link here, to make it easier to pick this up if we decide to postpone this until after the release of version 10.0.

The link: https://github.com/linkchecker/linkchecker/pull/384#discussion_r428763340

cjmayo:
Superseded by #486 and #489 (I'll let that close this PR to keep the review comments visible).

PR #385, opened 2020-05-15
Remove Python 2 imports

cjmayo:
Just a set of Python 2 imports being removed.

### DIFF
--- a/linkcheck/director/__init__.py
+++ b/linkcheck/director/__init__.py
@@ -17,11 +17,8 @@
 Management of checking a queue of links with several threads.
 """
 import os
-try: # Python 3
-    from _thread import error as thread_error
-except ImportError: # Python 2
-    from thread import error as thread_error
 import time
+
 from .. import log, LOG_CHECK, LinkCheckerInterrupt, plugins
 from ..cache import urlqueue, robots_txt, results
 from . import aggregator, console
@@ -54,7 +51,7 @@ def check_urls (aggregate):
         raise
     except KeyboardInterrupt:
         interrupt(aggregate)
-    except thread_error:
+    except RuntimeError:
         log.warn(LOG_CHECK,
              _("Could not start a new thread. Check that the current user" \
                " is allowed to start new threads."))
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -17,10 +17,6 @@
 Aggregate needed object instances for checker threads.
 """
 import threading
-try:  # Python 3
-    import _thread
-except ImportError:
-    import thread as _thread
 
 import requests
 import time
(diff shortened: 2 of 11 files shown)


linkcheck/director/__init__.py
```diff
@@ -54,7 +52,7 @@ def check_urls (aggregate):
         raise
     except KeyboardInterrupt:
         interrupt(aggregate)
-    except thread_error:
+    except _thread.error:
```
mgedmin, line 55: This nicely exposes that we're doing something that we shouldn't.  `_thread` is a private stdlib module in Python 3, so why are we using it directly, instead of using `threading`?

https://docs.python.org/3/library/_thread.html#_thread.error says
> Changed in version 3.3: This is now a synonym of the built-in RuntimeError.

so
```suggestion
    except RuntimeError:
```
and we can drop the import altogether.

linkcheck/director/aggregator.py
```diff
@@ -17,10 +17,7 @@
 Aggregate needed object instances for checker threads.
 """
 import threading
-try:  # Python 3
-    import _thread
-except ImportError:
-    import thread as _thread
+import _thread
```
mgedmin, line 20: This is used only for `_thread.get_ident()`.

I'm pretty sure https://docs.python.org/3/library/threading.html#threading.get_ident is the public alias for it.

> New in version 3.3.

yeah, we can use it.

linkcheck/director/logger.py
```diff
@@ -15,10 +15,8 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Logger for aggregator instances"""
 import threading
-try: # Python 3
-    import _thread
-except ImportError: # Python 2
-    import thread as _thread
+import _thread
```
mgedmin, line 18: This is used for _thread.interrupt_main().  I don't think there's a replacement.  We'll have to live with it.

linkcheck/director/task.py
```diff
@@ -13,10 +13,8 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-try: # Python 3
-    import _thread
-except ImportError: # Python 2
-    import thread as _thread
+import _thread
```
mgedmin, line 16: Another `interrupt_main()`.

TBH this business of catching KeyboardInterrupt in threads and calling _thread.interrupt_main()?  I could've sworn that only the main thread in a Python process ever gets KeyboardInterrupt (because only the main thread handles signals), but https://docs.python.org/3/library/_thread.html claims otherwise:

> Threads interact strangely with interrupts: the KeyboardInterrupt exception will be received by an arbitrary thread. (When the signal module is available, interrupts always go to the main thread.)

Then again, when is the `signal` module not available?  It's even available on Windows!

So I don't know.  I'd like to dig into git history and see if there's anything about adding this code in response to a user's bug report, or if it was added purely theoretically.

IOW out of scope for this PR, let's leave it as is for now.

tests/checker/test_noproxy.py
```diff
@@ -16,11 +16,9 @@
 """
 Test proxy handling.
 """
+from test.support import EnvironmentVarGuard
+
 from . import httpserver
-try:
-    from test.support import EnvironmentVarGuard
-except ImportError:  # Python 3
-    from test.test_support import EnvironmentVarGuard
```
mgedmin, line 23: Hey the comment claims EnvironmentVarGuard lived in test.test_support on Python 3!  The comment was wrong.

BTW regarding [test.support](https://docs.python.org/3/library/test.html#module-test.support):

> Note test.support is not a public module. It is documented here to help Python developers write tests. The API of this module is subject to change without backwards compatibility concerns between releases.

and we're naughty naughty people for using it.

linkcheck/director/__init__.py
cjmayo: That's good. Done.

linkcheck/director/aggregator.py
cjmayo: Also done.

linkcheck/director/task.py
cjmayo: Raised  386.

tests/checker/test_noproxy.py
cjmayo: Yes, fortunately  it is only in the test suite and hopefully it will fail noisily if it changes

PR #390, opened 2020-05-17
Add a test for VirusCheck

cjmayo:
An attempt to write a test to allow #329 to be merged. Let's see what happens (I haven't tested this).

### DIFF
--- a/linkcheck/plugins/viruscheck.py
+++ b/linkcheck/plugins/viruscheck.py
@@ -41,7 +41,7 @@ class VirusCheck(_ContentPlugin):
 
     def check(self, url_data):
         """Try to ask GeoIP database for country info."""
-        data = url_data.get_content()
+        data = url_data.get_raw_content()
         infected, errors = scan(data, self.clamav_conf)
         if infected or errors:
             for msg in infected:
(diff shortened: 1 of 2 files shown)


mgedmin:
Oooh,

> TypeError: a bytes-like object is required, not 'str'

in

>   File "/home/travis/build/linkchecker/linkchecker/linkcheck/plugins/viruscheck.py", line 113, in scan
>     line: self.wsock.sendall(data)

I think you found a genuine bug!

cjmayo:
That failed on master as expected with:
`TypeError: a bytes-like object is required, not 'str'`
https://travis-ci.com/github/linkchecker/linkchecker/jobs/335797524#L519

After adding the commit from 329, it nearly passes. Looks like It only fails because the full file URLs are being reported e.g.:
file:///home/travis/build/linkchecker/linkchecker/tests/checker/data/base1.html

mgedmin:
I think the fix is to modify `VirusCheck.check()` and replace

```python
        data = url_data.get_content()
```
with
```python
        data = url_data.get_raw_content()
```

mgedmin:
Damnation, I just looked at #329 and I have no memory of that place!

cjmayo:
Got there eventually.

PR #391, opened 2020-05-17
Bump version in git to 10.0.0.dev0

mgedmin:
It is confusing to have different versions of the code self-identify
with the same version number.  In my experience it's always best to
increment the version number and add a .dev0 suffix right after making a
release.  When it's time to make a new release, you remove the .dev0,
commit, tag that commit, then make second commit that bumps the version
and adds .dev0 back.

This way only releases identify themselves as "version X.Y.Z" with no
.dev0 suffix and it's immediatelly apparent when you've got a prerelease
installed from git.

### DIFF
--- a/linkcheck/updater.py
+++ b/linkcheck/updater.py
@@ -20,7 +20,7 @@ Function to check for updates.
 import os
 from .configuration import Version as CurrentVersion
 from .url import get_content
-from distutils.version import StrictVersion
+from distutils.version import LooseVersion
 
 # Use the Freecode submit file as source since that file gets updated
 # only when releasing a new version.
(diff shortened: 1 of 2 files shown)


cjmayo:
In principle I think it would be good and make bug reports easier to follow.

mgedmin:
test_updater fails:

```
___________________________ TestUpdater.test_updater ___________________________
tests/test_updater.py:31: in test_updater
    res, value = linkcheck.updater.check_update()
linkcheck/updater.py:49: in check_update
    if is_newer_version(version):
linkcheck/updater.py:75: in is_newer_version
    return StrictVersion(version) > StrictVersion(CurrentVersion)
/opt/python/3.5.6/lib/python3.5/distutils/version.py:40: in __init__
    self.parse(vstring)
/opt/python/3.5.6/lib/python3.5/distutils/version.py:137: in parse
    raise ValueError("invalid version number '%s'" % vstring)
E   ValueError: invalid version number '10.0.0.dev0'
```

test_updater needs to go read some [PEP 440](https://www.python.org/dev/peps/pep-0440/) and learn that 10.0.0.dev0 is too a valid version number!

PR #393, opened 2020-05-17
Assorted test tidy-ups

cjmayo:
One extra bit of testing for sitemap indexes, otherwise all minor tidying.


cjmayo:
Removed the new test for sitemap index, that needs more work.

Issue #401, opened 2020-05-19
add linting to our test suite / CI 


anarcat:
what i learned from #398 is that we are missing some linting steps that were previously somewhat part of the development process from CI and the test suite.

we could, for example, run `flake8` from tox.ini (which would hook it into CI, presumably), and also `twine check` for... er.... whatever that does.

we also discussed just throwing the entire thing into `black`, but that seems like a rather bold move at this stage, i would start slow unless we're totally flake8-broken and it would be easier to let black do the hard work.

mgedmin:
Well, currently flake8 finds these errors:

```
1     E111 indentation is not a multiple of four
1     E114 indentation is not a multiple of four (comment)
2     E122 continuation line missing indentation or outdented
3     E124 closing bracket does not match visual indentation
1     E125 continuation line with same indent as next logical line
21    E127 continuation line over-indented for visual indent
137   E128 continuation line under-indented for visual indent
2     E129 visually indented line with same indent as next logical line
4     E131 continuation line unaligned for hanging indent
1     E201 whitespace after '['
1     E202 whitespace before ']'
4     E203 whitespace before ','
7     E221 multiple spaces before operator
2     E222 multiple spaces after operator
31    E225 missing whitespace around operator
1     E228 missing whitespace around modulo operator
10    E231 missing whitespace after ','
46    E251 unexpected spaces around keyword / parameter equals
40    E261 at least two spaces before inline comment
12    E265 block comment should start with '# '
5     E266 too many leading '#' for block comment
2     E271 multiple spaces after keyword
2     E301 expected 1 blank line, found 0
71    E302 expected 2 blank lines, found 1
1     E303 too many blank lines (3)
6     E305 expected 2 blank lines after class or function definition, found 1
7     E306 expected 1 blank line before a nested definition, found 0
31    E402 module level import not at top of file
191   E501 line too long (110 > 80 characters)
24    E502 the backslash is redundant between brackets
32    E701 multiple statements on one line (colon)
4     E722 do not use bare 'except'
6     E731 do not assign a lambda expression, use a def
6     E741 ambiguous variable name 'l'
11    F401 '.logconf.LOG_ROOT' imported but unused
25    F821 undefined name 'unicode'
182   W191 indentation contains tabs
6     W291 trailing whitespace
1     W293 blank line contains whitespace
5     W391 blank line at end of file
```

[...]

PR #402, opened 2020-05-19
Add a 'tox -e flake8' and a Travis CI job

mgedmin:
See #401.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,13 @@ jobs:
   include:
     - python: "3.5"
       env: TOXENV=oldbs4
+    - name: flake8
+      env: TOXENV=flake8
+      before_script:
+      after_success:
+  allow_failures:
+    - name: flake8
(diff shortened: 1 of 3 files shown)


mgedmin:
> I was hoping we would add this check. I have been using flake8 and have a .flake8 that avoids some false positives. I'll propose that after this gets merged.

Neat!

I put the flake8 config in setup.cfg because I never remember its own config file name (is it .flake8rc? apparently not).  Feel free to move it into .flake8 if you prefer that.

anarcat:
> I put the flake8 config in setup.cfg because I never remember its own config file name (is it .flake8rc? apparently not). Feel free to move it into .flake8 if you prefer that.

By all means, stick with well-known filenames. I'd prefer to have everything in `setup.cfg` (or `pyproject.toml` or whatever new name they end up converging on) than having a billion tiny config files astray in the project root. :)

PR #403, opened 2020-05-19
Remove "from builtins import str as str_text"

cjmayo:
These were used in the conversion to Python 3 to support both Python 2 and 3.

I've been fairly cautious (I hope) in the balance between removing instances of str_text and replacing with str. It may be possible to remove more after closer analysis, they can still be found by searching for isinstance.

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -133,20 +133,20 @@ class UrlBase:
         """
         self.base_ref = base_ref
         if self.base_ref is not None:
-            assert isinstance(self.base_ref, str_text), repr(self.base_ref)
+            assert isinstance(self.base_ref, str), repr(self.base_ref)
         self.base_url = base_url.strip() if base_url else base_url
         if self.base_url is not None:
-            assert isinstance(self.base_url, str_text), repr(self.base_url)
+            assert isinstance(self.base_url, str), repr(self.base_url)
         self.parent_url = parent_url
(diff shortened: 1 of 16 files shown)


linkcheck/logger/customxml.py
```diff
@@ -48,20 +47,20 @@ def log_url(self, url_data):
         """
         self.xml_starttag('urldata')
         if self.has_part('url'):
-            self.xml_tag("url", str_text(url_data.base_url))
+            self.xml_tag("url", url_data.base_url)
```
mgedmin, line 50: These are strings, not URL objects with a `__str__`, right?

linkcheck/checker/urlbase.py
```diff
@@ -224,9 +223,7 @@ def set_result(self, msg, valid=True, overwrite=False):
               "Double result %r (previous %r) for %s", msg, self.result, self)
         else:
             self.has_result = True
-        if not isinstance(msg, str_text):
-            log.warn(LOG_CHECK, "Non-unicode result for %s: %r", self, msg)
-        elif not msg:
+        if not msg:
```
mgedmin, line 226: This part was not obvious to me.  Did a git grep for set_result(), saw that all values passed to it are strings.

linkcheck/logger/customxml.py
cjmayo: The are strings.

PR #407, opened 2020-05-21
Update setup.cfg flake8 section

cjmayo:
- Prevent "F821 undefined name" being caused by gettext prefixes
- Include the main executable when running flake8 from the top directory
  without specifying files

---

Minimal .flake8 to support running flake8 on its own when developing, and always avoid the F821s from gettext. Leaving specifying the objects to be checked by tox means the stand-alone use behaves as the developer expects. Add checking the main linkchecker executable from tox.

### DIFF
--- a/linkcheck/parser/__init__.py
+++ b/linkcheck/parser/__init__.py
@@ -131,4 +131,4 @@ def parse_itms_services(url_data):
             break
 
 
-from .sitemap import parse_sitemap, parse_sitemapindex
+from .sitemap import parse_sitemap, parse_sitemapindex  # noqa: F401
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,8 +19,26 @@ ignore-bad-ideas = *.mo
 ignore = *.rej
 
 [flake8]
-builtins = _
+filename =
+    *.py
+    ./linkchecker
+builtins =
+    _
+    _n
 max-line-length = 80
+per-file-ignores =
+# In several files imports intentionally cause:
+# E402: module level import not at top of file
+# F401: module imported but unused
+    linkchecker: E402
+    setup.py: E402
+    linkcheck/__init__.py: E402,F401
+    linkcheck/bookmarks/chrome.py: E402
+    linkcheck/checker/httpurl.py: E402
+    linkcheck/htmlutil/htmlsoup.py: E402
+    linkcheck/parser/__init__.py: E402
+    tests/__init__.py: F401
+    
 extend-ignore =
 # https://pep8.readthedocs.org/en/latest/intro.html#error-codes
 # these are ignored by default:
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps =
 [testenv:flake8]
 deps = flake8
 skip_install = true
-commands = flake8 setup.py linkcheck {posargs}
+commands = flake8 linkchecker setup.py linkcheck {posargs}
 
 [testenv:check-manifest]
 deps = check-manifest
--- a/linkcheck/bookmarks/chrome.py
+++ b/linkcheck/bookmarks/chrome.py
@@ -59,4 +59,4 @@ def find_bookmark_file(profile="Default"):
     return ""
 
 
-from .chromium import parse_bookmark_data, parse_bookmark_file
+from .chromium import parse_bookmark_data, parse_bookmark_file  # noqa: F401
--- a/linkcheck/checker/__init__.py
+++ b/linkcheck/checker/__init__.py
@@ -176,4 +176,4 @@ def get_index_html(urls):
 
 # all the URL classes
 from . import (fileurl, unknownurl, ftpurl, httpurl, dnsurl,
-    mailtourl, telneturl, nntpurl, ignoreurl, itmsservicesurl)
+    mailtourl, telneturl, nntpurl, ignoreurl, itmsservicesurl)  # noqa: E402


.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
```
anarcat, line 1: hmm... we already have a flake8 block in setup.cfg... could we keep editing it there?

linkcheck/__init__.py
```diff
@@ -22,14 +22,6 @@
 if sys.version_info < (3, 5, 0, 'final', 0):
     import platform
     raise SystemExit("This program requires Python 3.5.0 or later instead of %s." % platform.python_version())
-# require a reasonably recent requests module: 2.4.0 from 2014-08-29
-import requests
```
anarcat, line 26: it looks like this completely removes the import - it was unused?

.flake8
```diff
@@ -0,0 +1,7 @@
...
+builtins =
+	_
+	_n
```
mgedmin, line 4: Ooh, I missed the _n when I added `builtins = _` to setup.cfg!

Speaking of which, we should merge the two flake8 config sections into one file.  Whichever you prefer.

.flake8
```diff
@@ -0,0 +1,7 @@
...
+filename =
+	*.py
+	./linkchecker
```
mgedmin, line 7: TIL!

linkcheck/__init__.py
```diff
@@ -19,25 +19,17 @@
...
-# PEP 396 has only version strings, bummer! PEP 386 is also not helpful.
-requests_version = requests.__version__.split('.')
-# Depends on the version scheme of Python requests
-if int(requests_version[0]) < 2 or \
-   (int(requests_version[0]) == 2 and int(requests_version[1]) < 4):
-    raise SystemExit("This program requires Python requests 2.4.0 or later instead of %s." % requests.__version__)
```
mgedmin, line 32: I suppose we can rely on install_requires to get a sufficiently recent requests version?

linkcheck/better_exchook2.py
```diff
@@ -109,10 +110,6 @@ def grep_full_py_identifiers(tokens):
 		if token[0] in ".0123456789": continue
 		yield token
 
-def set_linecache(filename, source):
-	import linecache
-	linecache.cache[filename] = None, None, [line+'\n' for line in source.splitlines()], filename
-
```
mgedmin, line 115: :+1: for deleting unused code

linkcheck/parser/__init__.py
```diff
@@ -131,4 +131,4 @@ def parse_itms_services(url_data):
             break
 
 
-from .sitemap import parse_sitemap, parse_sitemapindex
+from .sitemap import parse_sitemap, parse_sitemapindex  # noqa: E402,F401
```
mgedmin, line 134: I'm seeing a lot of E402.  I've no idea what it means, but maybe let's put it in the `extends-ignore =` instead of sprinkling noqa comments everywhere?

I remember what F401 means (unused import).  Sometimes you can avoid it by including the imported name in `__all__ = [...]`, to signify explicit re-exports.

anarcat: agreed, i had the same thought

.flake8
anarcat: i don't mean to do a hard block on this. if it's too big of a pain in the back to fix this, don't worry and i'll remove by 'request changes' ;) i just figured you might have missed the entries in setup.cfg that @mgedmin added recently

linkcheck/__init__.py
cjmayo: Not in this file, just there to check the version.

.flake8
cjmayo: Indeed I did (I have had this for a while and published it without checking), so good block.

cjmayo: Moved to setup.cfg.

linkcheck/parser/__init__.py
cjmayo: E402 module level import not at top of file

I got rid of the majority of the `# noqa`, listing per-file-ignores for these. I left them when, like the above, they are creating aliases for other members of the subpackage to show those lines really are there for a reason.

I have found a fair bit of dead code with these (not always imports alone), indeed I just introduced one in the recent PR for the http zero length html fix, so I would be loathe to disable them globally.

cjmayo:
I've just left the direct flake8 commits, I'll put the rest in a new PR.

.flake8
cjmayo: Fewer files is better. Not sure of the balance between a hidden file that is called flake8 and a visible file that is called something else.

setup.cfg
```diff
@@ -15,8 +15,26 @@ python = python
 universal = 0
 
 [flake8]
-builtins = _
+filename =
+    *.py
+    ./linkchecker
+builtins =
+    _
+    _n
```
mgedmin, line 23: Oh, does this work?  flake8 documentation gave me the impression that these must be separated with commas, even if split on multiple lines:
- https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-filename
- https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-builtins

setup.cfg
```diff
@@ -15,8 +15,26 @@ python = python
 universal = 0
 
 [flake8]
-builtins = _
+filename =
+    *.py
+    ./linkchecker
```
mgedmin, line 20: What does this do, exactly?

tox.ini
```diff
@@ -35,4 +35,4 @@ deps =
 [testenv:flake8]
 deps = flake8
 skip_install = true
-commands = flake8 setup.py linkcheck {posargs}
+commands = flake8 linkchecker setup.py linkcheck {posargs}
```
mgedmin, line 38: Having specified `filename =` in the config file, do we have to list them here, again?

Actually this list is different: linkcheck is a directory, not a filename.

setup.cfg
cjmayo: That's for if you just type `flake8` in the top-level directory, else linkchecker is not scanned because it doesn't end in `.py` - the `./` does seem to matter.

tox.ini
cjmayo: This appears to override the `filename =` i.e. it doesn't scan tests or scripts. Maybe we should scan tests?

mgedmin: I would like to scan all Python files.

(My vim has flake8 error highlighting via vim-ale, so I see those warnings all the time.  This incentivizes me to keep my code flake8-clean.)

setup.cfg
cjmayo: Yes, before:
```
$ flake8 --select=F821
./linkcheck/director/console.py:38:15: F821 undefined name '_n'
./linkcheck/director/console.py:41:15: F821 undefined name '_n'
./linkcheck/director/console.py:43:15: F821 undefined name '_n'
./linkcheck/director/console.py:45:15: F821 undefined name '_n'
./linkcheck/logger/html.py:293:24: F821 undefined name '_n'
./linkcheck/logger/html.py:296:20: F821 undefined name '_n'
./linkcheck/logger/html.py:302:20: F821 undefined name '_n'
./linkcheck/logger/html.py:311:24: F821 undefined name '_n'
./linkcheck/logger/__init__.py:278:19: F821 undefined name '_'
./linkcheck/logger/__init__.py:340:16: F821 undefined name '_'
./linkcheck/logger/__init__.py:377:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:380:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:382:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:390:22: F821 undefined name '_'
./linkcheck/logger/text.py:228:20: F821 undefined name '_n'
./linkcheck/logger/text.py:232:24: F821 undefined name '_n'
./linkcheck/logger/text.py:235:24: F821 undefined name '_n'
./linkcheck/logger/text.py:246:22: F821 undefined name '_n'
./linkcheck/logger/text.py:259:26: F821 undefined name '_n'
```

after:
```
$ flake8 --select=F821
./linkcheck/logger/__init__.py:278:19: F821 undefined name '_'
./linkcheck/logger/__init__.py:340:16: F821 undefined name '_'
./linkcheck/logger/__init__.py:377:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:380:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:382:22: F821 undefined name '_'
./linkcheck/logger/__init__.py:390:22: F821 undefined name '_'
```

Whether it is dangerous for the future I don't know. I guess flake8 has to do the same as distutils/setuptools.

tox.ini
mgedmin: OTOH if the choice is between getting linkchecker/ flake8-clean today and making the CI job required vs. getting linkchecker/ and tests/ flake8-clean next year .... it may be better to settle for less but sooner.

I haven't looked at flake8 warnings in tests/.

setup.cfg
mgedmin: I was surprised to see these warnings remain, until I actually looked at the source and saw it do

```python
_ = lambda x: x
Fields = dict(
    realurl=_("Real URL"),
    cachekey=_("Cache key"),
...
    level=_("Level"),
    modified=_("Modified"),
)
del _
```

That'll be fun to untangle :/

mgedmin: Actually, I think it's a flake8 (or, more precisely, pyflakes) bug.  If you have a global that shadows a builtin, and you delete the global, that makes the builtin available again.

mgedmin: https://github.com/PyCQA/pyflakes/issues/548

mgedmin:
Oops there's now a conflict in tox.ini.  I apologize for the inconvenience.

setup.cfg
cjmayo: > [PyCQA/pyflakes#548](https://github.com/PyCQA/pyflakes/issues/548)

Subscribed!

tox.ini
cjmayo: It's almost all formatting, plus a few unused variables, bare excepts.

cjmayo:
> Oops there's now a conflict in tox.ini. I apologize for the inconvenience.

No problem.

Issue #396, opened 2020-05-19
Execution is never finished


aboettger-tuhh:
The execution is never finished.
```
docker run --rm -u nobody linkchecker/linkchecker --recursion-level 5 --threads 20 --no-warnings --check-extern --verbose --ignore-url=/wp-json --ignore-url=/xmlrpc \
            --ignore-url 'https://fonts.gstatic.com/{0,1}$'  $URL
```

Testing all links takes about 10 minutes. After that the following messages appear seemingly endless.

```
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 21 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
 1 thread active,     0 links queued, 3552 links in 4039 URLs checked, runtime 1 hour, 22 minutes
```

mgedmin:
Could run [py-spy](http://pypi.org/p/py-spy) --dump on the linkchecker process to see what that 1 active thread is doing?

(Docker might make this awkward?  Can the host see all the processes in docker containers?  Or would you need to docker exec into the running container and pip install py-spy there?)

aboettger-tuhh:
The host can see all processes:
```
# py-spy dump --pid 1536
Process 1536: /usr/local/bin/python /usr/local/bin/linkchecker --recursion-level 5 --threads 20 --no-warnings --check-extern --verbose --ignore-url=/wp-json --ignore-url=/xmlrpc --ignore-url https://fonts.gstatic.com/{0,1}$ https://...
Python v2.7.16 (/usr/local/bin/python2.7)

Thread 0x7F70B1794600 (idle)
    wait (threading.py:359)
    join (linkcheck/cache/urlqueue.py:204)
[...]

PR #409, opened 2020-05-22
Make sure login form fetching uses a timeout and sends User-Agent

mgedmin:
Also add a couple of XXX comments about using proxy and possibly disabling TLS certificate checking.

The lack of timeout was discovered while studying bug #396, but this is an independent bug.
OTOH with this PR and #408 we should be sure all HTTP fetches use timeouts.

I have not looked at timeout handling in other protocol support, and I don't intend to at the moment. 

There are no tests.  I'm not sure how to test this.

### DIFF
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -79,11 +79,12 @@ class Aggregate:
         if not user and not password:
             raise LinkCheckerError(
                 "loginurl is configured but neither user nor password are set")
-        session = requests.Session()
-        # XXX user-agent header
-        # XXX timeout
+        session = new_request_session(self.config, self.cookies)
         log.debug(LOG_CHECK, "Getting login form %s", url)
-        response = session.get(url)
+        kwargs = dict(timeout=self.config["timeout"])
+        # XXX: proxy?  sslverify?  can we reuse HttpUrl.get_request_kwargs()
+        # somehow?
+        response = session.get(url, **kwargs)
         response.raise_for_status()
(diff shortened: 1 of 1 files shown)


linkcheck/director/aggregator.py
```diff
@@ -79,11 +79,12 @@ def visit_loginurl(self):
         if not user and not password:
             raise LinkCheckerError(
                 "loginurl is configured but neither user nor password are set")
-        session = requests.Session()
-        # XXX user-agent header
-        # XXX timeout
+        session = new_request_session(self.config, self.cookies)
```
cjmayo, line 82: This looks like a worthwhile fix on its own.

linkcheck/director/aggregator.py
```diff
@@ -79,11 +79,12 @@ def visit_loginurl(self):
...
         log.debug(LOG_CHECK, "Getting login form %s", url)
-        response = session.get(url)
+        kwargs = dict(timeout=self.config["timeout"])
+        # XXX: proxy?  sslverify?  can we reuse HttpUrl.get_request_kwargs()
+        # somehow?
+        response = session.get(url, **kwargs)
```
mgedmin, line 87: I am tempted to switch from requests to [httpx](https://www.python-httpx.org/), once httpx 1.0 becomes available.

I looked at the API docs and it seems requests-like, but nicer.

anarcat: curious to discuss this more, but i'd say we should cross that bridge when we get there... what would be the biggest advantage for switching in linkchecker?

PR #410, opened 2020-05-22
Installation of test data, standardise Markdown extension, remove linkchecker.desktop

cjmayo:
I tried building, testing (without a network) and installing a tarball made using sdist and came up with these.

Hopefully using .md for all the Markdown files is not controversial. I can't see an ordinary use for the desktop file and doesn't seem worth the clutter for the majority of users, Gentoo has been deleting it before installation.

I've tried to follow the existing style of the MANIFEST.in, the `recursive-include tests` is getting a bit long.


mgedmin:
I ran `check-manifest` on git master, and I see the sdist is missing

  tests/checker/data/Bookmarks
  tests/checker/data/file.doc

(among other things)

cjmayo:
I've only run the tests which aren't need_network.

I wonder if the simplest solution is to distribute all the checker/data, and simplify the `recursive-include tests` line in the process.

I'm working on the other PRs for 10 for now if you want to fix this.

anarcat:
to be honest, i've never used a MANIFEST in my packages, and i wonder if we really need one...

mgedmin:
I saw you adding a bunch of files under tests/, and then I noticed that there were even more files there that maybe you've missed?  I don't know what files belong in an sdist!

(I mean, I have my personal opinion -- ALL OF THE FILES!  Even files such as .travis.yml that you can't actually _use_ in an sdist, but that may be helpful for other people to see how you're running your tests on your CI.  I just don't know what the lnkchecker project's opinion about this is.)

cjmayo:
Looks like packaging everything only adds about 10KB (a couple of percent) so that would be the simplest.
MANIFEST.in is now at what I need to test and install (as a user) but others may want something different - if it includes everything then everyone should be happy!

anarcat:
i opened #14 to discuss the MANIFEST removal

PR #412, opened 2020-05-23
Remove instances of Python 2 unicode

cjmayo:
Instances in MS Windows specific code still need to be fixed. Mark to be
ignored by flake8, find with:

flake8 --disable-noqa --select=F821

### DIFF
--- a/linkcheck/better_exchook2.py
+++ b/linkcheck/better_exchook2.py
@@ -122,14 +122,14 @@ def pp_extra_info(obj, depthlimit = 3):
 	s = []
 	if hasattr(obj, "__len__"):
 		try:
-			if type(obj) in (str,unicode,list,tuple,dict) and len(obj) <= 5:
+			if type(obj) in (bytes,str,list,tuple,dict) and len(obj) <= 5:
 				pass # don't print len in this case
 			else:
 				s += ["len = " + str(obj.__len__())]
 		except: pass
 	if depthlimit > 0 and hasattr(obj, "__getitem__"):
 		try:
-			if type(obj) in (str,unicode):
+			if type(obj) in (bytes,str):
 				pass # doesn't make sense to get subitems here
 			else:
 				subobj = obj.__getitem__(0)
--- a/linkcheck/bookmarks/chrome.py
+++ b/linkcheck/bookmarks/chrome.py
@@ -18,15 +18,12 @@ import os
 import sys
 from xdg import xdg_config_home
 
-# Windows filename encoding
-nt_filename_encoding="mbcs"
-
 
 def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
         if "LOCALAPPDATA" in os.environ:
-            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)
+            basedir = os.environ["LOCALAPPDATA"]
         else:
             # read local appdata directory from registry
             from ..winutil import get_shell_folder
--- a/linkcheck/bookmarks/chromium.py
+++ b/linkcheck/bookmarks/chromium.py
@@ -20,15 +20,11 @@ import json
 from xdg.BaseDirectory import xdg_config_home
 
 
-# Windows filename encoding
-nt_filename_encoding="mbcs"
-
-
 def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
         if "LOCALAPPDATA" in os.environ:
-            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)
+            basedir = os.environ["LOCALAPPDATA"]
         else:
             # read local appdata directory from registry
             from ..winutil import get_shell_folder
--- a/linkcheck/bookmarks/firefox.py
+++ b/linkcheck/bookmarks/firefox.py
@@ -27,13 +27,10 @@ except ImportError:
 extension = re.compile(r'/places.sqlite$', re.IGNORECASE)
 
 
-# Windows filename encoding
-nt_filename_encoding="mbcs"
-
 def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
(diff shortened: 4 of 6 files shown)


linkcheck/better_exchook2.py
```diff
@@ -122,14 +122,14 @@ def pp_extra_info(obj, depthlimit = 3):
 	s = []
 	if hasattr(obj, "__len__"):
 		try:
-			if type(obj) in (str,unicode,list,tuple,dict) and len(obj) <= 5:
+			if type(obj) in (str,list,tuple,dict) and len(obj) <= 5:
```
mgedmin, line 125: Ah, but this raises a question: should we add bytes in here?

(What even is `better_exchook2.py`?  Is it the thing that produces internal error messages with those very verbose tracebacks?)

linkcheck/better_exchook2.py
```diff
@@ -122,14 +122,14 @@ def pp_extra_info(obj, depthlimit = 3):
...
 				pass # don't print len in this case
 			else:
 				s += ["len = " + str(obj.__len__())]
 		except: pass
 	if depthlimit > 0 and hasattr(obj, "__getitem__"):
 		try:
-			if type(obj) in (str,unicode):
+			if isinstance(obj, str):
```
mgedmin, line 132: bytes?

linkcheck/bookmarks/chrome.py
```diff
@@ -26,7 +26,7 @@ def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
         if "LOCALAPPDATA" in os.environ:
-            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)
+            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)  # noqa: F821
```
mgedmin, line 29: We don't support Python 2 any more, and I'm pretty sure os.environ contains strings on Windows.
```suggestion
            basedir = os.environ["LOCALAPPDATA"]
```

linkcheck/bookmarks/chromium.py
```diff
@@ -28,7 +28,7 @@ def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
         if "LOCALAPPDATA" in os.environ:
-            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)
+            basedir = unicode(os.environ["LOCALAPPDATA"], nt_filename_encoding)  # noqa: F821
```
mgedmin, line 31: Same:
```suggestion
            basedir = os.environ["LOCALAPPDATA"]
```

linkcheck/bookmarks/firefox.py
```diff
@@ -33,7 +33,7 @@
 def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
-        basedir = unicode(os.environ["APPDATA"], nt_filename_encoding)
+        basedir = unicode(os.environ["APPDATA"], nt_filename_encoding)  # noqa: F821
```
mgedmin, line 36: Same: 
```suggestion
        basedir = os.environ["APPDATA"]
```

linkcheck/bookmarks/opera.py
```diff
@@ -28,7 +28,7 @@
 def get_profile_dir():
     """Return path where all profiles of current user are stored."""
     if os.name == 'nt':
-        basedir = unicode(os.environ["APPDATA"], nt_filename_encoding)
+        basedir = unicode(os.environ["APPDATA"], nt_filename_encoding)  # noqa: F821
```
mgedmin, line 31:
```suggestion
        basedir = os.environ["APPDATA"]
```

linkcheck/better_exchook2.py
cjmayo: > (What even is `better_exchook2.py`? Is it the thing that produces internal error messages with those very verbose tracebacks?)

Good point - yes it is only used in that one place.

Can we replace it? Is there anything now in the standard library that would be good enough? The one thing I do like is the clear printing of values.

mgedmin: I don't think there's anything in the standard library.

Other traceback printing alternatives that I know of are

- Sentry, which captures the stack trace with locals and submits it to a webserver somewhere for aggregation, not suitable for linkchecker
- zope.exceptions has a custom traceback formatter that annotates each stack frame with the value of special `__traceback_info__` attribute, not a good choice for linkchecker
- pytest has a custom traceback formatter for test failures, I'm not sure if it's extractable

I haven't done any pypi or google searches for rich Python traceback libraries.

I think "don't fix what ain't broke" applies here and we can keep better_exchook2.  Having said that, I still think it may be valuable to restore special handling of bytestring values that as accidentally lost in the Python 3 transition.  In case bytestrings show up in locals, which has happened.  The thing is I'm not sure what the special handling _is_ because I haven't looked at the surrounding code.  Is it printing the length?  Truncating long values?  (We'd want that for bytestrings, I'm sure.)

cjmayo: Thanks for describing those options. I think adding a dependency for this would need some careful thought. No problem adding bytes to the lists.

linkcheck/better_exchook2.py
cjmayo: Done.

linkcheck/bookmarks/chrome.py
cjmayo: Good enough for me.

linkcheck/bookmarks/chromium.py
cjmayo: Done.

linkcheck/bookmarks/firefox.py
cjmayo: Done.

linkcheck/bookmarks/opera.py
cjmayo: Done.

Issue #322, opened 2019-10-21
changelog doesn't mention 9.4 release date


mgedmin:
According to GitHub, the 9.4 release [exists](https://github.com/linkchecker/linkchecker/releases/tag/v9.4.0), but [`doc/changelog.txt`](https://github.com/linkchecker/linkchecker/blob/master/doc/changelog.txt) doesn't have a codename for it, nor a release date.

It might also be a good idea to compare the list of features/changes/fixes with the one [published on github](https://github.com/linkchecker/linkchecker/releases/tag/v9.4.0), to make sure it's complete.

mgedmin:
@anarcat I can do this, but I'd like to focus on the rest of `get_content()` calls from https://github.com/linkchecker/linkchecker/issues/317#issuecomment-544536320 so maybe you'd like to volunteer for this? ;)

anarcat:
> @anarcat I can do this, but I'd like to focus on the rest of get_content() calls from #317 (comment) so maybe you'd like to volunteer for this? ;)

sorry, i'm barely keeping up with the flood of reviews. ;)

anarcat:
> @anarcat I can do this, but I'd like to focus on the rest of get_content() calls from #317 (comment) so maybe you'd like to volunteer for this? ;)

sorry, i'm barely keeping up with the flood of reviews. ;)

if/when we do fix this, also update `doc/development.mdwn` to make sure we don't make this mistake again.

mgedmin:
Looks like https://github.com/linkchecker/linkchecker/blob/master/doc/development.mdwn#release-process mentions the changelog update now.

PR #413, opened 2020-05-23
Remove ineffective command-line options

cjmayo:
These options were replaced by plugins in:

7b34be59 ("Introduce check plugins, use Python requests for http/s
connections, and some code cleanups and improvements.", 2014-03-01)

Released in 9.0.

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -1,3 +1,8 @@
+10.0 "" (released xx.xx.xxxx)
+
+Changes:
+- cmdline: Remove options replaced by plugins and made ineffective in 9.0
+
 9.4 "just passing by" (released 12.4.2018)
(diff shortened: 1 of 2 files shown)


cjmayo:
> I think it's important to clarify what "replaced" means here. 

Changed the commit message (and PR title):

Subject: Remove ineffective command-line options

These options were replaced by plugins and made ineffective [1]. This change
was included in the 9.0 release.

[1] 7b34be59 ("Introduce check plugins, use Python requests for http/s
    connections, and some code cleanups and improvements.", 2014-03-01)

cjmayo:
> this should include, at least, a changelog entry.

First added a commit that fixes #322 and then started the entry for 10.0.

PR #417, opened 2020-05-25
Adjust tests for need_network

cjmayo:
Nice to enable some tests without the internet, important to mark one that can't:

Subject: Mark TestFile.test_html_url_quote as need_network

Else without the internet the test fails, eventually, with:

warning No MX mail host for users.sourceforge.net found

### DIFF
--- a/tests/checker/test_file.py
+++ b/tests/checker/test_file.py
@@ -22,7 +22,7 @@ import zipfile
 
 import pytest
 
-from tests import need_word, need_pdflib
+from tests import need_network, need_word, need_pdflib
 from . import LinkCheckTest, get_file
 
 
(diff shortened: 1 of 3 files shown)


mgedmin:
How are you testing network-less test runs?  I've used [Bubblewrap](https://github.com/containers/bubblewrap) in the past, it's nice:

```
bwrap --unshare-net --dev-bind / / -- $(which tox) -e py38
```

although maybe it's too strict, because it doesn't allow the test process to talk to clamd on localhost, breaking three tests.

cjmayo:
> How are you testing network-less test runs? I've used [Bubblewrap](https://github.com/containers/bubblewrap) in the past, it's nice:

I'm just using the standard Gentoo packaging process (well a custom ebuild of my own that pulls the code from my local repository and runs pytest) with FEATURES="network-sandbox" (which is pretty normal to stop build systems from accessing the internet). The advantage of that is it does all the building and testing away from my development area and is one easy command.

PR #420, opened 2020-05-26
Update GNOME proxy support for GNOME 3 and Python 3

cjmayo:
GConf is replaced by dconf and the GSettings API in GNOME 3.

---

I don't have a proxy setup, but the code being replaced definitely won't work.

### DIFF
--- a/doc/de.po
+++ b/doc/de.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: linkchecker 3.4\n"
-"POT-Creation-Date: 2020-06-04 17:30+0100\n"
+"POT-Creation-Date: 2020-06-05 16:18+0100\n"
 "PO-Revision-Date: 2020-06-04 17:33+0100\n"
 "Last-Translator: Chris Mayo <aklhfex@gmail.com>\n"
 "Language-Team: German - Germany <>\n"
@@ -25,9 +25,10 @@ msgid "LINKCHECKER"
 msgstr "LINKCHECKER"
 
 #. type: TH
-#: en/linkchecker.1:1 en/linkcheckerrc.5:1
-#, no-wrap
-msgid "2020-04-24"
+#: en/linkchecker.1:1
+#, fuzzy, no-wrap
+#| msgid "2020-04-24"
+msgid "2020-06-05"
 msgstr "2020-04-24"
 
 # type: TH
@@ -1116,12 +1117,20 @@ msgstr "PROXY UNTERSTÜTZUNG"
 # type: Plain text
 #. type: Plain text
 #: en/linkchecker.1:303
+#, fuzzy
+#| msgid ""
+#| "To use a proxy on Unix or Windows set the $http_proxy, $https_proxy or "
+#| "$ftp_proxy environment variables to the proxy URL. The URL should be of "
+#| "the form B<http://>[I<user>B<:>I<pass>B<@>]I<host>[B<:>I<port>].  "
+#| "LinkChecker also detects manual proxy settings of Internet Explorer under "
+#| "Windows systems, and gconf or KDE on Linux systems.  On a Mac use the "
+#| "Internet Config to select a proxy."
 msgid ""
 "To use a proxy on Unix or Windows set the $http_proxy, $https_proxy or "
 "$ftp_proxy environment variables to the proxy URL. The URL should be of the "
 "form B<http://>[I<user>B<:>I<pass>B<@>]I<host>[B<:>I<port>].  LinkChecker "
 "also detects manual proxy settings of Internet Explorer under Windows "
-"systems, and gconf or KDE on Linux systems.  On a Mac use the Internet "
+"systems, and GNOME or KDE on Linux systems.  On a Mac use the Internet "
 "Config to select a proxy."
 msgstr ""
 "Um einen Proxy unter Unix oder Windows zu benutzen, setzen Sie die "
@@ -1755,6 +1764,12 @@ msgstr "Copyright \\(co 2000-2014 Bastian Kleineidam"
 #, no-wrap
 msgid "LINKCHECKERRC"
 msgstr "LINKCHECKERRC"
+
+#. type: TH
+#: en/linkcheckerrc.5:1
+#, no-wrap
+msgid "2020-04-24"
+msgstr "2020-04-24"
 
 # type: Plain text
 #. type: Plain text
--- a/doc/de.po
+++ b/doc/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: linkchecker 3.4\n"
 "POT-Creation-Date: 2020-06-05 16:18+0100\n"
-"PO-Revision-Date: 2020-06-04 17:33+0100\n"
+"PO-Revision-Date: 2020-06-11 19:31+0100\n"
 "Last-Translator: Chris Mayo <aklhfex@gmail.com>\n"
 "Language-Team: German - Germany <>\n"
 "Language: de_DE\n"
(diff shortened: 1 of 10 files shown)


cjmayo:
More files to update gconf -> dconf.

cjmayo:
OK, I've updated what I can. Need some help with what to do about the translations.

doc/install.txt
```diff
@@ -67,7 +67,9 @@ First, install the required software.
    ClamAv from http://www.clamav.net/
 
 7. *Optional, for GNOME proxy setting parsing:*
-    Python Gtk from http://www.pygtk.org/downloads.html
+    PyGObject, GIO and its repository file.
+    Best installed from your distribution e.g. ``python3-gi`` and
+    ``libgirepository1.0-dev``
```
mgedmin, line 72: Is libgirepository1.0-dev needed?

> This package contains the headers for the C library for handling the
> introspection data.

AFAIU all that pygobject-introspection needs is typelib files, packaged in gir1.2-glib-2.0, which is pulled in by python3-gi.

linkcheck/configuration/__init__.py
```diff
@@ -428,43 +428,45 @@ def get_user_config():
...
 
 
-def get_gconf_http_proxy():
-    """Return host:port for GConf HTTP proxy if found, else None."""
+def get_dconf_http_proxy():
+    """Return host:port for dconf HTTP proxy if found, else None."""
     try:
-        import gconf
+        import gi
+        from gi.repository import Gio
```
mgedmin, line 435: I often see warnings telling me to call `gi.require_version('Gio', '2.0')` and such before importing stuff from gi.repository.

linkcheck/configuration/__init__.py
```diff
@@ -428,43 +428,45 @@ def get_user_config():
...
     except ImportError:
         return None
     try:
-        client = gconf.client_get_default()
-        if client.get_bool("/system/http_proxy/use_http_proxy"):
-            host = client.get_string("/system/http_proxy/host")
-            port = client.get_int("/system/http_proxy/port")
+        settings = Gio.Settings.new("org.gnome.system.proxy.http")
```
mgedmin, line 439: Ok, so this uses gsettings, not dconf.  There's a subtle difference.

gsettings can use different backends, and dconf is only one of them (which happens to be the default backend, unless you're running in a flatpak app container).

If we use gsettings, we'd better refer to it (rather than dconf) in documentation and function names.  (Although maybe a more user-friendly way would be to say we support GNOME's proxy configuration without going into details?)

linkcheck/configuration/__init__.py
```diff
@@ -428,43 +428,45 @@ def get_user_config():
...
+        from gi.repository import Gio
     except ImportError:
         return None
     try:
-        client = gconf.client_get_default()
-        host = client.get_string("/system/proxy/ftp_host")
-        port = client.get_int("/system/proxy/ftp_port")
+        settings = Gio.Settings.new("org.gnome.system.proxy.ftp")
+        host = settings.get_string("host")
+        port = settings.get_int("port")
```
mgedmin, line 463: This just begs for code reuse, although, hmm, the http proxy has an enabled flag (and a bunch of authentication options what we ignore?), but the ftp (and https) one doesn't.

setup.cfg
```diff
@@ -35,6 +35,7 @@ per-file-ignores =
     linkcheck/__init__.py: E402,F401
     linkcheck/bookmarks/chrome.py: E402
     linkcheck/checker/httpurl.py: E402
+    linkcheck/configuration/__init__.py: F401
```
mgedmin, line 38: I'm guessing this is from `import gi` and it would go away if you called `gi.require_version()`.

linkcheck/configuration/__init__.py
mgedmin: > If we use gsettings, we'd better refer to it (rather than dconf) in documentation and function names

and error messages!  I forgot to mention error messages.

doc/install.txt
cjmayo: Oops yes, thanks. It is me that likes reading the gir files not PyGObject.

linkcheck/configuration/__init__.py
cjmayo: Gio is the one I don't see those warnings for, but maybe it is safer. I'll take it for the reasons below anyway.

linkcheck/configuration/__init__.py
cjmayo: GNOME is better here and matches the use of KDE.

setup.cfg
cjmayo: Done.

linkcheck/configuration/__init__.py
cjmayo: Done.

cjmayo:
OK ready to go now, a real V2 incorporating the comments and updating translation files (without any new translations).

doc/de.po
```diff
@@ -25,9 +25,10 @@ msgid "LINKCHECKER"
...
 
 #. type: TH
-#: en/linkchecker.1:1 en/linkcheckerrc.5:1
-#, no-wrap
-msgid "2020-04-24"
+#: en/linkchecker.1:1
+#, fuzzy, no-wrap
+#| msgid "2020-04-24"
+msgid "2020-06-05"
 msgstr "2020-04-24"
```
mgedmin, line 32: lol

doc/de.po
```diff
@@ -1116,12 +1117,20 @@ msgstr "PROXY UNTERSTÜTZUNG"
 # type: Plain text
 #. type: Plain text
 #: en/linkchecker.1:303
+#, fuzzy
```
mgedmin, line 1120: This adds a `fuzzy` marker, which means this message will _not_ be translated.

The old translation mentions Windows and Mac but omits Linux.  I suggest dropping `fuzzy` and keeping the old translation.

(Maybe we could contact the last translator -- Chris Mayo <aklhfex@gmail.com> -- and ask them to add the missing sentence about Linux systems?  Oh, wait, is that you, @cjmayo?  Do you speak German?)

doc/de/linkchecker.1
```diff
@@ -298,12 +298,12 @@ Das untige Beispiel sendet zwei Cookies zu allen URLs die mit
...
-URL. Die URL sollte die Form
-\fBhttp://\fP[\fIuser\fP\fB:\fP\fIpass\fP\fB@\fP]\fIhost\fP[\fB:\fP\fIport\fP] besitzen. LinkChecker
-erkennt auch die Proxy\-Einstellungen des Internet Explorers auf einem
-Windows\-System. Auf einem Mac benutzen Sie die Internet Konfiguration.
+To use a proxy on Unix or Windows set the $http_proxy, $https_proxy or
+$ftp_proxy environment variables to the proxy URL. The URL should be of the
+form \fBhttp://\fP[\fIuser\fP\fB:\fP\fIpass\fP\fB@\fP]\fIhost\fP[\fB:\fP\fIport\fP].  LinkChecker
+also detects manual proxy settings of Internet Explorer under Windows
+systems, and GNOME or KDE on Linux systems.  On a Mac use the Internet
+Config to select a proxy.
```
mgedmin, line 306: Yeah, this is the regression because of that `fuzzy` flag.

doc/de.po
anarcat: wtf

mgedmin: Looks like the automated man page translation tool wants to translate everything, including the release date.

(Which maybe makes sense?  Different languages might use different date formats.)

anarcat: i would hope YYYY-MM-DD would be agreed on internationally at this point, but maybe you're right. would still seem crazy to translate that. :)

cjmayo: Copied the `2020-06-05` across.

doc/de.po
cjmayo: Well, I was tempted to translate that one. I guess I better had then...

PR #425, opened 2020-05-27
Fix xdg_config_home import in bookmarks/chrome.py

cjmayo:
```
>>> from xdg import xdg_config_home
ImportError: cannot import name 'xdg_config_home' from 'xdg' (/usr/lib/python3.7/site-packages/xdg/__init__.py)
```

Fix taken from the other bookmarks files.

This file isn't actually used by linkchecker, fix in case it saves anyone time in future.

### DIFF
--- a/linkcheck/bookmarks/chrome.py
+++ b/linkcheck/bookmarks/chrome.py
@@ -16,7 +16,7 @@
 
 import os
 import sys
-from xdg import xdg_config_home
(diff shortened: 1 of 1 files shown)


anarcat:
aren't we just removing this file?

cjmayo:
Probably (at some point), although perhaps it would be good to move it to linkchecker-gui, so perhaps worth making it correct (while I remember).

PR #426, opened 2020-05-28
Remove unused httputil.has_header_value() and bookmarks code

### LINKED ISSUE #422: linchecker-gui bookmarks code present in linkchecker
bookmarks/chrome.py in its entirety and all find_bookmark_file() (and supporting) functions are used by linchecker-gui but not linkchecker.

https://github.com/linkchecker/linkchecker-gui/blob/ccbd82316ac7671ea97c3468eaac7cd10211ab3e/linkcheck_gui/lineedit.py#L199-L226

cjmayo:
linkchecker does not need to find a bookmark file, it is given the URL.
Most bookmarks are detected by their MIME type, Firefox is different
because it uses a SQLite database

---

OK, let's remove the linkcheck-gui bookmarks code, I'll add a link to the commit from the issue in its project.

Also an unused function in httputil.

I don't plan looking for any more unused code.

Closes #422


anarcat:
are there docs refering to this that need to be removed?

PR #427, opened 2020-05-28
Resolve flake8 violations in linkchecker and setup.py

cjmayo:
black does some interesting thing to setup.py.

### DIFF
--- a/linkchecker
+++ b/linkchecker
@@ -264,9 +264,9 @@ group.add_argument(
     dest="configfile",
     metavar="FILENAME",
     help=_(
-        """Use FILENAME as configuration file. Per default LinkChecker uses
-~/.linkchecker/linkcheckerrc (under Windows
-%%HOMEPATH%%\\.linkchecker\\linkcheckerrc)."""
+        "Use FILENAME as configuration file. Per default LinkChecker uses\n"
+        "~/.linkchecker/linkcheckerrc (under Windows\n"
+        "%%HOMEPATH%%\\.linkchecker\\linkcheckerrc)."
     ),
 )
 group.add_argument(
@@ -275,23 +275,23 @@ group.add_argument(
     type=int,
     metavar="NUMBER",
     help=_(
-        """Generate no more than the given number of threads. Default number
-of threads is 10. To disable threading specify a non-positive number."""
+        "Generate no more than the given number of threads. Default number\n"
+        "of threads is 10. To disable threading specify a non-positive number."
     ),
 )
 group.add_argument(
-    "-V", "--version", action="store_true", help=_("""Print version and exit.""")
+    "-V", "--version", action="store_true", help=_("Print version and exit.")
 )
 group.add_argument(
     "--list-plugins",
     action="store_true",
     dest="listplugins",
-    help=_("""Print available check plugins and exit."""),
+    help=_("Print available check plugins and exit."),
 )
 group.add_argument(
     "--stdin",
     action="store_true",
-    help=_("""Read list of white-space separated URLs to check from stdin."""),
+    help=_("Read list of white-space separated URLs to check from stdin."),
 )
 
 # ================== output options =====================
@@ -302,13 +302,13 @@ group.add_argument(
     action="append",
     metavar="STRING",
     help=_(
-        """Print debugging output for the given logger.
-Available loggers are %(lognamelist)s.
-Specifying 'all' is an alias for specifying all available loggers.
-The option can be given multiple times to debug with more
-than one logger.
-
-For accurate results, threading will be disabled during debug runs."""
+        "Print debugging output for the given logger.\n"
+        "Available loggers are %(lognamelist)s.\n"
+        "Specifying 'all' is an alias for specifying all available loggers.\n"
+        "The option can be given multiple times to debug with more\n"
+        "than one logger.\n"
+        "\n"
+        "For accurate results, threading will be disabled during debug runs."
     )
     % {"lognamelist": logconf.lognamelist},
 )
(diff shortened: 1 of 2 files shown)


linkchecker
```diff
@@ -204,53 +245,81 @@ def has_encoding(encoding):
...
+    ),
+)
+group.add_argument(
+    "-t",
+    "--threads",
+    type=int,
+    metavar="NUMBER",
+    help=_(
+        """Generate no more than the given number of threads. Default number
+of threads is 10. To disable threading specify a non-positive number."""
```
mgedmin, line 279: I strongly want to convert these tripple-quoted strings into regular concatenated string literals like this

```suggestion
        "Generate no more than the given number of threads. Default number\n"
        "of threads is 10. To disable threading specify a non-positive number."
```

because the way they break the flow of indentation makes reading this code hard.

Alternatively we could use textwrap.dedent(), but when the strings are 2 or 3 lines I think that's overkill.

linkchecker
```diff
@@ -419,70 +580,83 @@ if options.output:
     logtype = logtype.lower()
     if logtype not in linkcheck.logger.LoggerNames:
         print_usage(
-     _("Unknown logger type %(type)r in %(output)r for option %(option)s") % \
-      {"type": logtype, "output": options.output, "option": "'-o, --output'"})
-    if logtype != 'none' and not has_encoding(encoding):
+            _("Unknown logger type %(type)r in %(output)r for option %(option)s")
+            % {"type": logtype, "output": options.output, "option": "'-o, --output'"}
```
mgedmin, line 584: This is actually a genius solution, to prevent translators from translating option names unnecessarily/misleadingly!

linkchecker
```diff
@@ -491,25 +665,29 @@ if options.timeout is not None:
...
     config.add_auth(pattern=".+", user=_username, password=_password)
 # read missing passwords
 for entry in config["authentication"]:
     if entry["password"] is None:
         attrs = entry.copy()
         attrs["strpattern"] = attrs["pattern"].pattern
-        msg = _("Enter LinkChecker password for user %(user)s" \
-                " at %(strpattern)s:") % attrs
+        msg = (
+            _("Enter LinkChecker password for user %(user)s" " at %(strpattern)s:")
```
mgedmin, line 688:
```suggestion
            _("Enter LinkChecker password for user %(user)s at %(strpattern)s:")
```

linkchecker
```diff
@@ -547,22 +725,32 @@ else:
 if do_profile:
     if has_profile:
         if os.path.exists(_profile):
-            print(_("""Overwrite profiling file %(file)r?
-Press Ctrl-C to cancel, RETURN to continue.""") % {"file": _profile})
+            print(
+                _(
+                    """Overwrite profiling file %(file)r?
+Press Ctrl-C to cancel, RETURN to continue."""
```
mgedmin, line 731:
```suggestion
                    "Overwrite profiling file %(file)r?\n"
                    "Press Ctrl-C to cancel, RETURN to continue."
```

(I am assuming xgettext or whatever is used to extract translations from the source files will see these strings as the same, because they are)

linkchecker
```diff
@@ -571,6 +759,7 @@ else:
     check_urls(aggregate)
 if config["debugmemory"]:
     import linkcheck.memoryutil
+
     if has_meliae:
         log.info(LOG_CMDLINE, _("Dumping memory statistics..."))
         filename = linkcheck.memoryutil.write_memory_dump()
```
mgedmin, line 765: I'm learning so much!

setup.py
```diff
@@ -278,87 +300,89 @@ def run(self):
...
     # Requirements, usable with setuptools or the new Python packaging module.
-    python_requires = '>= 3.5',
-    install_requires = [
-        'requests >= 2.4',
-        'dnspython',
-        'beautifulsoup4',
-        'pyxdg',
-    ],
+    python_requires=">= 3.5",
+    install_requires=["requests >= 2.4", "dnspython", "beautifulsoup4", "pyxdg"],
```
mgedmin, line 379: No, let's keep install_requires on separate lines for readability and future extensibility.

linkchecker
mgedmin: Also, these are inside _(), so textwrap.dedent() would mess up translation extraction, I'm sure.

setup.py
anarcat: problem with manual changes like this is that black will either complain or break them on the next run... if we go black, we go all the way...

mgedmin: We don't have to?  We can use black to fix flake8/pycodestyle, and then not use black any more.  flake8 in the CI will prevent style regressions.

anarcat: i am fine either way

linkchecker
cjmayo: I've done those - the only triples remaining are in the globals.

linkchecker
cjmayo: Done, I'll grep for those in future.

linkchecker
cjmayo: Done.

setup.py
cjmayo: Done.

My personal goal was to make flake8 usable on the command line and enforceable in Travis. There are too many formatting problems to fix by hand, the only other option would be to exclude a number of checks. black was a means to an end for me. Not that it couldn't be part of a workflow in the future (would be quite an adjustment) but the bugs that have shown up in the review seem to suggest that is a way off.

cjmayo:
Added a couple of commits in response to the review comments.

PR #430, opened 2020-05-29
Fix issue 258 (Proxy with HTTPS links)

### LINKED ISSUE #258: http server cannot be used as a https proxy
## Summary

Linkchecker does not pass proxy for HTTPS connections, even when `https_proxy` environment variable is present. LC shows that proxy is detected and configured, but actual connections go directly (being block by firewall in my case). It seems that applying proxy in [httpurl.py#L248](https://github.com/linkchecker/linkchecker/blob/ecba4e7ac861d45d6afb2366e5607b9d2b857b38/linkcheck/checker/httpurl.py#L248) is not passing the correct proxy type. When I replace `self.proxytype` with `"https"` the connections work well via HTTPS ( maybe there should be self.scheme used? - sorry didn't analyze further ).

## Steps to reproduce

 1. export https_proxy=http://myproxy:8080
 2. linkchecker https://www.my-website.com

## Actual result
Connection does not use proxy and fails.

## Expected result
Connection uses proxy and succeeds.

## Environment
 
 * Operating system: `CentOS 7 Linux iuwejsdfjiasjd 3.10.0-957.1.3.el7.x86_64 #1 SMP Thu Nov 15 17:36:42 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
 * Linkchecker version: `LinkChecker 9.4.0`
 * Python version: `Python 2.7.5`
 * Install method: `pip install from git on Apr 29, 2019 10:33:07 AM`
 * Site URL: No public site

## Configuration file

No configuration file

c-markwell:
Fixes #258.

### DIFF
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -231,7 +231,7 @@ class HttpUrl(internpaturl.InternPatternUrl, proxysupport.ProxySupport):
         Session.resolve_redirects()."""
         kwargs = dict(stream=True, timeout=self.aggregate.config["timeout"])
         if self.proxy:
-            kwargs["proxies"] = {self.proxytype: self.proxy}
+            kwargs["proxies"] = {self.scheme: self.proxy}
         if self.scheme == "https" and self.aggregate.config["sslverify"]:
             kwargs['verify'] = self.aggregate.config["sslverify"]
         else:


mgedmin:
There's similar code in linkcheck/cache/robots_txt.py that probably should be changed the same way:
https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/cache/robots_txt.py#L61-L62

mgedmin:
Current test coverage data indicates that we have zero tests that use a proxy.  I say let's merge this without blocking on a test.

But it would be good to fix the robots.txt fetching code to use https proxies correctly before merging.  @c-markwell, could you do that?  Or check the checkbox that allows maintainer edits in the right sidebar?
Ekrano nuotrauka iš 2020-05-29 22-41-17

mgedmin:
I had another thought: what about redirects that change the scheme?

If we do `requests.get('http://example.com', proxies={'http': 'http://localhost:8080'})`, and example.com redirects to https://example.com, requests will not know to use an https proxy!

This is not actually that straightforward: linkchecker sets allow_redirects=False and follows them manually.  But the way it does that:
https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/httpurl.py#L241-L245
it calls requests.Session.resolve_rediects() to follow a sequence of redirects, but that method gets only one proxy configured for one scheme from self.get_request_kwargs()!

I think the fix in this MR is insufficient, and instead we should be passing the full proxy configuration to  requests.

c-markwell:
> But it would be good to fix the robots.txt fetching code to use https proxies correctly before merging. @c-markwell, could you do that? Or check the checkbox that allows maintainer edits in the right sidebar?

I've just checked and the "Allow edits by maintainers" checkbox on this page is already ticked. It'd not be straightforward for me to test the robots fix at the moment, so if you can add that, that'd be great. Thanks for your help @mgedmin

cjmayo:
Comment above copied to a new issue (430). But even in its limited scope, think this change is a bit of a shorcut - that as proxytype exists it should be set to the correct scheme (but obviously that means more change as ProxySupport.set_proxy() currently assumes the type is the same as the scheme of the proxy server).

PR #437, opened 2020-06-04
Update man page translation and fixes for application translation process

cjmayo:
Changes to man page format in:
a205a372 ("Update man pages to optimise for both html and man", 2020-04-24)
broke the mapping in doc/de.po, so update the .pot and de.po and then move existing German strings and regenerate the de man pages.

Also some fixes for the application translation generation and a basic explanation of the use of make in a new document translations.md.


cjmayo:
Typo fixed.

Issue #439, opened 2020-06-04
Issues arising from black linkcheck/ review


cjmayo:
While reviewing the results of black on linkcheck/ a number of comments on the code were raised.
See: #432

mgedmin:
Things that seem worth bothering with:

- [ ] suboptimal use of startswith() in [confparse.py](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433635595), maybe reuse HTTP_SCHEMES
- [ ] there's a missing `_()` wrapper in [confparse.py](https://github.com/linkchecker/linkchecker/pull/432/files/3c9e6266d7bef890e22eac10ee5a3f724f78b18e#diff-51589283e29e7948ca02748855a6d237), around the message about "The configuration file %s contains password information ..."
- [ ] grep the codebase for (real) http:// urls and change them to https://, after verifying that the HTTPS version works -- [example](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433637910): "See http://support.microsoft.com/kb/308419 for ..."
  (that particular example is tricky in that it's inside a translatable string; we might want to replace it with %(url)s), and there's [more](https://github.com/linkchecker/linkchecker/pull/432#discussion_r434395552)
- [ ] grep the codebase for [`os.linesep`](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433640467) and make sure those are emitted only to files opened without newline conversion enabled (e.g. log files opened in binary mode, which is the habit for all the loggers); I'm also [suspicious about SQL literals](https://github.com/linkchecker/linkchecker/pull/432#discussion_r434396882)
- [ ] check [ftpparse.py](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433641245) filename handling code because it appears to be mishandling filenames with consecutive spaces in them
- [ ] there's an `except AttributeError: pass` in [csvlog.py somewhere](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433789366)
- [ ] check [csvlog.py](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433790532) for columns that may vary per row, depending on available information (line/column numbers)
- [ ] check [logger/html.py](https://github.com/linkchecker/linkchecker/pull/432#discussion_r433791663) for HTML injection problems
- [ ] long regexp flags in [markdowncheck.py](https://github.com/linkchecker/linkchecker/pull/432#discussion_r434425891)
[...]

PR #440, opened 2020-06-05
Assorted review comments and use python3 for setup.py

cjmayo:
Various review comments from #439 and before.

Plus use python3 in setup.py. I think this might save some reports  similar to issues we have seen - avoid an error for an Ubuntu 20.04 user with Python 2 installed who runs `./setup.py`.

### DIFF
--- a/linkcheck/configuration/confparse.py
+++ b/linkcheck/configuration/confparse.py
@@ -233,9 +233,11 @@ class LCConfigParser(RawConfigParser):
         if fileutil.is_accessable_by_others(fn):
             log.warn(
                 LOG_CHECK,
-                "The configuration file %s contains password information (in"
-                " section [%s] and options %s) and the file is readable by"
-                " others. Please make the file only readable by you.",
+                _(
+                    "The configuration file %s contains password information (in"
+                    " section [%s] and options %s) and the file is readable by"
+                    " others. Please make the file only readable by you."
+                ),
                 fn,
                 section,
                 fields,
--- a/linkcheck/configuration/confparse.py
+++ b/linkcheck/configuration/confparse.py
@@ -248,9 +248,8 @@ class LCConfigParser(RawConfigParser):
                 log.warn(
                     LOG_CHECK,
                     _(
-                        "See http://support.microsoft.com/kb/308419 for"
-                        " more info on setting file permissions."
-                    ),
+                        "See %(url)s for more info on setting file permissions."
+                    ) % {"url": "https://support.microsoft.com/kb/308419"}
                 )
 
     def read_filtering_config(self):
--- a/tests/test_po.py
+++ b/tests/test_po.py
@@ -34,6 +34,8 @@ def get_pofiles():
         pofiles = []
         pofiles.extend(glob.glob("po/*.po"))
         pofiles.extend(glob.glob("doc/*.po"))
+    if not pofiles:
+        raise Exception("No .po files found")
     return pofiles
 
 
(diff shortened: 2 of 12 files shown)


tests/test_po.py
```diff
@@ -56,15 +58,15 @@ class TestGTranslator(unittest.TestCase):
     def test_gtranslator(self):
         """Test all pofiles for GTranslator brokenness."""
         for f in get_pofiles():
-            with open(f, "rb") as fd:
+            with open(f) as fd:
```
mgedmin, line 61: `encoding="UTF-8"`?  All of the .po files in the source tree use UTF-8.

linkcheck/configuration/confparse.py
```diff
@@ -246,9 +248,8 @@ def check_password_readable(self, section, fields):
                 log.warn(
                     LOG_CHECK,
                     _(
-                        "See http://support.microsoft.com/kb/308419 for"
-                        " more info on setting file permissions."
-                    ),
+                        "See %(url)s for more info on setting file permissions."
+                    ) % {"url": "https://support.microsoft.com/kb/308419"}
                 )
```
mgedmin, line 253: This needs a corresponding change in po/de.po, where this message is currently translated.

(es.po and fr.po don't translate it so there's nothing to lose)

tests/test_po.py
cjmayo: Thanks, I just assume everyone must be on UTF-8 in 2002... Done in a new commit.

linkcheck/configuration/confparse.py
cjmayo: I think that will clash with the GNOME proxy commit which has also updated de.po. I think this PR is worth having (in case anyone is using W3C validation), and it saves a rebase...
I will submit another PR after the GNOME one for this.

mgedmin: Please do!

mgedmin: Ping?

cjmayo: I did try it but have been undecided. It is very sensitive to the environment, I get:

```
 po/de.mo           |  Bin 33015 -> 32536 bytes
 po/de.po           | 1714 ++++++++++++++++++++++++++++++++++++++++++++--------------------------------------------
 po/es.po           | 1617 ++++++++++++++++++++++++++++++++++++++++++-----------------------------------------
 po/fr.po           | 1613 +++++++++++++++++++++++++++++++++++++++++-----------------------------------------
 po/linkchecker.pot | 1287 +++++++++++++++++++++++++++++++++---------------------------------
 5 files changed, 3146 insertions(+), 3085 deletions(-)
```

an awful lot of just moving things around which makes that diff pretty unintelligible. OTOH it might make the next one clearer - if I create it.

 I don't really mind. I'm happy to submit that if you still want it.

cjmayo: That is to say, there will need to be an update. The question is just whether to do one now or only after the currently open PRs.

mgedmin: What about a simple manual :%s,https://support.microsoft.com/kb/308419,%(url)s, in the pot and po files?  Just so this doesn't get forgotten.

(I've intentionally been staying away from reviewing other PRs because I didn't want to forget about this one.)

cjmayo: I've created  #447, to be safe by copying the individual change in the pot file, regenerating the po's and editing in gtranslator, and then generating the mo.

PR #442, opened 2020-06-05
Remove strformat.unicode_safe()

cjmayo:
I don't have the energy to check each use individually - this is following the rip it out and fix the tests approach.

What do you think?

### DIFF
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -110,7 +110,7 @@ class HttpUrl(internpaturl.InternPatternUrl, proxysupport.ProxySupport):
             # Note that content-encoding causes size differences since
             # the content data is always decoded.
             try:
-                self.size = int(self.getheader("Content-Length"))
+                self.size = int(self.headers["Content-Length"])
             except (ValueError, OverflowError):
                 pass
         else:
@@ -284,17 +284,6 @@ class HttpUrl(internpaturl.InternPatternUrl, proxysupport.ProxySupport):
                 # run connection plugins for old connection
                 self.aggregate.plugin_manager.run_connection_plugins(self)
 
-    def getheader(self, name, default=None):
-        """Get header value.
-
-        @return: header value or default of not found
-        @rtype: str
-        """
-        value = self.headers.get(name)
-        if value is None:
-            return default
-        return value
-
     def check_response(self):
         """Check final result and log it."""
         if (
@@ -314,7 +303,8 @@ class HttpUrl(internpaturl.InternPatternUrl, proxysupport.ProxySupport):
 
             if self.url_connection.status_code == 429:
                 self.add_warning(
-                    "Rate limited (Retry-After: %s)" % self.getheader(_("Retry-After")),
+                    "Rate limited (Retry-After: %s)"
+                    % self.headers.get(_("Retry-After")),
                     tag=WARN_URL_RATE_LIMITED,
                 )
 
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -16,6 +16,9 @@
 """
 Handle http links.
 """
+
+import urllib.parse
+
 import requests
 
 # The validity of SSL certs is ignored to be able
@@ -34,7 +37,6 @@ import re
 from .. import (
     log,
     LOG_CHECK,
-    strformat,
     mimeutil,
     url as urlutil,
     LinkCheckerError,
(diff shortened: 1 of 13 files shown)


cjmayo:
Got a bit more rigorous about this and checked we actually had test coverage, two untested in plugins.locationinfo.get_location() 

https://github.com/linkchecker/linkchecker/blob/2a46b91327e7286edbf5a93d5fa71838033e630f/linkcheck/plugins/locationinfo.py#L114-L118

added a test. Unfortunately geoip-database does not have the city information so only one of these is tested but I think it is reasonable to assume the other would be OK.

linkcheck/plugins/parsepdf.py
```diff
@@ -44,14 +44,7 @@ def search_url(obj, url_data, pageno, seen_objs):
...
             if key == 'URI':
-                if isinstance(value, str):
-                    url = value
-                else:
-                    # URIs should be 7bit ASCII encoded, but be safe and encode
-                    # to unicode
-                    # XXX this does not use an optional specified base URL
-                    url = strformat.unicode_safe(value)
-                url_data.add_url(url, page=pageno)
+                url_data.add_url(str(value, encoding="ascii"), page=pageno)
```
mgedmin, line 47: Whoa, I don't think I've ever seen `str(something, encoding='ascii')` before.  `something.decode('ascii')` is the canonical spelling of this operation.

linkcheck/checker/httpurl.py
```diff
@@ -111,7 +110,7 @@ def add_size_info(self):
             # Note that content-encoding causes size differences since
             # the content data is always decoded.
             try:
-                self.size = int(self.getheader("Content-Length"))
+                self.size = int(self.headers.get("Content-Length"))
             except (ValueError, OverflowError):
```
mgedmin, line 114: int(None) will raise a TypeError.

I don't think you can get an OverflowError from int() in Python 3.

(Preexisting bug, fix, if any, belongs to a separate PR.)

linkcheck/checker/httpurl.py
```diff
@@ -273,7 +272,7 @@ def follow_redirections(self, request):
             # Reset extern and recalculate
             self.extern = None
             self.set_extern(newurl)
-            self.urlparts = strformat.url_unicode_split(newurl)
+            self.urlparts = list(urllib.parse.urlsplit(newurl))
```
mgedmin, line 275: Life would be better if we could retain the named tuple, but IIRC there's code that wants self.urlparts to be a mutable list.

linkcheck/checker/httpurl.py
```diff
@@ -315,7 +303,8 @@ def check_response(self):
 
             if self.url_connection.status_code == 429:
                 self.add_warning(
-                    "Rate limited (Retry-After: %s)" % self.getheader(_("Retry-After")),
+                    "Rate limited (Retry-After: %s)"
+                    % self.headers.get(_("Retry-After")),
```
mgedmin, line 307: Wait, are we translating the name of an HTTP header????  That would be a bug (the fix for which belongs to a separate PR).

linkcheck/checker/internpaturl.py
```diff
@@ -16,15 +16,18 @@
...
 from . import urlbase, absolute_url
-from .. import strformat, url as urlutil
+from .. import url as urlutil
 
 
 def get_intern_pattern(url):
     """Return intern pattern for given URL. Redirections to the same
     domain with or without "www." prepended are allowed."""
-    parts = strformat.url_unicode_split(url)
+    parts = list(urllib.parse.urlsplit(url))
```
mgedmin, line 30: No need for list(), it's a named tuple that supports indexing.

(But for readability we ought to refer to its members by name.  Separate PR material.)

linkcheck/checker/httpurl.py
mgedmin: Ok, not a bug because this is inside an `if "Content-Length" in self.headers:`, but in that case do not use self.headers.get, use `self.headers['Content-Length']`.

linkcheck/plugins/parsepdf.py
cjmayo: Done.

linkcheck/checker/httpurl.py
cjmayo: Done.

linkcheck/checker/httpurl.py
cjmayo: Raised issue #449.

linkcheck/checker/internpaturl.py
cjmayo: Removed list()

cjmayo:
Looks like needs a rebase even for Travis which requires the conflicts to be resolved:

So when you create or update a pull request, and Travis CI doesn’t create a build for it, make sure the pull request is mergeable. If it isn’t, rebase it against the upstream branch and resolve any merge conflicts. When you push the fixes up to the pull request, Travis CI will happily build them.

https://docs.travis-ci.com/user/pull-requests/#my-pull-request-isnt-being-built

cjmayo:
I've rebased to unstick Travis and squashed in the review comments.

PR #444, opened 2020-06-18
Remove or replace uses of isinstance()

cjmayo:
More info in the comments.

The one for FtpUrl looks like a fix.

### DIFF
--- a/linkcheck/checker/fileurl.py
+++ b/linkcheck/checker/fileurl.py
@@ -234,8 +234,7 @@ class FileUrl(urlbase.UrlBase):
         with links to the files."""
         if self.is_directory():
             data = get_index_html(get_files(self.get_os_filename()))
-            if isinstance(data, str):
-                data = data.encode("iso8859-1", "ignore")
+            data = data.encode("iso8859-1", "ignore")
         else:
             data = super().read_content()
         return data
--- a/linkcheck/fileutil.py
+++ b/linkcheck/fileutil.py
@@ -70,7 +70,7 @@ else:
 
 def path_safe(path):
     """Ensure path string is compatible with the platform file system encoding."""
-    if isinstance(path, str) and not os.path.supports_unicode_filenames:
+    if path and not os.path.supports_unicode_filenames:
         path = path.encode(FSCODING, "replace").decode(FSCODING)
     return path
 
--- a/linkcheck/robotparser2.py
+++ b/linkcheck/robotparser2.py
@@ -276,10 +276,6 @@ class RobotFileParser:
             useragent,
             url,
         )
-        if not isinstance(useragent, str):
-            useragent = useragent.encode("ascii", "ignore")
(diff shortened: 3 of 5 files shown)


cjmayo:
New commit because no need to encode/decode an empty string.

```diff
--- a/linkcheck/strformat.py
+++ b/linkcheck/strformat.py
@@ -67,7 +67,7 @@ def ascii_safe(s):
     @return: version of s containing only ASCII characters, or None if s was None
     @rtype: string or None
     """
-    if s is not None:
+    if s:
         s = s.encode('ascii', 'ignore').decode('ascii')
     return s
```

linkcheck/checker/ftpurl.py
```diff
@@ -115,8 +115,9 @@ def cwd(self):
         component.
         """
         path = self.urlparts[2]
-        if isinstance(path, bytes):
-            path = path.decode(self.filename_encoding, 'replace')
+        if self.filename_encoding != "utf-8":
+            path = path.encode(self.filename_encoding, 'replace').decode(
+                self.filename_encoding)
```
mgedmin, line 120: This change doesn't make sense to me.  Can you drop it from this PR please?

(There might be some logic errors in ftpurl's filename handling, but bugfixes shouldn't hide in PRs that purport to be refactorings.)

linkcheck/checker/fileurl.py
```diff
@@ -234,8 +234,7 @@ def read_content(self):
         with links to the files."""
         if self.is_directory():
             data = get_index_html(get_files(self.get_os_filename()))
-            if isinstance(data, str):
-                data = data.encode("iso8859-1", "ignore")
+            data = data.encode("iso8859-1", "ignore")
```
mgedmin, line 237: This is a correct refactoring, but it brings attention to buggy code.  ISO-8859-1 is not the right choice for arbitrary filenames.  UTF-8 might work better.  Ideally we'd plumb things up so that get_content() that calls get_raw_content() that calls download_content() that calls read_content() wouldn't need to have BeautifulSoup guess what the encoding is, given that it's hardcoded right here!

(Maybe get_index_html() ought to generate a `<meta charset="UTF-8">`.)

In case it wasn't clear: this change here is fine, and the bugfix belongs to a separate PR.

linkcheck/checker/ftpurl.py
cjmayo: Dropped that change.

cjmayo: Raised issue #450.

linkcheck/checker/fileurl.py
cjmayo: Raised issue #451.

cjmayo:
Dropped the "Replace isinstance() in FtpUrl.cwd()" and its revert commit. Renamed the squash commit to "No need to process an empty string in str_format.ascii_safe()".

Issue #446, opened 2020-06-19
installation instructions - incorrect for MacOS


dfkotz:
## Steps to reproduce

 1. read https://github.com/linkchecker/linkchecker/blob/master/doc/install.txt
 2. follow instructions for "Manual setup for OSX systems"

## Actual result

 1. note that `pip3 install -r requirements.txt --use-mirrors` will fail because `use-mirrors` is unknown option.
 2. note that `make app` will fail because it forces python2.7 but there are later checks that print `This program requires Python 3.5 or later.`

## Expected result

 1. Makefile would build dmg with LinkChecker.app

## Environment

 * Operating system: `Darwin Kotzbook2019-June-2.local 18.7.0 Darwin Kernel Version 18.7.0: Mon Apr 27 20:09:39 PDT 2020; root:xnu-4903.278.35~1/RELEASE_X86_64 x86_64`
 * Linkchecker version: not successfully installed
 * Python version: `python -V` = Python 2.7.17, `python3 -V` = Python 3.7.7
 * Install method: followed [https://github.com/linkchecker/linkchecker/blob/master/doc/install.txt](instructions from install.txt)
 * Site URL: https://github.com/linkchecker/linkchecker/blob/master/doc/install.txt

## Configuration file

does not exist yet

## Logs

n/a

## Other notes

I suspect your documentation is just out of date.
It may be useful to also indicate whether the user must modify the Makefile to specify a python version.

cjmayo:
Thanks for trying the instructions and reporting. None of the project members use MacOS.

py2app and `make app` support was removed in e3ab902, hence these manual instructions cannot work - I propose they should be removed #448.

PR #448, opened 2020-06-19
Remove manual setup for OSX from install instructions

cjmayo:
py2app support was removed in:
e3ab902 ("Remove platform-specific installer stuff and ensure a build
.whl wheel file can be built.", 2016-01-17)

---

The instructions do not work as reported in #446 and cannot because the support for pyapp and the app Makefile target were removed with wheels as the intended replacement.

Also a commit to build universal wheels only.

### DIFF
--- a/doc/install.txt
+++ b/doc/install.txt
@@ -99,49 +99,6 @@ After installation
 ------------------
 LinkChecker is now installed. Have fun!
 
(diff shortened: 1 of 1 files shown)


cjmayo:
Looks like universal wheels have to work with Python 2 and 3, remove that change.

PR #462, opened 2020-07-27
Fix anchor checking

### LINKED ISSUE #460: AnchorCheck plugin is broken
With current a977e4d7129450ba9fda8389724c80c1bde66883:

No warnings reported when this file is served as fragment.html and checked with `-r1` (or `-r2`):

```html
<html>
<body>
<a href="fragment.html">a</a>
<a href="fragment.html#one">a1</a>
<a href="fragment.html#two">a2</a>
<a href="fragment.html#three">a3</a>
<a href="fragment.html#four">a4</a>
<a href="fragment.html#five">a5</a>
</body>
</html>
```

Commenting out these two lines:

https://github.com/linkchecker/linkchecker/blob/a977e4d7129450ba9fda8389724c80c1bde66883/linkcheck/cache/urlqueue.py#L128-L129

then checking with `-r1` results in `5 warnings found` of the type:
```Warning    [None] Anchor `two' not found. Available anchors: -.```

checking with `-r2` gives `13 warnings found`, with several repeats.

9.3 and 9.4 do not report any warnings either. The commented out lines were introduced in 9.4:

eaa538c8 ("don't check one url multiple times", 2016-11-09)

### PR DESCRIPTION
**Pull request by cjmayo**
Query and fragment URL parts for filesystem URLs are ignored, therefore
test over http.

---

First a new test, let's see it fail.

Closes #460.

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -363,11 +363,13 @@ class UrlBase:
 
     def set_cache_url(self):
         """Set the URL to be used for caching."""
-        # remove anchor from cached target url since we assume
-        # URLs with different anchors to have the same content
-        self.cache_url = urlutil.urlunsplit(self.urlparts[:4] + [''])
-        if self.cache_url is not None:
-            assert isinstance(self.cache_url, str), repr(self.cache_url)
+        if "AnchorCheck" in self.aggregate.config["enabledplugins"]:
+            self.cache_url = self.url
+        else:
+            # remove anchor from cached target url since we assume
+            # URLs with different anchors to have the same content
+            self.cache_url = urlutil.urlunsplit(self.urlparts[:4] + [''])
+        log.debug(LOG_CHECK, "cache_url '%s'", self.cache_url)
 
     def check_syntax(self):
         """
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cjmayo** [comment]
Commit added that makes the test pass. Not completely happy with depending on a plugin being configured in the main code.

Also should close #179.

File: linkcheck/checker/urlbase.py [hunk 1]
```diff
@@ -363,11 +363,13 @@ def add_info(self, s):
 
     def set_cache_url(self):
         """Set the URL to be used for caching."""
-        # remove anchor from cached target url since we assume
-        # URLs with different anchors to have the same content
-        self.cache_url = urlutil.urlunsplit(self.urlparts[:4] + [''])
-        if self.cache_url is not None:
-            assert isinstance(self.cache_url, str), repr(self.cache_url)
+        if "AnchorCheck" in self.aggregate.config["enabledplugins"]:
+            self.cache_url = self.url
```
mgedmin, line 367: Ugh.  This kind of check feels wrong, but having AnchorCheck not work at all feels even wrong-er.

mgedmin (approved):
LGTM

cjmayo:
Merged, but raised #500 to note this "fix".

PR #490, opened 2020-08-22
Fix deprecation warning for resolver.query(), Drop Python 3.5 and add 3.9

cjmayo:
```
/home/travis/build/linkchecker/linkchecker/linkcheck/checker/mailtourl.py:321: DeprecationWarning: please use dns.resolver.resolve() instead
    answers = resolver.query(domain, 'MX')
```


cjmayo:
Updated with documentation and version checks and ready to go.

Fixing the dnspython warning did mean upping the minimum version to 2, but as the previous version was released in 2018 I guess that is not so bad.

linkcheck/bookmarks/safari.py
```diff
@@ -23,6 +25,9 @@
 except ImportError:
     has_biplist = False
 
+if sys.version_info >= (3, 9):
+    has_biplist = False
```
mgedmin, line 29: This just cries for a comment explaining why we're setting `has_biplist = False` even if `import biplist` succeeded on Python 3.9+.

tests/__init__.py
```diff
@@ -173,6 +173,9 @@ def has_pyftpdlib():
 @lru_cache(1)
 def has_biplist():
     """Test if biplist is available."""
+    if sys.version_info >= (3, 9):
+        return False
+
```
mgedmin, line 178: Why not import `has_biplist` from `linkcheck.bookmarks.safari` instead of duplicating the logic?

tests/checker/test_http.py
```diff
@@ -60,7 +60,7 @@ def _test_status(self, status):
             resultlines.append("warning No Content")
         if status == 429:
             resultlines.append("warning Rate limited (Retry-After: None)")
-        if (status not in [101, 102] and status < 200) or (
+        if (status not in [101, 102, 103] and status < 200) or (
```
mgedmin, line 63: How is this related to dnspython API deprecation, and what is HTTP status code 103?

mgedmin (approved):
LGTM, with some questions?

linkcheck/bookmarks/safari.py
mgedmin: I mean one can dig out that information from git commit messages, but.

Also, that comment should include a link to a biplist bug report about the Python 3.9 compatibility problem, so we can click it and go see when it gets fixed and remove the override in the future.

tests/checker/test_http.py
mgedmin: Once again the git commit message answers my question!

I wish github made it more natural to do reviews commit-by-commit rather than in one lump.

linkcheck/bookmarks/safari.py
anarcat: in general, what i sometimes do here is:

```suggestion
try:
    import biplist
except ImportError:
    biplist = None
```

that avoids the extra variable. but that's somewhat unrelated. :)

tests/checker/test_http.py
anarcat: i would rather the code answer the question than having to dig into git history. both places is best.

linkcheck/bookmarks/safari.py
cjmayo: Filed a bug and added a link in our code.

tests/__init__.py
cjmayo: Reused the linkcheck.bookmarks.safari function.

PR #495, opened 2020-08-27
MS Windows Python 3.7 and MS Store compatibility

cjmayo:
os.path.splitunc() removed in Python 3.7.

https://docs.python.org/3/whatsnew/3.7.html#api-and-feature-removals


mgedmin (approved):
lgtm

PR #496, opened 2020-08-29
Use Python:3-slim base image in Dockerfile

cjmayo:
Currently reduces image size from 896MB to 126MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3-slim
 
 # needed to allow linkchecker create plugin directory and initial configuration file in "home" dir
 ENV HOME /tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,12 +4,7 @@ FROM python:3-slim
 ENV HOME /tmp
 
 RUN set -x \
-    && apt-get update -qq \
-    && apt-get install -y -qq --no-install-recommends git \
-    && pip install --no-cache-dir git+https://github.com/linkchecker/linkchecker.git@master \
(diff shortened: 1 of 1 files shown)


cjmayo:
Use https://github.com/linkchecker/linkchecker/archive/master.zip instead of git. Ubuntu git requires a number of dependencies to be installed.

Dockerfile
```diff
@@ -1,15 +1,10 @@
...
 
 RUN set -x \
-    && apt-get update -qq \
-    && apt-get install -y -qq --no-install-recommends git \
-    && pip install --no-cache-dir git+https://github.com/linkchecker/linkchecker.git@master \
-    && apt-get -y -qq purge git \
-    && apt-get autoremove -y -qq \
-    && apt-get clean -y -qq all \
+    && pip install --no-cache-dir https://github.com/linkchecker/linkchecker/archive/master.zip \
     && rm -rf /var/lib/apt/lists/* /usr/share/doc /usr/share/locale /usr/share/man
```
mgedmin, line 8: The `rm -rf` is here to reduce the size of the docker image by removing things installed by apt-get in the same `RUN` command.

The `python:3-slim` base image has nothing in /var/lib/apt, /usr/share/locale, /usr/share/man, and removing the 1 megabyte of files in /usr/share/doc will only increase the size of our image due to the way Docker layers work.

mgedmin (approved):
LGTM

Dockerfile
cjmayo: rm removed.

anarcat (commented):
:+1:

PR #494, opened 2020-08-27
Use GitHub Actions to test on MS Windows

cjmayo:
Includes adapted file tests, partially from #127 by Petr Dlouhý.

If anyone using Windows wants to help:

* A couple of tests fail, the reasons need investigating
* Testing takes too long, some tests appear to be waiting for something

PR #497, opened 2020-08-30
Use GitHub Actions to test on MS Windows

cjmayo:
* A couple of tests fail, the reasons need investigating
* Testing takes too long, some tests appear to be waiting for something


cjmayo:
Duplicate of #494 because Actions weren't working, now they are...

PR #498, opened 2020-08-30
Tidy linkchecker

cjmayo:
Tidy intro, update http links and replace from linkcheck.

### DIFF
--- a/linkcheck/logger/text.py
+++ b/linkcheck/logger/text.py
@@ -93,7 +93,8 @@ class TextLogger(_Logger):
 
     def write_intro(self):
         """Log introduction text."""
-        self.writeln(configuration.AppInfo)
+        self.writeln(configuration.App)
+        self.writeln(configuration.Copyright)
         self.writeln(configuration.Freeware)
         self.writeln(
             _("Get the newest version at %(url)s") % {'url': configuration.Url}
--- a/linkchecker
+++ b/linkchecker
@@ -109,7 +109,7 @@ Setting a proxy on the Windows command prompt:
 RegularExpressions = _(
     """REGULAR EXPRESSIONS
 Only Python regular expressions are accepted by LinkChecker.
-See http://www.amk.ca/python/howto/regex/ for an introduction in
+See https://docs.python.org/howto/regex.html for an introduction in
 regular expressions.
 
 The only addition is that a leading exclamation mark negates
@@ -325,7 +325,7 @@ group.add_argument(
         "The ENCODING specifies the output encoding, the default is that of your\n"
         "locale.\n"
         "Valid encodings are listed at "
-        "http://docs.python.org/lib/standard-encodings.html.\n"
+        "https://docs.python.org/library/codecs.html#standard-encodings.\n"
         "The FILENAME and ENCODING parts of the 'none' output type will be ignored,\n"
         "else if the file already exists, it will be overwritten.\n"
         "You can specify this option more than once. Valid file output types\n"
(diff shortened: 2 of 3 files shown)


cjmayo:
@mgedmin if you have a chance to review, thanks.

linkcheck/logger/text.py
```diff
@@ -93,7 +93,8 @@ def start_output(self):
 
     def write_intro(self):
         """Log introduction text."""
-        self.writeln(configuration.AppInfo)
+        self.writeln(configuration.App)
+        self.writeln(configuration.Copyright)
```
mgedmin, line 97: I don't entirely understand the purpose of this change?

linkchecker
```diff
@@ -42,13 +43,13 @@ from linkcheck.cmdline import (
     LCArgumentParser,
     print_plugins,
 )
-from linkcheck import log, i18n, strformat
```
mgedmin, line 45: This is slightly iffy: if `linkcheck.log` is a module, then `import linkcheck` doesn't guarantee that `linkcheck.log` will be imported and available, unless `linkcheck/__init__.py` contains an `import .log` somewhere.  (Which it does, but will that always be the case?  Probably.)

As usual I'm splitting hairs for no good reason.  It's fine.

linkchecker
```diff
@@ -578,7 +579,7 @@ if options.output:
     if "/" in options.output:
         logtype, encoding = options.output.split("/", 1)
     else:
-        logtype, encoding = options.output, i18n.default_encoding
+        logtype, encoding = options.output, linkcheck.i18n.default_encoding
```
mgedmin, line 582: Same here: since we're not explicitly importing `linkcheck.i18n` here any longer, we now rely on `linkcheck/__init__.py` having an `import .i18n`.

linkchecker
```diff
@@ -109,7 +110,7 @@ Setting a proxy on the Windows command prompt:
 RegularExpressions = _(
     """REGULAR EXPRESSIONS
 Only Python regular expressions are accepted by LinkChecker.
-See http://www.amk.ca/python/howto/regex/ for an introduction in
+See https://docs.python.org/howto/regex.html for an introduction in
```
mgedmin, line 113: :100:

linkchecker
```diff
@@ -325,7 +326,7 @@ group.add_argument(
         "The ENCODING specifies the output encoding, the default is that of your\n"
         "locale.\n"
         "Valid encodings are listed at "
-        "http://docs.python.org/lib/standard-encodings.html.\n"
+        "https://docs.python.org/library/codecs.html#standard-encodings.\n"
```
mgedmin, line 329: :100:

linkchecker
```diff
@@ -358,7 +359,7 @@ group.add_argument(
         "The ENCODING specifies the output encoding, the default is that of your\n"
         "locale.\n"
         "Valid encodings are listed at "
-        "http://docs.python.org/lib/standard-encodings.html."
+        "https://docs.python.org/library/codecs.html#standard-encodings."
```
mgedmin, line 362: :100:

mgedmin (approved):
The URL updates are excellent and very welcome, the rest of the changes I'm ambivalent about, but I won't stand in your way.

linkcheck/logger/text.py
cjmayo: The copyright line has got rather long. (I saved a line elsewhere, so we are even).

mgedmin: Ah, makes sense!

anarcat (approved):

linkchecker
cjmayo: Keep splitting! Rewritten to use `from linkcheck import log, logconf` and `from linkcheck.i18n import default_encoding`

cjmayo: Don't think `from linkcheck.i18n import default_encoding` is correct. default_encoding is set as a global in i18n.init() which is called from linkcheck/__init__.py...

linkchecker
cjmayo: As we are relying on `__init__.py` to call `i18n.init()` I think this is OK.

PR #501, opened 2020-09-01
Document and test Sphinx documentation

cjmayo:
`make -C doc check` to check the man pages was already present, I switched it to use mandoc - I could have documented why better, man-db was outputting warnings which it wasn't possible to do anything about. Maybe the whole idea of the check is a bit redundant now that they are being generated for us, but it doesn't cost much.

I put the whole check in Travis, as well as mandoc it does use the graphviz deb.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: bionic
 language: python
 python:
   - "3.5"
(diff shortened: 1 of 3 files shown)


mgedmin (approved):
lgtm

anarcat (commented):
:+1: :)

PR #503, opened 2020-09-14
Document MarkdownCheck plugin

cjmayo:
Closes #502.

PR #507, opened 2020-09-23
Add MarkdownCheck plugin to linkcheckerrc(5) on the website

cjmayo:
Also code documentation is updated for the changes from:

0912e8a2 ("Don't strip the URL fragment from cache key if using AnchorCheck", 2020-07-27)
2fbd49dd ("Replace os.path.splitunc() with os.path.splitdrive()", 2020-08-29)

---

This is meant to be just publishing the MarkdownCheck documentation from #503, of course it ends up quite big... Maybe using the date as the Sphinx version wasn't that great an idea, although here the navigation is also updated so wouldn't have saved changed files.


mgedmin (approved):
Oh wow the diff is brutal.

I'm sure it would LGTM if I could bear looking at the diff ;)

PR #509, opened 2020-09-25
Update man pages because of multiline hyperlink

cjmayo:
Hyperlinks were added to the linkcheckerrc.rst for use in the HTML output back in:
67bdc54a ("Add links to MULTILINE section in linkcheckerrc(5)", 2020-08-23)

This just catches the man pages up, which was complicated by the need to update the translations.


mgedmin (approved):
lgtm, after a cursory glance

PR #512, opened 2020-10-02
Correct documentation of aborttimeout, no matching option

cjmayo:
```
$ grep -rn aborttimeout linkcheck
linkcheck/configuration/confparse.py:167:        self.read_int_option(section, "aborttimeout", min=1)
linkcheck/configuration/__init__.py:184:        self["aborttimeout"] = 300
linkcheck/director/aggregator.py:173:            timeout=strformat.strduration_long(self.config["aborttimeout"]),
linkcheck/director/aggregator.py:200:        timeout = self.config["aborttimeout"]

config["aborttimeout"] is only set from aborttimeout and is the only
value used by Aggregate.abort().
```
---

First some missed multiline hyperlinks in German linkcheckerrc(5) from previous commit.

### DIFF
--- a/doc/man/de/linkcheckerrc.5
+++ b/doc/man/de/linkcheckerrc.5
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH "LINKCHECKERRC" "5" "September 14, 2020" "2020-09-14" "LinkChecker"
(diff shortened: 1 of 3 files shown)


mgedmin (approved):
LGTM, I suppose.

(maybe we should fix the code to match documentation?  but I've no energy for even thinking about it)

PR #513, opened 2020-10-02
Update GitHub Pages aborttimeout has no option

cjmayo:
The GitHub pages update for #512.

I thought I had a strategy for separating out the date updates into a separate commit, but then found there were existing unpublished changes relating to the dropping of Python 3.5, support for Python 3.9 and changes to the intro.


mgedmin (approved):
another huge diff that LGTM if I just close my eyes and not look.

cjmayo:
Yes, I might have a look at automating these (not just yet though).

PR #515, opened 2020-10-12
Automatically publish documentation on GitHub Pages

cjmayo:
To save the separate PRs for gh-pages with massive diffs.

Uses https://github.com/peaceiris/actions-gh-pages to do the publishing.

### DIFF
--- a/.github/workflows/publish-pages.yml
+++ b/.github/workflows/publish-pages.yml
@@ -1,0 +1,39 @@
+name: Publish LinkChecker Documentation on GitHub Pages
+
+on:
+  push:
+    branches: ["master"]
+
+jobs:
+  run:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Ubuntu packages
+        run: sudo apt install graphviz
+
+      # Ensure sphinx-autogen is installed in PATH
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - name: Install Python packages
+        run: >
+            pip install dnspython beautifulsoup4 pyxdg requests \
+                        sphinx sphinx_epytext sphinx_rtd_theme
+
+      - name: Build
+        run: |
+            python3 setup.py build
+            make -C doc code
+            make -C doc html
+
+      - name: Publish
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-pages.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish LinkChecker Documentation on GitHub Pages
+
+on:
+    push:
+      branches: ["master"]
+
+jobs:
+  run:
```
mgedmin, line 8: The indentation seems inconsistent, but who's cares?

(Me.  I care.  I"m cursed with care for irrelevant detail, while letting important things slide.  Also, I have yamllint integrated with my text editor of choice.  2-space indents for YAML files look nice.)

mgedmin (approved):
:+1:

.github/workflows/publish-pages.yml
cjmayo: All good, another top tip. yamllint errors and warnings (except document start and truthy for on:) fixed.
Also updated documentation.md for the automation.

cjmayo:
peaceiris/actions-gh-pages@v3 created the .nojekyll in the root directory, which stops the CSS being published on the web...

For now I pushed a manual commit to gh-pages, I'll find a fix for the action.

mgedmin:
Can you elaborate how the existence of `.nojekyll` stops CSS from being published?

I've used github pages, but not github actions.

.github/workflows/publish-pages.yml
mgedmin: It's beautiful now.

:sheds tear:

mgedmin:
From https://github.com/linkchecker/linkchecker/commit/79e34c8554388e74af26b6bcb339e5ea166f7712 I think I misunderstood, and the problem was not the existence of `.nojekyll`, but its location.

Interesting.  Why are we keeping the docs in a subdirectory in the gh-pages branch?

Is this a bug in the actions-gh-pages action, or some misconfiguration on our end?

cjmayo:
It was deliberate, when creating the pages manually in the same repository it made it easier, e.g. `rm -rf docs`. But even then I often ended up using two repositories for this. Let's go all in on automation and move to root.

PR #516, opened 2020-10-12
Stop using biplist

cjmayo:
plistlib has supported binary files since Python 3.4.

---

Didn't spot this myself first time, but fortunately a comment was added to the Bitbucket bug.

 `test_safari_bookmarks_binary` works with plistlib.


mgedmin (approved):
Excellent!

PR #517, opened 2020-10-14
Publish documentation to the root of gh-pages

cjmayo:
This gives the Action full control of the branch, it will wipe out anything else e.g. the .gitignore (that isn't needed though if not building in the same repository).

Before merging this I will push a manual commit to gh-pages moving the existing files to root, and point the GitHub pages to the root directory in the project configuration.

### DIFF
--- a/doc/documentation.md
+++ b/doc/documentation.md
@@ -55,18 +55,3 @@ The Web Site is hosted by GitHub Pages from the docs/ directory of the gh-pages 
 
 When updates to LinkChecker are pushed, the web site is built and published
 automatically by a GitHub action ``.github/workflows/publish-pages.yml``.
-
-For information, a manual process to build and publish the web site would look like:
-
-    git checkout master
-
-    ./setup.py build  # for copyright, author and version info
-    make -C doc code
-    make -C doc html
-
-    git checkout -b <branch> gh-pages
-
-    rm -rf docs/*
-    cp -a doc/html/* docs/
-
-    git commit -a -m "Update documentation"
--- a/doc/documentation.md
+++ b/doc/documentation.md
@@ -14,6 +14,13 @@ sphinx_epytext
 
 sphinx_rtd_theme
 
+Configuration
+-------------
+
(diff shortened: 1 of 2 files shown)


doc/documentation.md
```diff
@@ -55,18 +55,3 @@ The Web Site is hosted by GitHub Pages from the docs/ directory of the gh-pages
 
 When updates to LinkChecker are pushed, the web site is built and published
 automatically by a GitHub action ``.github/workflows/publish-pages.yml``.
-
-For information, a manual process to build and publish the web site would look like:
```
anarcat, line 59: what would be the manual process now? is it documented somewhere? if not, why shouldn't it be? :)

anarcat (changes requested):

doc/documentation.md
cjmayo: I haven't worked it out! Well, I think it needs 2 repositories but I don't have the exact steps.

The section above does document how to create the HTML, maybe `setup.py build` should be added.

anarcat: well maybe it's fine to remove the outdated docs in the meantime, i retract my objection but it would be nice to see how this works outside of CI.

anarcat (approved):

doc/documentation.md
mgedmin: By "2 repositories" do you mean two checkouts of the same repository?

I've used [ghp-import](https://pypi.org/project/ghp-import/) in the past, it works great.  No extra checkouts needed, just a docs build in a local .gitignore'd directory that gets imported into the root of the gh-pages branch by the ghp-import script.

But I don't think we'll need that if we have the GH action.

mgedmin (approved):

doc/documentation.md
cjmayo: I have added the setup.py build step, in its own section because it is needed for man pages too.
That does mean our steps are recorded, even if there isn't a recipe for a manual publish. At the moment any merge will break the web site so best get this done.

Yes, two local ones. ghp-import looks interesting, although with the action there is no need to do anything, as long as it works...

PR #521, opened 2020-11-11
Unhandled LookupError in robotparser2.py

bertrama:
Sites presenting an invalid content-type character set raise a LookupError exception.

For an example, see the single link on [this page I've prepared](http://www-personal.umich.edu/~bertrama/linkchecker/index.html).

### DIFF
--- a/linkcheck/robotparser2.py
+++ b/linkcheck/robotparser2.py
@@ -123,6 +123,9 @@ class RobotFileParser:
             # no network or other failure
             self.allow_all = True
             log.debug(LOG_CHECK, "%r allow all (request error)", self.url)
+        except LookupError:
+            self.allow_all = True
+            log.debug(LOG_CHECK, "%r allow all (lookup error)", self.url)
 
     def _add_entry(self, entry):
         """Add a parsed entry to entry list.


cjmayo:
Yes, but even though it is strictly the server that is incorrect I would still be wary of LinkChecker just ignoring the robots.txt if we can help it. (that is the impact of `self.allow_all = True` I guess).

If this works:
```diff
--- a/linkcheck/robotparser2.py
+++ b/linkcheck/robotparser2.py
@@ -99,6 +99,7 @@ class RobotFileParser:
             response = self.session.get(self.url, **kwargs)
             response.raise_for_status()
             content_type = response.headers.get('content-type')
+            response.encoding = response.apparent_encoding
             self.encoding = response.encoding
             if content_type and content_type.lower().startswith('text/plain'):
                 self.parse(response.iter_lines(decode_unicode=True))
```
Then I think the actual fix should be something that tries the response charset and falls back to apparent_encoding. Maybe replacing .iter_lines() with somethig using .text (which may even do the fallback itself).

cjmayo:
Turns out the standard library version urllib.robotparser assumes UTF-8 and so do Google:
https://developers.google.com/search/docs/advanced/robots/robots_txt#file-format

Let's do that #591.

Thanks for the report and the suggestion. If 591 doesn't fix it please say.

PR #524, opened 2021-01-08
Add Dutch translation

gvmelle:
I made a Dutch translation


cjmayo:
Translations are welcome - I just wanted to stop including the binary catalog files in the source (could have made that clearer sooner). But now I finally have a proposed solution #563. Reviews welcome. After that is merged I **will** look at this one.

cjmayo:
I created a new PR #574 so that I could remove `nl_NL.mo` from the commit. Do check.
Thanks a lot!

cjmayo:
I did slightly tweak the headers including an attribution

PR #528, opened 2021-01-19
Add h1-h4 as AnchorTags

NigelThomas:


linkcheck/htmlutil/linkparse.py
```diff
@@ -65,6 +65,10 @@
 # HTML anchor tags
 AnchorTags = {
     'a': ['name'],
+    'h1': ['id'],
+    'h2': ['id'],
+    'h3': ['id'],
+    'h4': ['id'],
     None: ['id'],
```
mgedmin, line 72: AFAIU the `None: ['id']` means that `id` attributes on all tags (including h1-h4) be treated as anchor names, so this PR does not have any effect.

Am I mistaken?

cjmayo: That is my understanding -  a bit confusing but `None` is actually acting as wild card.

NigelThomas:
Sorry, you are right of course; the pull request was very premature. I am investigating why implicit anchors of the form <h1 id="slug">text</h1> seem get called out as missing anchors, and don't get included in the list of available anchors. If I find anything I'll raise a sensible PR. Closing this.

PR #536, opened 2021-04-15
Better handling for link rel dns-prefetch and preconnect

vdwijngaert:
I changed a few things in regards to handling link tags, in particular `dns-prefetch` and `preconnect`.

* `rel` attribute can have multiple values, separated by a space, so switched to substring comparison
* `preconnect` usually uses the root URL (eg `https://fonts.googleapis.com`) for setting up a connection, but that URL can return a 404 in some cases, resulting in a false positive. I changed it to a DNS check for now. This could be changed in the future to use a check for a valid http connection.

I added some tests, too :)

### DIFF
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -157,7 +157,8 @@ class LinkFinder:
                 base = self.base_ref
             # note: value can be None
             value = attrs.get(attr)
-            if tag == 'link' and ('dns-prefetch' in attrs.get('rel') or 'preconnect' in attrs.get('rel')):
+            rel = attrs.get('rel', '').lower()
+            if tag == 'link' and ('dns-prefetch' in rel or 'preconnect' in rel):
                 if ':' in value:
                     value = value.split(':', 1)[1]
                 value = 'dns:' + value.rstrip('/')
--- a/linkcheck/htmlutil/linkparse.py
+++ b/linkcheck/htmlutil/linkparse.py
@@ -157,10 +157,10 @@ class LinkFinder:
                 base = self.base_ref
             # note: value can be None
             value = attrs.get(attr)
-            if tag == 'link' and attrs.get('rel') == 'dns-prefetch':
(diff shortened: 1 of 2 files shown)


linkcheck/htmlutil/linkparse.py
```diff
@@ -157,7 +157,7 @@ def html_element(self, tag, attrs, element_text, lineno, column):
                 base = self.base_ref
             # note: value can be None
             value = attrs.get(attr)
-            if tag == 'link' and attrs.get('rel') == 'dns-prefetch':
+            if tag == 'link' and ('dns-prefetch' in attrs.get('rel') or 'preconnect' in attrs.get('rel')):
```
mgedmin, line 160: What does `attrs.get('rel')` return when there's no `rel` attr?  If `None`, the proposed code will crash with a TypeError.

Ideally there'd be a unit test with a `<link>` without `rel`, to see that this works.

mgedmin:
Looks like we ran out of Travis CI credits and need to migrate to something else (GitHub Actions) before we can see whether tests pass on pull requests.  *sigh* more work...

mgedmin:
We now have working GitHub actions.  Let's see what happens if I push the 'Update branch' button GitHub shoves at me...

linkcheck/htmlutil/linkparse.py
mgedmin: Also, welp, flake8 complains:

```
./linkcheck/htmlutil/linkparse.py:160:89: E501 line too long (106 > 88 characters)
```

vdwijngaert: You're absolutely right. I added a fallback for this, and a test that makes sure no TypeError is raised (there's probably a better way, but it does the job).

cjmayo (approved):
Checking DNS only is allowed even in the Resource Hints Editor's Draft
https://w3c.github.io/resource-hints/#preconnect

cjmayo:
Thanks for the improvement and especially the tests.

vdwijngaert:
You are most welcome! :) Glad to see this merged.

Issue #146, opened 2018-04-13
Feature Request - disable "intro"


peterhoeg:
linkchecker writes a lot of junk to stdout when launching:

```sh
./linkchecker https://example.com
INFO linkcheck.cmdline 2018-04-13 15:57:51,210 MainThread Checking intern URLs only; use --check-extern to check extern URLs.
LinkChecker 9.4.0              Copyright (C) 2000-2014 Bastian Kleineidam
LinkChecker comes with ABSOLUTELY NO WARRANTY!
This is free software, and you are welcome to redistribute it
under certain conditions. Look at the file `LICENSE' within this
distribution.
Get the newest version at http://wummel.github.io/linkchecker/
Write comments and bugs to https://github.com/wummel/linkchecker/issues
Support this project at http://wummel.github.io/linkchecker/donations.html
```

Apart from the fact that the links are wrong, it would be great to be able to pass a flag that makes linkchecker not show the license/version/copyright/links.

anarcat:
the broken links issue is #110.

as fort he verbosity, i think it would be fine to move that cruft to --help or --version flags.

thezoggy:
any plans to add this?

cjmayo:
There is already the "parts" setting in linkcheckerrc i.e. don't include the "intro" logger part:

https://linkchecker.github.io/linkchecker/man/linkcheckerrc.html#text

PR #561, opened 2021-10-29
Add no-intro option (Issue #146)

agxies:
I added the functionality to not show the intro message when the program is started (#146).
Can be activated via `--no-intro` flag or the config file.

### DIFF
--- a/linkchecker
+++ b/linkchecker
@@ -60,8 +60,6 @@ _profile = "linkchecker.prof"
 _username = None
 _password = None
 
-linkcheck.configuration.initialize()
-
 # usage texts
 Notes = _(
     """NOTES
@@ -678,7 +676,10 @@ if options.profile:
 if options.quiet:
     config["logger"] = config.logger_new("none")
 if options.no_intro:
-    linkcheck.configuration.noIntro = True
+    if not config["text"] or config["text"]["parts"][0] == "all":
+        config["text"]["parts"] = ["realurl", "result", "extern", "base", "name", "parenturl", "info", "warning", "dltime", "checktime", "url", "outro", "stats"]
+    else:
+        config["text"]["parts"].remove("intro")       
 if options.recursionlevel is not None:
     config["recursionlevel"] = options.recursionlevel
 if options.status is not None:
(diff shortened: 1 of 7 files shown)


cjmayo:
The concept of intro as an item, and something that is configurable as one of "parts" setting in linkcheckerrc, already exists in the code. A `--no-intro` that still runs write_intro() doesn't seem very clean to me. And having two similar settings in linkcheckerrc even less desirable. Maybe having just a command line `--no-intro` to remove intro from the parts argument before creating the logger would be OK. Not sure how cleanly that could be done.

agxies:
Sorry, I totally overlooked this option in the config file. I now have changed the implementation so that the command line option uses the parts setting. I'm just not quite sure if it is clean enough like that.

linkchecker
```diff
@@ -669,6 +675,11 @@ if options.profile:
     do_profile = options.profile
 if options.quiet:
     config["logger"] = config.logger_new("none")
+if options.no_intro:
+    if not config["text"] or config["text"]["parts"][0] == "all":
+        config["text"]["parts"] = ["realurl", "result", "extern", "base", "name", "parenturl", "info", "warning", "dltime", "checktime", "url", "outro", "stats"]
```
cjmayo, line 680: I can see why you did this but hardcoding the list of parts seems too high a price to pay.

cjmayo:
I can't see a way to do this without adding complexity. Because the current assumption is everything is enabled unless there is an explicit list of parts.

I didn't realise that configuration option was there either before this - so I learnt something!  Probably someone who doesn't want the intro would be happy to set the option and leave it.

agxies:
Learning something is always great :)
Another option may be to change the default behavior. So that by selecting 'all', the parts object will automatically get the list of all parts, but that won't solve the complexity issue either.

From my perspective the complete implementation of the parts function feels somehow inconvenient. Because to deactivate one of the parts, it is mandatory to add all others in the config file. Maybe another approach with simple boolean variables would be sensible, so that every feature can be enabled/disabled individually. This would make it also much easier to have flags for the command line. Don't know what your opinion would be on that?

cjmayo:
Closing because I can't imagine this can be done. Maybe I lack imagination...

Issue #533, opened 2021-03-10
Encoding strings doesn't work


alfador1:
## Summary
Text not encoded correctly.
I have set the locales correctly:
/opt/linkchecker # locale -a
C
C.UTF-8
nb_NO.UTF-8
en_GB.UTF-8
en_US.UTF-8
es_ES.UTF-8
it_IT.UTF-8
ru_RU.UTF-8
fr_FR.UTF-8
ch_DE.UTF-8
pt_BR.UTF-8
sv_SE.UTF-8
de_DE.UTF-8
nl_NL.UTF-8
de_CH.UTF-8

I keep receiving malformed text like:

[Bes√∂k oss p√•
 
Beg√§r   en offert     f√∂r DHL Express
Klicka   h√§r f√∂r att g√• till Freight     DHL Active Tracing

## Steps to reproduce

 1.  linkchecker --no-robots --output=csv/utf_8 --check-extern https://www.dhl.com/se-sv/home.html | tee Test.csv
 2. Check the CSV and see that the name column contains malformed names.

## Expected result
Properly encoded text
## Environment

 * Operating system: 
 * Linkchecker version: 
 * Python version: 3.9.1
 * Install method: from source

cjmayo:
I've identified a LinkChecker problem and a solution in #572 for http(s) links. Testing welcome. I think that is all LinkChecker can do - it may not be sufficient.

Even without that commit, checking the page as a local file from linkchecker (i.e. not via http) can help debug your system. If it still doesn't get the encoding right, do you have chardet installed? Which version?

Then one further option:
- install [cchardet](https://pypi.org/project/cchardet/) #571

lpirl:
Using LinkChecker 10.0.1 on Python 3.9 here, no matter if `cchardet` installed or not, no matter if directory or via http ([e.g., hosted here](https://static.lukas-pirl.de/linkchecker-utf8/)), I cannot get rid of false positives checking the following pages:

`index.html`:
``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>linkchecker utf-8 test, index page</title>
</head>
<body>
  index page,
  <a href="ö.html">link to ö page</a>,
  <a href="🍞.html">link to 🍞 page</a>
</body>
</html>
```

`ö.html`:
``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>linkchecker utf-8 test, ö page</title>
[...]

PR #572, opened 2021-11-22
Ignore an encoding of ISO-8859-1 returned by Requests

cjmayo:
ISO-8859-1 is a fallback for Requests and causes us to mangle UTF-8
content.

Requests' utils.py:

```py
def get_encoding_from_headers(headers):
    """Returns encodings from given HTTP Header Dict.

    :param headers: dictionary to extract encoding from.
    :rtype: str
    """

    content_type = headers.get('content-type')

    if not content_type:
        return None

    content_type, params = _parse_content_type_header(content_type)

    if 'charset' in params:
        return params['charset'].strip("'\"")

    if 'text' in content_type:
        return 'ISO-8859-1'

    if 'application/json' in content_type:
        # Assume UTF-8 based on RFC 4627: https://www.ietf.org/rfc/rfc4627.txt since the charset was unset
        return 'utf-8'
```

### DIFF
--- a/linkcheck/checker/httpurl.py
+++ b/linkcheck/checker/httpurl.py
@@ -177,7 +177,14 @@ class HttpUrl(internpaturl.InternPatternUrl, proxysupport.ProxySupport):
         log.debug(LOG_CHECK, "Request headers %s", request.headers)
         self.url_connection = self.session.send(request, **kwargs)
         self.headers = self.url_connection.headers
-        self.encoding = self.url_connection.encoding
+        log.debug(LOG_CHECK, "Response headers %s", self.headers)
+        if self.url_connection.encoding == "ISO-8859-1":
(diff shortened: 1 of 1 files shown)


matkoniecz:
Maybe it would be a good idea to report it as a bug/feature request to Requests?

Or is it not worth doing?

cjmayo:
Don't think there is anything they can do about it. They are just reporting what they were told, and otherwise picking a safe (i.e. won't cause an Exception) default as we do.

lpirl:
Works excellent for me (see https://github.com/linkchecker/linkchecker/issues/533#issuecomment-979871071 for what didn't work before).

cjmayo:
> Works excellent for me (see [#533 (comment)](https://github.com/linkchecker/linkchecker/issues/533#issuecomment-979871071) for what didn't work before).

Thanks for testing. Not much to say when it works as expected, I'll write something on 533).

PR #576, opened 2021-11-25
Release new Docker image hosted on GitHub Packages

cjmayo:
Ready to publicise now I think, but reviews welcome.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -18,9 +18,6 @@ jobs:
       packages: write
 
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
       - name: Log in to the Container registry
         uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
         with:
@@ -37,7 +34,6 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@a66e35b9cbcf4ad0ea91ffcaf7bbad63ad9e0229
         with:
-          context: .
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -5,6 +5,7 @@ name: Create and publish a Docker image
 on:
   push:
     branches: ['master']
+    tags: 'v*'
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-docker.yml
```diff
@@ -37,7 +34,6 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@a66e35b9cbcf4ad0ea91ffcaf7bbad63ad9e0229
```
ruzickap, line 35: I believe `docker/login-action@v1` should be enough here...

.github/workflows/publish-docker.yml
```diff
@@ -37,7 +34,6 @@ jobs:
...
         with:
-          context: .
           push: true
           tags: ${{ steps.meta.outputs.tags }}
```
ruzickap, line 38: You should tag the docker images according the `git tags` like it is for example here: https://github.com/fluxcd/helm-controller/blob/main/.github/workflows/release.yml

README.rst
```diff
@@ -47,14 +47,12 @@ For other options see ``linkchecker --help``.
...
 -------------
 
-*The Docker images are out-of-date, pip installation is the only currently recommended method.*
-
 If you do not want to install any additional libraries/dependencies you can use the Docker image.
 
 Example for external web site check::
 
-  docker run --rm -it -u $(id -u):$(id -g) linkchecker/linkchecker --verbose https://www.example.com
+  docker run --rm -it -u $(id -u):$(id -g) ghcr.io/linkchecker/linkchecker:master --verbose https://www.example.com
```
ruzickap, line 54: Using `master` or `latest` is not a good practice.
If you do the proper tagging of container images then you should use something like:
`ghcr.io/linkchecker/linkchecker:10.0.1` (or just `linkchecker:10`)

.github/workflows/publish-docker.yml
cjmayo: No build-push-action? N.B. in edc974e7 I have now added configuration to docker/metadata-action.

.github/workflows/publish-docker.yml
cjmayo: See next.

README.rst
cjmayo: Well there are multiple use cases and different sorts of users.

For the rolling image we could use 10, but it would be a bit of an illusion, 9 came out in 2014. Updated in edc974e, to latest provided by docker/metadata-action.

But fortunately we can have multiple tags so some choice now after that commit - semvar tags for application releases and git commit checksums as well.

Also updated the version the application reports to include the git commit checksum if post a release.

cjmayo:
Merging now because I believe it has benefit to users and for resolving issues (especially with PRs about to be merged).

PR #589, opened 2021-12-10
Remove linkcheck.checker.proxysupport

### LINKED ISSUE #152: HttpUrl.proxy_type vs HttpUrl.proxytype
I think there's a bug in [RobotsTxt._allows_url()](https://github.com/linkchecker/linkchecker/blob/268bf162c7bdead21f78c472e4c2635533df5cb9/linkcheck/cache/robots_txt.py#L61):

```python
        if hasattr(url_data, "proxy") and hasattr(url_data, "proxy_type"):
            kwargs["proxies"] = {url_data.proxytype: url_data.proxy}
```

The hasattr() check looks for an attribute `proxy_type`, but then the code uses `proxytype`.
I think `proxytype` is correct, which means the if branch will never be taken and proxies will be ignored.
I also think the `hasattr()` checks are unnecessary -- they were added in a misguided attempt to fix an AttributeError for the misspelled `proxy_type` attribute.

If I'm right, user-visible effects of this bug are such that fetching /robots.txt for a site will ignore proxy settings.  (Unless those are configured in the session, in which case this bug has no user-visible effects at all?)

Why I think `proxytype` is correct: `git grep proxytype` shows multiple assignments and uses, whereas `git grep proxy_type` shows only this single `hasattr()` check.

Why I think `hasattr()` is unnecessary: `_allows_url()` is called from only one location (`allows_url()` in the same class), which unconditionally invokes `url_data.get_robots_txt_url()`, and `git grep get_robots_txt_url` shows that this method is defined only in one class, `HttpUrl`, which has `proxy` and `proxytype` attributes (they're accessed in `HttpUrl.get_request_kwargs` but are actually defined in the `ProxySupport` superclass).

It would be nice to have a unit test covering this part of the code, to prevent regressions and to make sure it's correct once fixed.

See https://github.com/linkchecker/linkchecker/issues/151#issuecomment-386018536 for background on how I stumbled upon this.

### LINKED ISSUE #431: Proxy code review
While reviewing #430 I'm looking at the linkchecker code that deals with proxies, and I thought it might be useful to record some of my observations:

- [ ] ProxySupport.set_proxy restricts proxy URLs to http: or https: because urllib2 doesn't support anything else:

 
 https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/proxysupport.py#L40-L45

  But we're not using urllib2, we're using requests, and requests also [supports socks5: and socks5h: schemes](https://requests.readthedocs.io/en/master/user/advanced/#socks).  We could remove this restriction and gain socks proxy support!

- [ ] ProxySupport.proxyauth is assigned twice and never used.  Requests supports proxy authentication directly by extracting the username and password from the proxy URL.

https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/proxysupport.py#L35

https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/proxysupport.py#L57-L61

This code could be removed.

- [ ] ProxySupport.ignore_proxy_host is calling urllib.request.proxy_bypass, which always checks the no_proxy environment variable, plus some system-specific other proxy bypass lists on Windows and MacOS.  Then ProxySupport.ignore_proxy_host checks the no_proxy environment variable again.  Why?

https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/proxysupport.py#L63-L73

The [commit](https://github.com/linkchecker/linkchecker/commit/e6f43b6822343f1625e78db07c9d5bd4875b3c93) message says "Fixed the no_proxy handling", and the changelog entry added in that commit says "Honor the no_proxy environment variable".  It was using urllib.proxy_bypass back then (2009).  That function in Python 2.7 also already checked the no_proxy environment variable.  What am I missing?

- [ ] Configuration.sanitize_proxies checks for http proxy and for ftp proxy, but does nothing about https proxy.
https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/configuration/__init__.py#L325-L337

GNOME at least has separate https proxy configuration in gsettings.  This may be relevant to PR #420.

- [ ] ProxySupport.get_netloc appears to be dead code, never called from anywhere.
https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/proxysupport.py#L75-L89

It can be removed

### LINKED ISSUE #586: Pass full proxy configuration to Requests
What about redirects that change the scheme?

If we do `requests.get('http://example.com', proxies={'http': 'http://localhost:8080'})`, and example.com redirects to https://example.com, requests will not know to use an https proxy!

This is not actually that straightforward: linkchecker sets allow_redirects=False and follows them manually.  But the way it does that:
https://github.com/linkchecker/linkchecker/blob/152dbeb9b886adf9bd71265460325c7781985e95/linkcheck/checker/httpurl.py#L241-L245
it calls requests.Session.resolve_rediects() to follow a sequence of redirects, but that method gets only one proxy configured for one scheme from self.get_request_kwargs()!

I think we should be passing the full proxy configuration to requests.

From https://github.com/linkchecker/linkchecker/pull/430#issuecomment-636289973 by **mgedmin**

cjmayo:
Set up the requests.Session() with the complete proxy configuration
to fix a problem with using an HTTP server as an HTTPS proxy and
potential redirection issues.

Requests will handle no_proxy; indeed if we dropped GNOME and KDE proxy
support we could also drop the remaining proxy code and leave it all to
Requests.

---

Support for the ftp_proxy environment variable is removed, but this was always partial and as well as the bug fix the scale of simplification is clear even from the number of lines removed.

Closes #586.
Closes #431.
Closes #152.


cjmayo:
Remove built-in GNOME and KDE proxy support

Only http_proxy was ever supported.

Requests uses urllib.request.getproxies().

Fedora 35 and Ubuntu 20.04 do set proxy environment variables when
settings are added through the GUI.

GNOME location of proxy settings is subject to change:
[1] https://wiki.gnome.org/Projects/NetworkManager/Proxies
https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas/-/issues/27

---

GNOME and KDE support was not quite right in v1 of these patches. On reflection the best choice is to remove it. It only ever provided HTTP proxy support (not HTTPS). On Fedora and Ubuntu it isn't being used because they set the environment variables. The KDE code is complex and possibly incomplete. The state of proxy configuration on Linux is not straightforward [1]. Let's follow the spirit of LinkChecker 10 of using the specialist libraries, where we introduced Beautiful Soup, and use Requests proxy support, simplifying LinkChecker and making it easier to maintain.

PR #608, opened 2022-01-19
dot command should prefer -o to direct pipe >

jvanalst:


cjmayo:
The Graphviz Command Line documentation uses a redirect:

```console
$ echo 'digraph { a -> b }' | dot -Tsvg > output.svg
```
https://www.graphviz.org/doc/info/command.html

Absent evidence of a problem, I think we are OK as we are.

PR #612, opened 2022-02-04
feat: add custom ignore schemes

JanMalte:


cjmayo:
Adding more schemes is great (never ending!).

Why create `custom` though and not just add to `other`? Other being something not in [1]'s Provisional, Permanent or Historical.

steam is already present in provisional:
https://github.com/linkchecker/linkchecker/blob/ff7e157a189b2e2a0608b376c9c260831c01f1f9/linkcheck/checker/unknownurl.py#L350

In addition to telegram (older?) looks like there is also tg [2] which pops up more in a quick Google.

Assuming they are added to other, it is defined in scripts/update_iana_uri_schemes.py:
[iana_uri_schemes_other = {](https://github.com/linkchecker/linkchecker/blob/ff7e157a189b2e2a0608b376c9c260831c01f1f9/scripts/update_iana_uri_schemes.py#L13)

Then linkcheck/checker/unknownurl.py is created by running scripts/update_iana_uri_schemes.sh.

Both update_iana_uri_schemes.py and unknownurl.py are committed.

A single commit and a force push is fine (no need for the history here).

[1] https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
[2] https://manpages.debian.org/testing/telegram-desktop/telegram-desktop.1.en.html#URI

PR #613, opened 2022-02-06
Add Telegram and WhatsApp schemes

JanMalte:
New MR for https://github.com/linkchecker/linkchecker/pull/612

I've created a new MR, as there are more changes then just the two new schemes


cjmayo:
Thanks - for the IANA update and script fix too!

PR #619, opened 2022-03-10
feat: add option to log warnings about redirected links

### LINKED ISSUE #617: feature: option to report/error on redirected links
## Summary

An option for warn/error on redirects would be super useful for catching useless redirects that only hurt page performance.  E.g. http to https redirections, trailing-slash redirects, `www` prefix addition/removal, etc..

`linkchecker --redirect=[warn|error] <URL>`

major0:
closes #617


ssbarnea:
I really hope to see this essential feature in. One of the most common problems is about URLs that get outdated and replaced by redirects. Now we cannot detect these redirects. What makes it worse is that sometimes these redirects are not sending to right content and instead to the external site root.

PR #647, opened 2022-09-05
Bring Developer Documentation and Tools up to date

### LINKED ISSUE #398: get rid of our Makefile
I suspect we do not need a Makefile anymore. First off, it starts with:

```
# This Makefile is only used by developers.
```

as a "developer", I can say I barely know not only how, but even *if* I need to use the darn thing. From #394, it seems that we still need it to regenerate translations, but that's something we might be able to work around with code like the `build_trans` command I imported from "deluge" in Monkeysign:

https://0xacab.org/monkeysphere/monkeysign/-/blob/2.x/monkeysign/translation.py

... itself dependent on a python implementation of `msgfmt`:

https://0xacab.org/monkeysphere/monkeysign/-/blob/2.x/monkeysign/msgfmt.py

... that seems to be just floating around the internetz and vendored like this in a few projects. I mention those because I think we might need them to get rid of the makefile but, now that I think about it, I can't imagine why we couldn't reuse `msgfmt` the same way that the Makefile does: gettext is just a build-dep and that's that...

Anyways, it seems we should get rid of the makefile. It has the following targets:

 * `all`: actually just an error message pointing at `doc/install.txt`
 * `clean`: calls setup.py clean and removes all sorts of other garbage
 * `distclean`: same with more garbage and egg files
 * `MANIFEST`: `setup.py sdist --manifest-only`
 * `locale`: chain loads into `po/Makefile`
 * `localbuild`: `setup.py build` and `locale`
 * `release`: `distclean` `releasecheck` `filescheck` but also chain-loads `dist sign register upload homepage tag changelog deb` (ugh)
 * `tag`: `git tag upstream/$(VERSION)`, `git push --tags origin upstream/$(VERSION)`
 * `upload`: `twine upload dist/$(ARCHIVE_SOURCE) dist/$(ARCHIVE_SOURCE).asc`
 * `homepage`: commits metadata and chain-loads into `doc/Makefile` and `doc/web/Makefile` (ugh!)
 * `register`: `setup.py register`
 * `deb`: custom debian package build procedure I (as the package maintainer) never used
 * `chmod`: what the hell
 * `dist`: custom wrapper around `setup.py sdist`
 * `check`: a test suite I didn't know about, calls `check-copyright`, `check-pofiles`, `py-tabdaddy`, `py-unittest2-compat` (no idea what those are), chain-loads into `doc/Makefile` with the `check` target, `doccheck`, and `pyflakes` targets
 * `doccheck`: `py-check-docstrings` no idea what that is
 * `filescheck`: fun fact: linkchecker runs linkchecker against its own compiled documentation here
 * `update-copyright`: calls unknown command `update-copyright`, presumably to change copyright headers
 * `releasecheck`: verifies version numbers and calls `setup.py check --restructuredtext`
 * `sign`: signs the release with GnuPG
 * `test`: ah? another test suite? chains into `pytest`
 * `pyflakes`: why not, let's hook pyflakes in the build too
 * `pep8`: yes, and pep8 too
 * `diff`: check vendored copies of Python stdlib code. we had `gzip.py` (gone) but it seems we still vendor `linkcheck/robotparser2.py` (now `urllib/robotparser.py`), which we should probably stop doing too
 * `changelog`: calls `github-changelog` (another unknown command), presumably to manage `doc/changelog.txt`
 * `count`: calls `sloccount` on the source
 * `ide`: calls Eclipse IDE on the project

Phew! What a mess! There are very few things we need from there, and I think most of them can be moved to `setup.py`, CI, or release documentation. In particular, we should:

 * [x] we already call pytest in from Tox, and in CI, so that's good
 * [ ] call `flake8` or similar against the source again, from Tox, but that doesn't need to be in our build setup
 * [ ] add docstrings checks and (why not) run linkchecker against itself in CI
 * [ ] build .mo files as part of CI? at least move it into a setup.py target

Everything else in there I think we can throw away. All that fancy code to release things has not been used forever, I don't even know if it still works. Our release process currently lives in `doc/development.mdwn` and doesn't even mention this makefile. So I wonder if we could already j ust throw away this entire file and be happy... I guess the checklist above would be nice sugar on top.

cjmayo:
Closes #398

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,6 @@ artifacts = [
 exclude = [
     ".github",
     ".gitattributes",
-    ".gitignore",
 ]
 
 [tool.hatch.build.targets.sdist]


pacenathan:
I was able to check out this branch and build and run with `hatch run linkchecker`. 

I did not get an `.eggs` directory.

However, `tox` gave me an error trying to install GeoIP, which the documentation says is optional, but presumably I don't have it and that's why it's choking. Should I install it, or should `tox` treat it as optional?

`hatch -e test run tests` actually ran the tests, but had the same failure I saw from before this PR (which I fixed in my CheckAnchor PR):

```
_________________________________________________ TestDecorators.test_timeit2 _________________________________________________
tests/test_decorators.py:48: in test_timeit2
    self.assertEqual(log.getvalue(), "f took 1.00 seconds\n()\n{}\n")
E   AssertionError: 'f took 1.01 seconds\n()\n{}\n' != 'f took 1.00 seconds\n()\n{}\n'
E   - f took 1.01 seconds
E   ?           ^
E   + f took 1.00 seconds
E   ?           ^
E     ()
E     {}
```

(`sleep()` is not guaranteed to sleep exactly as long as you ask for)

`hatchling build` worked, once I installed `hatch-vcs` (as documented in `install.txt`)

Note that I don't have `venv` installed, even though `install.txt` lists it.

PR #658, opened 2022-09-08
Fix a major throughput bottleneck for HTTP URLs

pacenathan:
This PR is meant to be based on my prior PR, #657, but I don't see a way to do that.

### DIFF
--- a/linkcheck/director/aggregator.py
+++ b/linkcheck/director/aggregator.py
@@ -21,7 +21,6 @@ import threading
 import requests
 import time
 import urllib.parse
-import random
 from .. import log, LOG_CHECK, strformat, LinkCheckerError
 from ..decorators import synchronized
 from ..cache import urlqueue
@@ -67,7 +66,6 @@ class Aggregate:
         self.cookies = None
         requests_per_second = config["maxrequestspersecond"]
         self.wait_time_min = 1.0 / requests_per_second
-        self.wait_time_max = max(self.wait_time_min + 0.5, 0.5)
         self.downloaded_bytes = 0
 
     def visit_loginurl(self):
@@ -152,8 +150,7 @@ class Aggregate:
                 wait = due_time - t
                 time.sleep(wait)
                 t = time.time()
-        wait_time = random.uniform(self.wait_time_min, self.wait_time_max)
-        self.times[host] = t + wait_time
+        self.times[host] = t + self.wait_time_min
 
     @synchronized(_threads_lock)
     def print_active_threads(self):


pacenathan:
Fixed. It looks like the force-push also resolved the extra commit.

pacenathan:
I treated it like a bug because it's clearly documented as a feature with a default of 10, yet the code enforces an effectively-hard cap of 4. You're assuming the code is right and the documentation is wrong. But there's very little evidence of that in the commit (7b34be590b52a85b207fae335d0885c85eb7b0aa); just a commit message mentioning "some improvements" plus a million code changes including some that faithfully recreate the old (default) cap of 10 per second, plus a few lines of code buried in all that that break the cap feature in an overly-complicated way. That doesn't seem intentional, to me. It seems like something that was accidentally left in, after an experiment that was never finished.

It also doesn't seem like a good idea for a link checker to have an artificial performance limit hard-coded into it. Sure, there are some cases when checking against public servers where a cap might be wise - but there's already a feature for that! Having a hard-coded max - that a user cannot possibly override - is no benefit to anybody.

Maybe if you want to be extra-cautious, just change the default cap to 4? That way anybody who was unknowingly depending on this bug (even though they thought they were getting a rate of 10/s) won't see any change, and people who want better link-check performance can at least still turn the setting up.

One other possible enhancement is to have a separate rate cap for internal vs. external, or for specific URLs. That's probably a feature I'd use, because I'm happy with a cap of 1000 against my local dev server (above that and it sometimes rejects requests), but I'd want to be more polite when checking against external servers.

But I'm not going to to bother implementing any of that; this is a bug in the system, and it's fixed now, and that's good enough. Someone else can add enhancements if they feel it's worth their time. If you want to be cautious, change the default to 4, and then it will be backwards compatible.

pacenathan:
Hm, I just realized that maybe you didn't see my analysis of this in my original commit message? For ease of finding it:

```
From my reading of the code, it looks like Aggregator / HttpUrl are
hard-coded to throttle requests to an HTTP server to a max of (an
average of) four times a second, even though the default documented
maxrequestspersecond is 10. This is because of code that was introduced
9 years ago (!) in commit https://github.com/linkchecker/linkchecker/commit/7b34be590b52a85b207fae335d0885c85eb7b0aa, such that the wait_for_host method
will always wait an average of at least a quarter of a second.

There's no explanation in the commit for why this was done, and no
reasonable explanation that I can understand for it. So I'm just taking
it out.

FWIW, this made runs that were running incredibly slowly against my local
HTTP server (i.e. minutes for HTTP vs 6 seconds for local file access)
run comparably-fast over both HTTP and local file access, after turning
maxrequestspersecond way up.

Note that I can still get back to the old slow performance, almost
exactly, by setting maxrequestspersecond to 4 :)
```

PR #659, opened 2022-09-08
Fix bug where resultscachesize setting was ignored

pacenathan:
This is meant to be based on #658, but I don't see a way to do that.


pacenathan:
Doing this rebase to take it out of the sequence forces me to rebase all the downstream branches, and re-resolve all the merge conflicts. It's not worth it.

PR #663, opened 2022-09-13
Add a feature to report all references to a URL

pacenathan:
Often when there is a broken URL, e.g. from documentation changes that moved a file or changed an anchor, there are many other pages that reference that URL. Before this commit, linkchecker would only report the first reference to that broken URL. With this commit, linkchecker is now able to optionally report all references to that broken url.

Command line argument: --allrefs

Config setting: output.reportallreferences

It is off by default, mainly because I didn't want to risk breaking existing behavior people might expect, and it would take a lot of existing-test overhaul. There is very little performance penalty to enabling it.

I intentionally built this on top of my other AnchorCheck code, because I need it to work with anchor checks. But it's just the one new commit.


pacenathan:
I forgot to mention that I didn't update the translations / docs, for this change, either. I'm not sure how to do that.

pacenathan:
FWIW, I found that bug when I was first looking into this problem. And reasoned that since it's been over two years since you touched the bug, and five years since it was introduced, that I should be careful about backwards compatibility. Of course, it's nicer to have simpler code instead of my additional complexity, but I'm surprised you're not worried about backwards compatibility.

Also, I suspect that commit had a purpose, and that its purpose was performance - I assume it made many check runs much faster. So I assume that if you revert it, you'll make many check runs much slower. (And generate a lot more traffic to web servers.)

Also, I would think the tests I introduced in 2a9ec3f should be easily adaptable to this revert, if you want to do that. That would at least ensure this bug doesn't crop up again, unintentionally.

A nice compromise might be to keep the new argument, but make it default to on, instead of off.

PR #669, opened 2022-09-26
Re-enable AnchorCheck plugin

cjmayo:


pacenathan:
I diffed this against my PR and reviewed it, and it mostly seems fine. My only objection is that you dropped the `decoded: <value>` bit from the warning output. That bit is valuable if you have a link that has a URL-encoded anchor, because it lets you know what value it's actually trying to match. If the issue is a typo or whatever, it's far easier to read e.g. `bad_|_(and—em-dashes)` than `bad_%%7C_(and%%E2%%80%%94em-dashes)`, and then find whatever value is close, in the actual target page.

pacenathan:
I think I went through that same sequence of implementations, when I first did it. (The first pass didn't have a variable, but then I intentionally added one so I could emit it in the log message. I think that was all in one commit, though.)

ofek:
When might this be released?

PR #679, opened 2022-10-17
Fix tests failing when run with pytest

cjmayo:
TypeError: 'NoneType' object is not callable

As per:
2cbff492 ("Fix http tests failing with pytest due to missing _()", 2022-10-03)


mgedmin (approved):
LGTM

PR #683, opened 2022-10-24
Move AnchorCheck local file handling into a new class

cjmayo:
When checking local files with AnchorCheck, anchors in URLs like "example/#anchor" are not supported.

Without AnchorCheck enabled, the Real URL reported for such URLs was changed to include the anchor when local file checking was added to AnchorCheck, but it is the directory that is checked. The same URL was also then used as the Parent URL for the check of each of the contents of that directory.

For FileUrl this is a revert of:
c221afda ("Enable AnchorCheck to be used with local files", 2022-10-03)

### DIFF
--- a/linkcheck/checker/fileurl.py
+++ b/linkcheck/checker/fileurl.py
@@ -26,7 +26,8 @@ from datetime import datetime
 from . import urlbase, get_index_html
 from .. import log, LOG_CHECK, fileutil, mimeutil, LinkCheckerError, url as urlutil
 from ..bookmarks import firefox
-from .const import WARN_FILE_MISSING_SLASH, WARN_FILE_SYSTEM_PATH
+from .const import (WARN_FILE_MISSING_SLASH, WARN_FILE_SYSTEM_PATH,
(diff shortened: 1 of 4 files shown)


linkcheck/checker/fileurl.py
```diff
@@ -314,3 +305,73 @@ def add_url(self, url, line=0, column=0, page=0, name="", base=None):
...
+        if self.parent_url is not None:
+            # URL joining with the parent URL only works if the query
+            # of the base URL are removed first.
+            # Otherwise the join function thinks the query is part of
+            # the file name.
+            urlparts = list(urllib.parse.urlsplit(self.base_url))
+            # ignore query part for filesystem urls
+            urlparts[3] = ''
+            self.base_url = urlutil.urlunsplit(urlparts)
+        super(FileUrl, self).build_url()
```
pacenathan, line 330: I see what you're doing, and it seems like a reasonable compromise for the short term. But things like this should be a sign that something is still off. So far, I can't envision any particular problem it might cause.

PR #710, opened 2022-12-31
Cache ClamAV signatures

### LINKED ISSUE #709: GHA failure due to ClamAV CDN rate limiting
I got an email that [the last GHA run](https://github.com/linkchecker/linkchecker/actions/runs/3810730541/jobs/6482987547) failed, and I see that the failure is caused by the "Download the ClamAV signature database" step:

```
Sat Dec 31 05:10:09 2022 -> main database available for download (remote version: 62)
Sat Dec 31 05:10:09 2022 -> ^Can't download main.cvd from https://database.clamav.net/main.cvd
Sat Dec 31 05:10:09 2022 -> ^FreshClam received error code 429 from the ClamAV Content Delivery Network (CDN).
Sat Dec 31 05:10:09 2022 -> This means that you have been rate limited by the CDN.
Sat Dec 31 05:10:09 2022 ->  1. Run FreshClam no more than once an hour to check for updates.
Sat Dec 31 05:10:09 2022 ->     FreshClam should check DNS first to see if an update is needed.
Sat Dec 31 05:10:09 2022 ->  2. If you have more than 10 hosts on your network attempting to download,
Sat Dec 31 05:10:09 2022 ->     it is recommended that you set up a private mirror on your network using
Sat Dec 31 05:10:09 2022 ->     cvdupdate (https://pypi.org/project/cvdupdate/) to save bandwidth on the
Sat Dec 31 05:10:09 2022 ->     CDN and your own network.
Sat Dec 31 05:10:09 2022 ->  3. Please do not open a ticket asking for an exemption from the rate limit,
Sat Dec 31 05:10:09 2022 ->     it will not be granted.
Sat Dec 31 05:10:09 2022 -> ^You are on cool-down until after: 2023-01-01 02:36:08
```

We ought to set up some cache if we want to keep using ClamAV in our test suite on GHA.

mgedmin:
We don't care if the signatures are stale (we only need the EICAR virus test signature), so run freshclam only if the cache is missing entirely.

Closes #709.

Also, use apt-get instead of apt to avoid

> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,10 +32,18 @@ jobs:
     steps:
       - name: Install OS dependencies
         run: |
-          sudo apt update
-          sudo apt install -y clamav-daemon geoip-database libgeoip-dev gettext
+          sudo apt-get update
+          sudo apt-get install -y clamav-daemon geoip-database libgeoip-dev gettext
+
+      - name: ClamAV cache
+        id: cache-clamav
+        uses: actions/cache@v3
+        with:
+          path: /var/lib/clamav
+          key: clamav-v1
 
       - name: Download the ClamAV signature database
+        if: ${{ steps.cache-clamav.outputs.cache-hit != 'true' }}
         run: |
           sudo service clamav-freshclam stop
           sudo freshclam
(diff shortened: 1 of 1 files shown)


mgedmin:
I force-pushed to fix a typo in the commit message.  Before that happened, GHA ran on my earlier push and actually saved the new cache in [this step](https://github.com/linkchecker/linkchecker/actions/runs/3812667851/jobs/6485993366).

I haven't yet seen a job that would use the new cache and skip the freshclam step.  I'm half-tempted to do another force-push.

PR #763, opened 2023-10-14
Python 3.12 final is out, stop using -dev

mgedmin:
Fixes the check-python-versions job in CI.


mgedmin:
(I've also updated the [branch protection rules](https://github.com/linkchecker/linkchecker/settings/branch_protection_rules/593487) to require Python 3.12 instead of Python 3.12-dev.

PR #767, opened 2023-10-31
Allow FIFOs to be used as config files

cdown:
There are some config options which have no equivalent command line option. Some may want to set these options dynamically or on a one-off basis where a static config file is not ideal, and one very easy way to do that is using process substitution:

    linkchecker --config <(printf '%s\n' '[filtering]' 'ignorewarnings=http-redirected') ...

This, however, does not work in the current code because these are typically implemented as FIFOs, which don't pass the `os.path.isfile` check:

    WARNING linkcheck.cmdline 2023-10-31 00:12:09,678 MainThread Unreadable config file: '/dev/fd/63'

Allow reading FIFOs as config input so that this is possible. `fileutil.is_readable` also now doesn't check if the path leads to a regular file: this is only used as part of cookie and config file input, and in both cases that's not really relevant.


cdown:
Much appreciated, thanks for getting back!

PR #771, opened 2023-11-02
Fix setting config["cookiefile"] regardless of checks

cjmayo:
options.cookiefile is set correctly later on in setup_config().

### DIFF
--- a/tests/configuration/test_config.py
+++ b/tests/configuration/test_config.py
@@ -191,3 +191,13 @@ class TestConfig(TestBase):
         # blacklist logger section
         self.assertEqual(config["failures"]["filename"], "blacklist")
         self.assertEqual(config["failures"]["encoding"], "utf-8")
+
+    def test_confparse_empty(self):
+        config = linkcheck.configuration.Configuration()
+        files = [get_file("config.empty")]
+        self.assertRaises(linkcheck.LinkCheckerError, config.read, files)
+
+    def test_confparse_missing(self):
+        config = linkcheck.configuration.Configuration()
+        files = [get_file("no_such_config")]
+        self.assertRaises(linkcheck.LinkCheckerError, config.read, files)
--- a/tests/test_cookies.py
+++ b/tests/test_cookies.py
@@ -18,6 +18,7 @@ Test cookie routines.
 """
 
 import os
+from pathlib import Path
 
 import linkcheck.cookies
 import linkcheck.configuration
@@ -80,3 +81,10 @@ class TestCookies(TestBase):
         aggregate.add_request_session()
         session = aggregate.get_request_session()
         self.assertEqual({c.name for c in session.cookies}, {"om", "multiple", "are"})
+
+    def test_empty_cookie_file(self):
+        self.assertRaises(
+            linkcheck.LinkCheckerError,
+            linkcheck.cookies.from_file,
+            Path(__file__).parent / "configuration/data/config.empty",
+        )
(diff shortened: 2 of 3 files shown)


linkcheck/command/setup_config.py
```diff
@@ -204,8 +202,11 @@ def setup_config(config, options):
     if options.useragent is not None:
         config["useragent"] = options.useragent
     if options.cookiefile is not None:
-        if fileutil.is_readable(options.cookiefile):
-            config["cookiefile"] = options.cookiefile
-        else:
+        if not fileutil.is_valid_config_source(options.cookiefile):
+            log.error(
+                LOG_CMDLINE, _("Cookie file %r does not exist."), options.cookiefile)
+        elif not fileutil.is_readable(options.cookiefile):
```
cdown, line 208: I did wonder while reading this code earlier, why log instead of terminating for invalid or unreadable config and cookie files? It feels like if the user explicitly specifies a cookie or config file and we were unable to use it, we should probably be as noisy as possible about that. Otherwise it can easily get lost in other things being printed.

linkcheck/command/setup_config.py
```diff
@@ -204,8 +202,11 @@ def setup_config(config, options):
     if options.useragent is not None:
         config["useragent"] = options.useragent
     if options.cookiefile is not None:
-        if fileutil.is_readable(options.cookiefile):
-            config["cookiefile"] = options.cookiefile
-        else:
+        if not fileutil.is_valid_config_source(options.cookiefile):
```
cdown, line 205: Feel free to rebase on #767 in this branch, if you prefer.

cdown (approved):
Thanks! Obviously I'm fairly new to this codebase, so take my approval with a grain of salt, but the concept looks sound to me.

linkcheck/configuration/confparse.py
```diff
@@ -57,6 +57,9 @@ def read(self, files):
         assert isinstance(files, list), "Invalid file list %r" % files
         try:
             self.read_ok = super().read(files)
+            if not self.sections():
+                raise LinkCheckerError(
+                    _("configuration files %s contain no sections.") % files)
```
cdown, line 62: nit: Capitalisation for "Configuration"?

linkcheck/cookies.py
```diff
@@ -40,6 +42,8 @@ def from_file(filename):
                 lines.append(line)
         if lines:
             entries.extend(from_headers("\r\n".join(lines)))
+        if not entries:
+            raise LinkCheckerError(_("No entries found"))
```
cdown, line 46: nit: maybe "No cookie entries found" so it's clear?

linkcheck/director/aggregator.py
```diff
@@ -45,8 +45,14 @@ def new_request_session(config, cookies):
...
     if config["cookiefile"]:
-        for cookie in from_file(config["cookiefile"]):
-            session.cookies.set_cookie(cookie)
+        try:
+            for cookie in from_file(config["cookiefile"]):
+                session.cookies.set_cookie(cookie)
+        except Exception as msg:
+            log.error(
+                LOG_CHECK,
+                _("Could not parse cookie file: %s. %s"), config["cookiefile"], str(msg)
```
cdown, line 54: nit: %s will already call `__str__`, bare `msg` is fine

linkcheck/fileutil.py
```diff
@@ -107,4 +107,5 @@ def is_writable_by_others(filename):
 
 def is_valid_config_source(filename):
     """Check if the file is a valid config file."""
-    return os.path.isfile(filename) or stat.S_ISFIFO(os.stat(filename).st_mode)
+    return os.path.exists(filename) and (
```
cdown, line 110: Ah, good thinking to add this. Thanks!

cdown (approved):
This looks great, thank you! I added some small nits but nothing material.

Issue #782, opened 2023-12-21
Ignore a warning for a specific URL only


nodet:
## Summary

I have documents to check that purposefully have links that are redirected.  I'd like to ignore the `http-redirected` warning on these URLs, but I don't want to ignore this warning for all other URLs in the document.

`ignorewarnings` in [the documentation](https://linkchecker.github.io/linkchecker/man/linkcheckerrc.html#filtering) for `linkcheckerrc` would apply to all URLs in my document, and would prevent me from detecting redirects on other URLs.

Did I miss something?

## Steps to reproduce

Here's a URL that reproduces the 'redirect', and that I would want to ignore:

 ```
$ linkchecker https://www.gurobi.com/release-notes/server/v11.0/
```

## Environment

 * Operating system: `Darwin my-macbookpro-2022.local 22.6.0 Darwin Kernel Version 22.6.0: Tue Nov  7 21:40:08 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_ARM64_T6000 arm64`
 * Linkchecker version: 10.4.0
 * Python version: 3.11.6
 * Install method: `pip install linkchecker`

## Configuration

```
$ linkchecker linkchecker -Dcmdline -q
DEBUG linkcheck.cmdline 2023-12-21 19:46:59,885 MainThread Python 3.11.6 (main, Oct 12 2023, 14:23:47) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
INFO linkcheck.cmdline 2023-12-21 19:46:59,891 MainThread Checking intern URLs only; use --check-extern to check extern URLs.
DEBUG linkcheck.cmdline 2023-12-21 19:46:59,891 MainThread configuration: [('aborttimeout', 300),
 ('allowedschemes', []),
 ('authentication', []),
 ('checkextern', False),
 ('cookiefile', None),
 ('csv', {}),
 ('debugmemory', False),
 ('dot', {}),
 ('enabledplugins', []),
 ('externlinks', []),
 ('failures', {}),
 ('fileoutput', []),
 ('gml', {}),
 ('gxml', {}),
 ('html', {}),
 ('ignoreerrors', []),
 ('ignorewarnings', []),
 ('internlinks', []),
 ('localwebroot', None),
 ('logger', 'NoneLogger'),
 ('loginextrafields', {}),
 ('loginpasswordfield', 'password'),
 ('loginurl', None),
 ('loginuserfield', 'login'),
[...]

PR #788, opened 2024-01-19
Ignore specific warnings -- Am I on the right track?

nodet:
I've been working on https://github.com/linkchecker/linkchecker/issues/782. The code as I have it now fits my needs. Before I invest more time in it to refine and document the behaviour, I would appreciate a confirmation that I'm on the right track.
Could one of the maintainers look at 12e47c987...b928c0e8c and let me know how this looks like to them?

I would then keep working on the topics mentioned in https://github.com/linkchecker/linkchecker/issues/782#issuecomment-1879621616. Please let me know if I'm missing something.

### DIFF
--- a/tests/checker/test_ignorewarnings.py
+++ b/tests/checker/test_ignorewarnings.py
@@ -28,6 +28,7 @@ class TestIgnoreWarnings(LinkCheckTest):
     Test whether ignoring of warnings per URL works.
     """
 
+    @need_network
     def test_given_warning_for_given_url_ignored(self):
         confargs = {
             "ignorewarningsforurls": [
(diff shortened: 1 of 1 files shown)


Kristinita:
**Type: Opinion** 💬

I think that while it’s better to have the `ignorewarningsforurls` option than not to have it, ignoring redirects isn’t the best idea.

`ignorewarningsforurls` will prevent linkchecker users from seeing unwanted redirects in the future. For example, redirects to spam sites: `https://examplesite.com/examplepage` → `https://spamsite.com`. Or, for example, redirects to the main page of the site, which happens when the URL is removed and the site redirects requests to the `index.html`: `https://examplesite.com/examplepage` → `https://examplesite.com`.

In my opinion, it’s better to **control** redirects rather than **ignore** them. It would be nice to allow linkchecker users to specify regular expressions for expected (allowed) redirects in the `linkcheckerrc` file. I provided the details of desired behavior in issue **#789**.

Thanks.

PR #794, opened 2024-02-03
Gh 782 -- Ignore specific warnings

nodet:
A previous discussion about the code was in https://github.com/linkchecker/linkchecker/pull/788.  
I simplified the history, added some logging and the documentation.


ssbarnea:
When we will see this feature being released? I cannot believe I spent an hour trying to make it work just to realize that no release happened since.

Issue #778, opened 2023-12-07
The number of threads has no impact


mmuehlfeldRH:
## Summary
It makes no difference how many threads I specify, linkchecker always needs the same amount of time to complete.

## Steps to reproduce
```
$ time linkchecker --threads=10 --recursion-level=1 --quiet --file-output=csv/utf-8//tmp/output.csv --check-extern https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/managing_networking_infrastructure_services/index
10 threads active,   146 links queued,    1 link in 157 URLs checked, runtime 1 seconds
10 threads active,   130 links queued,   17 links in 157 URLs checked, runtime 6 seconds
10 threads active,   115 links queued,   32 links in 157 URLs checked, runtime 11 seconds
10 threads active,    97 links queued,   50 links in 164 URLs checked, runtime 16 seconds
10 threads active,    79 links queued,   68 links in 179 URLs checked, runtime 21 seconds
10 threads active,    53 links queued,   94 links in 191 URLs checked, runtime 26 seconds
10 threads active,    42 links queued,  105 links in 195 URLs checked, runtime 31 seconds
10 threads active,    24 links queued,  123 links in 197 URLs checked, runtime 36 seconds
 3 threads active,     0 links queued,  154 links in 204 URLs checked, runtime 41 seconds

real	0m46.912s
user	0m6.654s
sys	0m0.341s
```

```
$ time linkchecker --threads=30 --recursion-level=1 --quiet --file-output=csv/utf-8//tmp/output.csv --check-extern https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/managing_networking_infrastructure_services/index
30 threads active,   124 links queued,    3 links in 157 URLs checked, runtime 1 seconds
30 threads active,   108 links queued,   19 links in 157 URLs checked, runtime 6 seconds
30 threads active,    95 links queued,   32 links in 157 URLs checked, runtime 11 seconds
30 threads active,    78 links queued,   49 links in 162 URLs checked, runtime 16 seconds
30 threads active,    59 links queued,   68 links in 176 URLs checked, runtime 21 seconds
30 threads active,    41 links queued,   86 links in 189 URLs checked, runtime 26 seconds
30 threads active,    25 links queued,  102 links in 193 URLs checked, runtime 31 seconds
30 threads active,    12 links queued,  115 links in 195 URLs checked, runtime 36 seconds
[...]

PR #796, opened 2024-02-07
Update linkcheckerrc.rst

ineiti:
Highlight the need of `LinkChecker` header for `maxrequestspersecond`.

See #778


ineiti:
Does that look better?

PR #797, opened 2024-02-07
Adding ignoreddosprotection configuration flag

ineiti:
This PR adds the following configuration flag:

ignoreddosprotection - ignore a missing 'LinkChecker' response header if the user requests more than 6 requests per second.

Help needed: I don't know how to create the `doc/i18n/*.po(|t)` files - they seem
to be auto-generated, but I don't see how to do that.

See #778


ineiti:
Thanks for the feedback - would the following be a good change of the PR?

- `s/DDOS/DOS/i` everywhere
- remove the `ignoredosprotection`
- add a `maxratedhosts` which takes an array of hostnames

Issue #813, opened 2024-03-26
Unable to run tests on macos due to missing GeoIP.h


ssbarnea:
## Summary

While trying to run tests on macos, via `tox -e py`, I got an error complaining about missing some headers.

## Actual result

```
      py_GeoIP.c:23:10: fatal error: 'GeoIP.h' file not found
      #include "GeoIP.h"
```

## Expected result

Be able to run tets.
## Environment

 * Operating system: Darwin m1 23.4.0 Darwin Kernel Version 23.4.0: Wed Feb 21 21:44:43 PST 2024; root:xnu-10063.101.15~2/RELEASE_ARM64_T6000 arm64
 * Linkchecker version: main branch
 * Python version: 3.12
 * Install method: pip

## Other notes

I installed the currently deprecated geoip from https://formulae.brew.sh/formula/geoip but that was not enough.

I think that it would be very useful to add a github actions pipeline that also runts the tests on macos, to catch such issues.

Interestingly I did not had problems using linkchecker itself on macos, only running the tests.

PR #814, opened 2024-03-26
Start testing on macos

ssbarnea:
Fixes: #813


ssbarnea:
@cjmayo hi! maybe allowing action to run by default? I don't see any security issue in that. Is bit frustrating to not even be able to test the PR.

PR #815, opened 2024-03-26
fix: header check should be case-insensitive

shepherdjerred:
HTTP headers are not case sensitive. This becomes an issue when using Caddy to host pages to be checked by the linkchecker project. Caddy will automatically lowercase header names, which results in checker throughput being limited.

This PR modifies the check for the `LinkChecker` header to perform the check case-insensitively.

>   Just as in HTTP/1.x, header field names are strings of ASCII
>   characters that are compared in a case-insensitive fashion.  However,
>   header field names MUST be converted to lowercase prior to their
>   encoding in HTTP/2.  A request or response containing uppercase
>   header field names MUST be treated as malformed

https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2


shepherdjerred:
Thank you for looking into this. I made a few large improvements to our linkchecking setup a little while ago. Based on your findings, my guess is that something else is responsible for the speedup I've seen.

Issue #47, opened 2017-03-23
[Question] LinkChecker doesn't check kristinita.ru


Kristinita:
### 1. Summary

LinkChecker don't check site **<http://kristinita.ru>**.

### 2. Settings

+ Source of my GitHub Pages site — **<https://github.com/Kristinita/Kristinita.github.io>**;
+ Sitemap — **<https://github.com/Kristinita/Kristinita.github.io/blob/master/sitemap.xml>**;
+ `robots.txt` — **<https://github.com/Kristinita/Kristinita.github.io/blob/master/robots.txt>**.

```
User-agent: *
Disallow: /images/
Disallow: /js/
Disallow: /css/
Disallow: /zohoverify/
Disallow: /Life-hacks/
Disallow: /Smert-svobode/

User-Agent: LinkChecker
Allow: /
```

### 3. Expected behavior

I read [**manual**](https://wummel.github.io/linkchecker/man1/linkchecker.1.html):

> The most common use checks the given domain recursively, plus any URL pointing outside of the domain:
   linkchecker http://www.example.net/
Beware that this checks the whole site which can have thousands of URLs. Use the -r option to restrict the recursion depth. 

Expected behavior — check all links in pages of **<http://kristinita.ru>** site. For example, for **<http://filimania.ru>** I get output:

```python
Statistics:
Downloaded: 12.2MB.
Content types: 719 image, 11027 text, 0 video, 0 audio, 12443 application, 61 mail and 2768 other.

URL lengths: min=19, max=18686, avg=61.

The check has been interrupted; results are not complete.
That's it. 27018 links in 378 URLs checked. 0 warnings found. 381 errors found.
Stopped checking at 2017-03-23 16:46:44+003 (4 minutes, 45 seconds)
```

### 4. Actual behavior

Examples of my attempts:

```
D:\Kristinita>linkchecker http://kristinita.ru
INFO 2017-03-23 19:07:00,473 MainThread Checking intern URLs only; use --check-extern to check extern URLs.
LinkChecker 9.3              Copyright (C) 2000-2014 Bastian Kleineidam
LinkChecker comes with ABSOLUTELY NO WARRANTY!
This is free software, and you are welcome to redistribute it
under certain conditions. Look at the file `LICENSE' within this
distribution.
Get the newest version at http://wummel.github.io/linkchecker/
Write comments and bugs to https://github.com/wummel/linkchecker/issues
Support this project at http://wummel.github.io/linkchecker/donations.html

[...]

PR #927, opened 2026-07-02
Allow for urlset starting further into a sitemap

mradcliffe:
### Problem/Motivation

Hi, I was testing and evaluating linkchecker and its support for parsing a provided sitemap.xml URL.

When using linkchecker with a URL to a sitemap.xml, if the server responds with `application/xml`, then linkchecker will report the URL is not parseable. This is a follow-up to #47.

Example output snippet:

```sh
linkchecker --verbose -f .linkcheckerrc https://example.com/sitemap.xml
LinkChecker 10.6.0
Copyright (C) 2000-2016 Bastian Kleineidam, 2010-2025 LinkChecker Authors
LinkChecker comes with ABSOLUTELY NO WARRANTY!
This is free software, and you are welcome to redistribute it under
certain conditions. Look at the file `COPYING' within this distribution.
Read the documentation at https://linkchecker.github.io/linkchecker/
Write comments and bugs to https://github.com/linkchecker/linkchecker/issues

Start checking at 2026-07-02 12:31:50-004

URL        `https://example.com/sitemap.xml`
Real URL   https://example.com/sitemap.xml
Check time 0.028 seconds                                                                                                                                                                                                             
D/L time   0.000 seconds
Size       14.24KB
Warning    [url-content-type-unparseable] The URL with content
           type 'application/xml' is not parseable.                                                                                                                                                                                  
Result     Valid: 200 OK

Statistics:
Downloaded: 14.24KB.
Content types: 0 image, 0 text, 0 video, 0 audio, 1 application, 0 mail and 0 other.
URL lengths: min=36, max=36, avg=36.

That's it. 1 link in 1 URL checked. 1 warning found. 0 errors found.
Stopped checking at 2026-07-02 12:31:51-004 (1 seconds)
```

### Proposed resolution

Increases bytes read in `guess_mimetype_read` from 70 to 255 to account for some extra lines.

### Questions

- Is increasing the bytes read a good idea? Arbitrarily increasing the string length to pass into regex checking might cause performance issues?

I tried searching for a place where I could add a test where the server would respond with the various content mime types, but I think that is beyond my abilities as I am not familiar with Python. I don't think the tests cover all the content mime type cases. If this is necessary, please do fork and create a new pull request.

### DIFF
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -102,6 +102,7 @@ class UrlBase:
         "text/plain+chromium": "chromium",
         "application/x-plist+safari": "safari",
         "text/vnd.wap.wml": "wml",
+        "application/xml": "sitemap",
         "application/xml+sitemap": "sitemap",
         "application/xml+sitemapindex": "sitemapindex",
         "application/pdf": "pdf",
--- a/linkcheck/checker/urlbase.py
+++ b/linkcheck/checker/urlbase.py
@@ -102,7 +102,6 @@ class UrlBase:
         "text/plain+chromium": "chromium",
         "application/x-plist+safari": "safari",
         "text/vnd.wap.wml": "wml",
-        "application/xml": "sitemap",
         "application/xml+sitemap": "sitemap",
         "application/xml+sitemapindex": "sitemapindex",
         "application/pdf": "pdf",
--- a/linkcheck/mimeutil.py
+++ b/linkcheck/mimeutil.py
@@ -92,7 +92,7 @@ def guess_mimetype_read(read):
     """Try to read some content and do a poor man's file(1)."""
     mime = None
     try:
-        data = read()[:70]
+        data = read()[:255]
     except Exception:
         pass
     else:


cjmayo:
I haven't had time to fully check but I think the proposed change is not valid because it would suggest everything application/xml is a sitemap.

If the sitemap is valid then it should already be detected:
https://github.com/linkchecker/linkchecker/blob/f0ecbe8e8e218342955bb84de2768349489bbc24/linkcheck/checker/httpurl.py#L382-L383

enable debug logging to check:
https://github.com/linkchecker/linkchecker/blob/f0ecbe8e8e218342955bb84de2768349489bbc24/linkcheck/checker/httpurl.py#L386-L387

mradcliffe:
Thank you very much for the feedback, cjmayo.

> If the sitemap is valid then it should already be detected:

The sitemap is valid because the hack that I have works fine and the sitemap.xml is parsed and checked completely. I checked `mimeutil.guess_mimetype_read` to see what it is doing, and then tested on the file I downloaded locally with curl:

```sh
file sitemap.xml
sitemap.xml: XML 1.0 document, ASCII text
```

```sh
head -c 70 sitemap.xml 
<?xml version="1.0" encoding="UTF-8"?>
<?xml-stylesheet type="text/xsl%    
```

The urlset element does not start on this file until after 230 bytes, which fails the simple check in that method. There's a comment in there as well, which if I get rid of, then urlset starts at 128 bytes.

So I think the issue is with the simple mimetype guesser, but I understand that reading an arbitrary number of bytes itself might not be the correct fix since the string to regex check gets longer.  I'll change the code to do that at least.

cjmayo:
Seems reasonable. The increase to 70 happened in 2014! 380f1445

